PR #73, opened 2023-03-29
Added a namespace task

scohen:
When editing the remote_control app while lexical is running as your language server, it's easy to change a file and change lexical's behavior. To prevent this, the task `mix namespace` will read all the beam files, strip out any mentions of `Elixir.Lexical` in atoms and replace that with `Elixir.LXRelease`.

There are a copule caveats:
1. The mod: keys in the mix.exs files in server and remote_control need to know about namespacing
2. I'm going to bet this wreaks havoc with test files.
3. My lsp doesn't connect to the server when it's namespaced, so there's more work to do


config/dev.exs
```diff
@@ -1 +1,8 @@
 import Config
+
+if System.get_env("NAMESPACE") do
```
scohen, line 3: can't we do this in the mix task?
We can, we can do this in the mix task.

apps/common/lib/mix/tasks/namespace.ex
```diff
@@ -36,7 +36,7 @@ defmodule Mix.Tasks.Namespace do
       IO.write("Applying namespace: #{percent_complete} complete")
     end)
 
-    Mix.Shell.IO.info("\nNamspace applied. Generating apps")
+    Mix.Shell.IO.info("\nNamespace applied. Generating apps")
```
scohen, line 39: what, you didn't like my spelling?
At least you know _why_ I mistyped it.

Sleepful: the datahand speaks

config/dev.exs
Sleepful: The thing is that the `mix release lexical` steps for release always runs *after* the `mix compile` and this needs to be cleaned before `mix compile`, here it does run before compile.

scohen: Release is totally configurable. You can even Alia's it.

PR #77, opened 2023-03-30
Enabled strict credo

scohen:
In order to get alias sorting, I enabled strict credo and fixed up the existing issues.

### DIFF
--- a/apps/server/test/lexical/server/code_intelligence/completion_test.exs
+++ b/apps/server/test/lexical/server/code_intelligence/completion_test.exs
@@ -1,16 +1,16 @@
 defmodule Lexical.Server.CodeIntelligence.CompletionTest do
-  alias Lexical.Server.Project.Dispatch
-  alias Lexical.Server.CodeIntelligence.Completion
   alias Lexical.Project
   alias Lexical.Protocol.Types.Completion.Context, as: CompletionContext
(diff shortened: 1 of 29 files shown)


Blond11516 (approved):

apps/server/test/lexical/server/code_intelligence/completion_test.exs
```diff
@@ -1,16 +1,16 @@
...
   alias Lexical.Project
   alias Lexical.Protocol.Types.Completion.Context, as: CompletionContext
   alias Lexical.Protocol.Types.Completion.Item, as: CompletionItem
   alias Lexical.RemoteControl
   alias Lexical.Server
+  alias Lexical.Server.CodeIntelligence.Completion
+  alias Lexical.Server.Project.Dispatch
   alias Lexical.SourceFile
 
   use ExUnit.Case
```
scottming, line 11: Now we put the `use` before import?

scohen: isn't that how it is?

Credo by default enforces use before _alias_, which sucks and is wrong, IMO. If I start doing strict module order, I'm going to definitely change the defaults.

scottming: https://github.com/elixir-lsp/elixir-ls/pull/840

our rule is put the `use` at last. @scohen

PR #78, opened 2023-03-30
Refactored completions

scohen:
The completions module was getting pretty hard to follow and understand, and with a mind to offering a modular way to incorporate lexical completions into other projects, I did the following.

1. Removed the completion context from the completion environment, freeing it from any dependency on LSP
2. Made a Translatable protocol, so that other libraries can translate their completion items into LSP completion items.
3. Made a use-able translator module that assists in building completion items.

I also split our completions into translators, so that each completion type is now in its own file. Prior, they were in completions.ex

I also separated unit tests into their own modules, which made it apparent that we wern't testing some code paths. I also reafactored the unit tests themselves to depend on a `CompletionCase` which uses some source code rewriting that `CodeModCase` used as well.


apps/common/test/support/code_sigil.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Lexical.Test.CodeSigil do
+  def sigil_q(text, opts \\ []) do
+    ["", first | rest] = text |> String.split("\n")
+    base_indent = indent(first)
+    indent_length = String.length(base_indent)
```
Blond11516, line 5:
```suggestion
    indent_length =
      first
      |> indent()
      |> String.length()
```

apps/common/test/support/code_sigil.ex
```diff
@@ -0,0 +1,41 @@
...
+
+  defp maybe_trim(iodata, [?t]) do
+    iodata
+    |> IO.iodata_to_binary()
+    |> String.trim_trailing()
+  end
+
+  defp maybe_trim(iodata, _) do
+    IO.iodata_to_binary(iodata)
+  end
```
Blond11516, line 20:
```suggestion
    [first | rest]
    |> Enum.map_join("\n", &strip_leading_indent(&1, indent_length))
    |> IO.iodata_to_binary()
    |> maybe_trim(opts)
  end

  defp maybe_trim(binary, [?t]) do
    String.trim_trailing(binary)
  end

  defp maybe_trim(binary, _) do
    binary
  end
```
Don't know if there's a particular reason why `iodata_to_binary` was duplicated but `maybe_trim` doing the conversion feels unexpected to me.

apps/server/test/lexical/server/code_intelligence/completion/translations/function_test.exs
```diff
@@ -0,0 +1,97 @@
...
+  use Lexical.Test.Server.CompletionCase
+
+  describe "function completions" do
+    test "deprecated functions are marked", %{project: project} do
+      assert {:ok, completion} =
+               project
+               |> complete("Enum.filter_map|")
+               |> fetch_completion("filter_map")
+
+      assert completion.detail
```
scottming, line 11: Are you asserting the `completion.detail`  not nil?

scottming (approved):

Blond11516 (approved):

apps/server/test/lexical/server/code_intelligence/completion/translations/function_test.exs
scohen: correct; that there's something there.

apps/common/test/support/code_sigil.ex
scohen: agreed.

PR #79, opened 2023-03-31
Support go to definition

scottming:
This is a basic go-to-definition feature that uses a tracer to create an index of our interests and then looks for the index under our current cursor. Currently, I have implemented the jumping functionality for **internal functions and modules** within the project. In normal circumstances, it should be able to jump to all modules and functions within the project, including those imported through use or import statements (ElixirLs does not implement this), or those in test files (which need editing because the test files are not compiled).

I have created a Draft PR to confirm that my approach is correct. The biggest advantage of not relying on ElixirSense is that we have complete control over the process, for example, implementing the go to Elixir source code is not a difficult task.

### Todo list

* [ ] More tests; currently, only a few unit tests are available
* [ ] Index structure could be optimized
* [ ] go to functions and modules in deps
* [ ] go to functions and modules of Elixir source code

### Other ideas

The current approach for jumping to internal project modules and functions is functional (however, it doesn’t work in the quote block), but for deps and Elixir source code, we do not have to build indexes for all of them. My idea is to make some determinations when there is a remote reference and decide whether it is a dep or Elixir source code, and then build separate indexes.

```elixir
iex(4)> BeamFile.chunk(Enum,  :compile_info)
{:ok,
 [
   version: '8.2',
   options: [
     :no_spawn_compiler_process,
     :from_core,
     :no_core_prepare,
     :no_auto_import,
     :inline_list_funcs,
     {:inline,
      [reduce_enumerable: 3, reduce_by: 3, reduce: 3, entry_to_string: 1]}
   ],
   source: '/home/build/elixir/lib/elixir/lib/enum.ex'
 ]}
iex(5)> BeamFile.chunk(Rewrite,  :compile_info)
{:ok,
 [
   version: '8.2.3',
   options: [:no_spawn_compiler_process, :from_core, :no_core_prepare,
    :no_auto_import],
   source: '/Users/scottming/Code/recode/deps/rewrite/lib/rewrite.ex'
 ]}
```

for example, use BeamFile to determine the source from where.


scohen:
This PR uses code taken from elixir-ls. Closing with extreme prejudice. 

Never take code from other projects that you didn't write.

PR #80, opened 2023-03-31
Fix diagnostics in namespace (but with old commit oops)

Sleepful:
worked fine when ran with

`mix clean && NAMESPACE=1 mix release lexical --overwrite`

### DIFF
--- a/apps/common/lib/mix/tasks/namespace/release.ex
+++ b/apps/common/lib/mix/tasks/namespace/release.ex
@@ -4,21 +4,42 @@ defmodule Mix.Tasks.Namespace.Release do
   """
   use Mix.Task
 
-  def run([release_version_path, remote_control_ebin]) do
+  def run(_) do
+    release = Mix.Release.from_config!(:lexical, Mix.Project.config(), [])
+    release_path = release.path
+    release_version_path = release.version_path
+
+    app_names = ~w(remote_control common common_protocol server protocol)a
+    app_paths = Enum.map(app_names, fn app ->
+      vsn = Mix.Project.in_project(app, Mix.Project.apps_paths()[app],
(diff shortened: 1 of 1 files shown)


apps/common/lib/mix/tasks/namespace/release.ex
```diff
@@ -4,21 +4,42 @@ defmodule Mix.Tasks.Namespace.Release do
...
+    paths = script_paths ++ app_paths
 
     Enum.each(paths, fn p ->
       string = File.read!(p)
       # avoid replacing directory path with negative lookbehind:
       string = String.replace(string, ~r/(?<!\/)remote_control/, "lx_remote_control")
+      string = String.replace(string, ~r/(?<!\/)common(?!_)/, "lx_common")
+      string = String.replace(string, ~r/(?<!\/)common_protocol/, "lx_common_protocol")
+      string = String.replace(string, ~r/(?<!\/)server/, "lx_server")
+      string = String.replace(string, ~r/(?<![\/_])protocol/, "lx_protocol")
```
Sleepful, line 32: some of these are funny because e.g.

`protocol` should be replaced to `lx_protocol`
`common` should be replaced to `lx_common`
but
`common_protocol` should be replaced to `lx_common_protocol`

it should *not* replace to `lx_common_lx_protocol`

Sleepful: also this is a bit of a "hit and miss" job, I am assuming I don't need to add the `proto` app, because this worked fine. But through usage we find ze bugz.

scohen: why no pipeline?

Sleepful:
closing in favor of https://github.com/lexical-lsp/lexical/pull/89

PR #84, opened 2023-04-01
Support go to definition with elixir_sense

scottming:


apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -0,0 +1,76 @@
...
+        {:error, "Could not open source file: #{inspect(file_path)}"}
+    end
+  end
+
+  defp to_persisted_range(text, line, column) do
+    case Code.Fragment.surround_context(text, {line, column}) do
+      %{begin: begin, end: end_} ->
+        {begin_line, begin_column} = begin
+        {end_line, end_column} = end_
+        to_range({begin_line, begin_column}, {end_line, end_column})
```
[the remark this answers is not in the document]
scottming: This is indeed a mistake, an unnecessary operation.

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -0,0 +1,76 @@
...
+    parse_location(maybe_location, source_file)
+  end
+
+  defp parse_location(%ElixirSense.Location{} = location, source_file) do
+    %{file: file, line: line, column: column} = location
+    file_path = file || source_file.path
+
+    case location_to_range(file_path, line, column) do
+      {:ok, {source_file, range}} ->
+        {:ok, %{source_file: source_file, range: range}}
```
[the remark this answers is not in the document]
scottming: I prefer to return `{:ok, {source_file, range}}` here, because the result will be `nil` too, now the spec looks like

```elixir
  @spec definition(SourceFile.t(), Position.t()) ::
          {:ok, {SourceFile.t(), Range.t()}} | {:error, String.t()}
  def definition(%SourceFile{} = source_file, %Position{} = position) do
```

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -0,0 +1,76 @@
...
+      source_file
+      |> SourceFile.to_string()
+      |> ElixirSense.definition(position.line + 1, position.character + 1)
+
+    parse_location(maybe_location, source_file)
+  end
+
+  defp parse_location(%ElixirSense.Location{} = location, source_file) do
+    %{file: file, line: line, column: column} = location
+    file_path = file || source_file.path
```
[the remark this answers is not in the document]
scottming: %Location{file: nil, line: line, column: column}

nil means current file.

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -0,0 +1,76 @@
...
+  defp to_persisted_range(text, line, column) do
+    case Code.Fragment.surround_context(text, {line, column}) do
+      %{begin: begin, end: end_} ->
+        {begin_line, begin_column} = begin
+        {end_line, end_column} = end_
+        to_range({begin_line, begin_column}, {end_line, end_column})
+
+      _ ->
+        # NOTE: sometimes the column is 1, the cursor looks strange
+        # so we need to normalize it to the first char column
```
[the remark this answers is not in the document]
scottming: image

I mean the above picture.

and after plus the leading_spaces, then become:

image

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -0,0 +1,76 @@
...
+    with {:ok, source_file} <- SourceFile.Store.open_temporary(uri),
+         {:ok, {:line, text, _, _, _}} <- SourceFile.fetch_line_at(source_file, line - 1) do
+      {:ok, {source_file, to_persisted_range(text, line, column)}}
+    else
+      _ ->
+        {:error, "Could not open source file: #{inspect(file_path)}"}
+    end
+  end
+
+  defp to_persisted_range(text, line, column) do
```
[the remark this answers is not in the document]
scottming: This is a spelling mistake. it should be `precise`

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -0,0 +1,76 @@
...
+      %{begin: begin, end: end_} ->
+        {begin_line, begin_column} = begin
+        {end_line, end_column} = end_
+        to_range({begin_line, begin_column}, {end_line, end_column})
+
+      _ ->
+        # NOTE: sometimes the column is 1, the cursor looks strange
+        # so we need to normalize it to the first char column
+        column = if column == 1, do: first_char_column(text), else: column
+        to_range({line, column}, {line, column})
```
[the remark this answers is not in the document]
scottming: I changed to `to_range(start_pos, end_pos)`, but passing the tuple to the `to_range` is correct because returned column is `1`, which means there is no particular definition there, just something generated by the macro, like `Responses.GotoDefinition.new`, the `new` does not exits, only the `defresponse`.

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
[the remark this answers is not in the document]
scohen: when will the sourcefile and range be nil?

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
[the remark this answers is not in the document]
scohen: Can you update the comment so it describes the situation rather than saying it looks weird?
for example:

If the column is 1, but the code doesn't start on the first column, the cursor will be placed to the left of the actual definition, which isn't what we want.

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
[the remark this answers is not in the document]
scohen: oh heavens. Elixir sense has the worst API. Would it have killed them to always have the path?

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -0,0 +1,76 @@
...
+      |> ElixirSense.definition(position.line + 1, position.character + 1)
+
+    parse_location(maybe_location, source_file)
+  end
+
+  defp parse_location(%ElixirSense.Location{} = location, source_file) do
+    %{file: file, line: line, column: column} = location
+    file_path = file || source_file.path
+
+    case location_to_range(file_path, line, column) do
```
scohen, line 23: why bother with the case here?

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
scohen: you're unpacking two tuples just to recreate them. 
```suggestion
      %{begin: start_pos, end: end_pos} ->
        to_range(start_pos, end_pos)
```

scohen (commented):
Except the case, looks good
Please add tests.

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -0,0 +1,76 @@
...
+
+  defp to_range(begin, end_) do
+    {begin_line, begin_column} = begin
+    {end_line, end_column} = end_
+
+    %Range{
+      start: %Position{line: begin_line - 1, character: begin_column - 1},
+      end: %Position{line: end_line - 1, character: end_column - 1}
+    }
+  end
```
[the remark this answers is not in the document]
scottming: I changed the code; you should need to reflash it.

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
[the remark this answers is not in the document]
scottming: I prefer this:

> If the column is 1, but the code doesn't start on the first column, which isn't what we want. 
> The cursor will be placed to the left of the actual definition.

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -0,0 +1,370 @@
...
+
+    with :ok <- SourceFile.Store.open(uri, content, 1),
+         {:ok, uses_file} <- SourceFile.Store.fetch(uri) do
+      uses_file
+    end
+  end
+
+  defp cursor_to_position(cursor, source_file) do
+    find_line = fn source_file, cursor ->
+      Enum.find(Tuple.to_list(source_file.document.lines), fn line(text: text) ->
```
scohen, line 43: we already have a module called `CursorSupport` to handle this, maybe you should use that instead, and just turn the source file into a string.
It's a violation of encapsulation to pull out the lines like this.
Also, the underlying document is `Enumerable`, so you don't need to ever do this kind of thing.

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -0,0 +1,370 @@
...
+        String.contains?(text, cursor_full_text)
+      end)
+    end
+
+    line(line_number: line_number) = find_line.(source_file, cursor)
+    {:ok, text} = SourceFile.fetch_text_at(source_file, line_number)
+    [cursor | _] = String.split(cursor, "|", parts: 2)
+    [before_cursor, _] = String.split(text, cursor, parts: 2)
+    character = String.length(before_cursor) + String.length(cursor) + 1
+    Position.new(line_number, character)
```
scohen, line 54: all this duplicates a lot of what you get in `CursorSupport`

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -0,0 +1,370 @@
...
+      position = cursor_to_position("MyDefinition.gree|", uses_file)
+
+      {:ok, {source_file, range}} = Definition.definition(uses_file, position)
+
+      assert annotate(source_file, range) == ~q[
+        defmodule MyDefinition do
+        ...
+
+          def greet(name) do
+        #     ^^^^^
```
scohen, line 97: this is copied from elixir_ls. I really don't like this pattern, it's too weird; just use cursor support.

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -0,0 +1,370 @@
...
+        %%
+        ...
+
+        binary_to_atom(Binary) ->
+        ^^^^^^^^^^^^^^
+      ]
+    end
+  end
+
+  defp annotate(source_file, range) do
```
scohen, line 348: this approach makes several assumptions that will bite us in the future, one of which is that the first line of a file is a module definition. Using CursorSupport will make doing this not necessary.

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -0,0 +1,370 @@
...
+    with :ok <- SourceFile.Store.open(uri, content, 1),
+         {:ok, uses_file} <- SourceFile.Store.fetch(uri) do
+      uses_file
+    end
+  end
+
+  defp cursor_to_position(cursor, source_file) do
+    find_line = fn source_file, cursor ->
+      Enum.find(Tuple.to_list(source_file.document.lines), fn line(text: text) ->
+        cursor_full_text = String.replace(cursor, "|", "")
```
scohen, line 44: this is irrelevant, but you're doing this replacement in a loop rather than doing it at the beginning of the loop. this means you're doing `n` allocations rather than 1.
It's also logically incorrect, as you won't be able to work with a line that has both a pipe and a cursor.

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -0,0 +1,370 @@
+defmodule Lexical.RemoteControl.CodeIntelligence.DefinitionTest do
```
scohen, line 1: This approach seems off to me. You have a project with files in it (I assume they're for the definitions), but you're also polluting the current VM with compiled artifacts via `Code.compile_string`. What are you aiming for?

Keeping this all encapsulated inside a project VM would make this somewhat easier, though I must admit, I'm not a huge fan of the fixtures being so far away from the testing code (but at the same time, there's stuff you just cannot reproduce in the current VM)

I suspect things are this way to make the tests run faster, but frankly, having a fast test that doesn't work like the prod code is of little value when you get to those edge cases.

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -0,0 +1,370 @@
...
+    %{project: project(:navigations)}
+  end
+
+  defp with_referenced_file(%{project: project}, file \\ "my_definition.ex") do
+    path = file_path(project, Path.join("lib", file))
+    Code.compile_file(path)
+    %{uri: SourceFile.Path.ensure_uri(path)}
+  end
+
+  defp open_uses_file(project, content) do
```
scohen, line 29: the naming of this function is confusing, as is the return value.
calling it `source_file` seems a little better, and having it return `{:ok, source_file}` or `{:error, reason}` is better. 
Right now, it returns an error tuple or a naked source_file, making it hard to pattern match against.

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -0,0 +1,370 @@
...
+      |> SourceFile.Path.ensure_uri()
+
+    with :ok <- SourceFile.Store.open(uri, content, 1),
+         {:ok, uses_file} <- SourceFile.Store.fetch(uri) do
+      uses_file
+    end
+  end
+
+  defp cursor_to_position(cursor, source_file) do
+    find_line = fn source_file, cursor ->
```
scohen, line 42: nit: cursor_line

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
scottming: The purpose here is to make the test run faster, but you are right that fast tests that are incorrect have no value in edge cases. I have changed it to run on an isolated node.

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
scottming: changed it to return `{:ok, uses_file} | {:error, reason}`

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
scottming: Removed

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
scottming: changed to `«range of definition»`

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
scottming: Removed

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -0,0 +1,58 @@
...
+    %{file: file, line: line, column: column} = location
+    file_path = file || source_file.path
+    uri = SourceFile.Path.ensure_uri(file_path)
+
+    with {:ok, source_file} <- SourceFile.Store.open_temporary(uri),
+         {:ok, text} <- SourceFile.fetch_text_at(source_file, line) do
+      {:ok, {source_file, to_precise_range(text, line, column)}}
+    else
+      _ ->
+        {:error, "Could not open source file: #{inspect(file_path)}"}
```
scohen, line 30: this error message could possibly be wrong; you could also be fetching a line that's out of bounds.

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -0,0 +1,318 @@
...
+
+  defp open_uses_file(project, content) do
+    uri = uses_file_uri(project)
+
+    with :ok <- RemoteControl.call(project, SourceFile.Store, :open, [uri, content, 1]) do
+      RemoteControl.call(project, SourceFile.Store, :fetch, [uri])
+    end
+  end
+
+  setup_all ~w(with_navigation_project)a
```
scohen, line 47: i think i'd prefer a normal `setup_all` block, it's clearer, and you're not reusing this. Also, using `~w` in this instance doesn't save anything compared to `[:with_navigation_project]`

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -0,0 +1,318 @@
...
+
+  setup %{project: project} do
+    uri = uses_file_uri(project)
+
+    # NOTE: We need to make sure every tests start with fresh caller content file
+    on_exit(fn ->
+      :ok = RemoteControl.call(project, SourceFile.Store, :close, [uri])
```
scohen, line 54: why's the source file store running in the remote control project? That should be running in the `server` project. You can also set it up via `start_supervised` in a test.

apps/remote_control/test/fixtures/navigations/mix.exs
```diff
@@ -0,0 +1,28 @@
...
+    [
+      extra_applications: [:logger]
+    ]
+  end
+
+  # Run "mix help deps" to learn about dependencies.
+  defp deps do
+    [
+      # {:dep_from_hexpm, "~> 0.3.0"},
+      # {:dep_from_git, git: "https://github.com/elixir-lang/my_dep.git", tag: "0.1.0"}
```
scohen, line 25: delete?

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -0,0 +1,318 @@
...
+
+    {text_before_range, remainder} = Enum.split(graphemes, start_column)
+    {range_text, text_after_range} = Enum.split(remainder, end_column - start_column)
+    {text_before_range, range_text, text_after_range}
+  end
+
+  defp mark_range(range_text) do
+    start_symbol = ~s[«]
+    end_symbol = ~s[»]
+    [start_symbol | range_text ++ List.wrap(end_symbol)]
```
scohen, line 316: make the symbols strings and use iolists:
```elixir
@range_start_marker "«"
@range_end_marker "»"

def mark_range(range_text) do 
  [@range_start_marker, range_text, @range_end_marker]
end

File: apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs [hunk 22]
```diff
@@ -0,0 +1,318 @@
...
+      {:ok, source_file.uri, definition_line}
+    end
+  end
+
+  defp definition_line(source_file, range) do
+    with {:ok, line_text} <- SourceFile.fetch_text_at(source_file, range.start.line) do
+      graphemes = String.graphemes(line_text)
+      {text_before_range, range_text, text_after_range} = extract_range(graphemes, range)
+      marked_text = mark_range(range_text)
+      {:ok, IO.iodata_to_binary(text_before_range ++ marked_text ++ text_after_range)}
```
scohen, line 300: use IOlists properly.
```elixir
marked_text = `IO.iodata_to_binary([text_before_range, mark_range(range_text), text_after_range])
{:ok, marked_text}
```
You don't need to build a single list (and you can use improper lists too). Concatenating lists together with `++` removes all the performance advantages to IOlists.

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -0,0 +1,318 @@
...
+      uses_content = ~q[
+        String.to_integer|("1")
+      ]
+
+      # credo:disable-for-next-line Credo.Check.Design.TagTODO
+      # TODO: this should be fixed when we have a call tracer
+      {:ok, nil} = definition(project, uses_content)
+    end
+
+    test "find the definition when call a erlang module", %{project: project} do
```
scohen, line 268: nit: "when calling"

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
scohen: i still think the `uses_file` is very awkward naming. I'd prefer `subject_module` something like that. Does that file need to actually exist? you're just replacing the file's contents with whatever is under test (which is fine). If you need a source file, just make one and pass that around.

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -0,0 +1,318 @@
...
+    on_exit(fn ->
+      :ok = RemoteControl.call(project, SourceFile.Store, :close, [uri])
+    end)
+  end
+
+  describe "definition/2 when making remote call by alias" do
+    setup [:with_referenced_file]
+
+    test "find the definition of a remote function call", %{project: project, uri: referenced_uri} do
+      uses_content = ~q[
```
scohen, line 62: nit: `subject_module`

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -0,0 +1,318 @@
...
+      :ok = RemoteControl.stop(project)
+    end)
+
+    RemoteControl.Api.schedule_compile(project, true)
+    assert_receive project_compiled(), 5000
+
+    %{project: project}
+  end
+
+  defp with_referenced_file(%{project: project}, file \\ "my_definition.ex") do
```
scohen, line 28: you are never using the optional param here.

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -0,0 +1,318 @@
...
+            MyDefinition.print_hello|()
+          end
+        end
+        ]
+
+      assert {:ok, ^referenced_uri, definition_line} = definition(project, uses_content)
+      assert definition_line == ~S[  defmacro «print_hello» do]
+    end
+
+    test "it can't find the right arity function definition", %{project: project} do
```
scohen, line 106: make a comment here explaining the issue. You should consider skipping this test rather than using a TODO.
TODOs almost never get done, but an asterisk in our tests tell us that something is not quite right.

scohen (approved):

PR #85, opened 2023-04-03
Added unit tests

scohen:
Skipped most of the conversions test until I reinstate the elixir <-> lsp conversions

### DIFF
--- a/apps/common/test/lexical/line_parser_test.exs
+++ b/apps/common/test/lexical/line_parser_test.exs
@@ -1,0 +1,102 @@
+defmodule Lexical.SourceFile.LineParserTest do
+  alias Lexical.SourceFile.LineParser
+
+  import Lexical.SourceFile.Line
+  use ExUnit.Case
+  use ExUnitProperties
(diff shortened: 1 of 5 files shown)


apps/common/test/lexical/line_parser_test.exs
```diff
@@ -0,0 +1,102 @@
...
+
+  import Lexical.SourceFile.Line
+  use ExUnit.Case
+  use ExUnitProperties
+
+  test "parsing a single \n" do
+    assert length(LineParser.parse("\n", 1)) == 1
+  end
+
+  def parse(lines) do
```
scottming, line 12: I noticed that you often use 'public def' instead of 'defp' in tests. I'm curious about the reason behind this choice.

scohen: I don't give it much thought tbh. They're unit tests, they're supposed to be simple and dumb so I use def. 
One thing to consider is that setup and callback functions need to be def.

PR #86, opened 2023-04-04
Converted range-able response types

scohen:
I was noticing a lot of manual conversions in the code, and that kind of stinks. To help with this, I'm going to change how things work in the project, but this requires some additions to the protocol.

The overarching idea is to push conversions into the protocol, so that on the way in, we convert from LSP structures to native and on the way out we convert from native to LSP structures. Presently, there was no conversion on the way out (there is conversion on the way in), so this PR adds conversion when we write data.

The next step will be to change our native structures to be one-based (thanks, Elixir, for not making that configurable) so all we need to deal with in the language server are our data structures.

### DIFF
--- a/apps/proto/lib/lexical/protocol/proto/convert.ex
+++ b/apps/proto/lib/lexical/protocol/proto/convert.ex
@@ -3,7 +3,7 @@ defmodule Lexical.Protocol.Proto.Convert do
   alias Lexical.SourceFile
 
   def to_elixir(%{lsp: lsp_request} = request) do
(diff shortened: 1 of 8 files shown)


Sleepful (approved):
I was looking really hard for something that said `- 1` or `+ 1`, is that kind of translation present in this PR?

scohen:
you see it a lot in the interfaces between the LSP and elixir. Look at #84 and you'll see what I mean.

scottming (approved):

PR #87, opened 2023-04-05
Converted positions to be one-based

### LINKED ISSUE #81: Protocol improvements
When I started lexical, I thought that I could eliminate the conversion between elixir's positions (one-based everything) and the LSP's positions (zero-based everything) by setting compiler options. However this turned out to be a documentation issue; starting an elixir file on line 0 and column 0 is not possible consistently. 

Knowing this, we should give up on this approach and instead work on handling these conversions at the protocol layer. There, when we read data, we'll add 1 to positions and on the way out, we'll subtract one. 

The goal here is to not have `character = pos.character + 1` in the user-land code.

scohen:
When I designed lexical, I thought it would be good to have zero-based positions like the LS protocol and elixir's conventions. I thought that getting the compiler to emit zero-based messages would be as simple as adding a compiler setting. However, it's not so easy. In fact, it's impossible at the present moment.

So, we can see the code is starting to get a lot of `line + 1` and `line - 1` kind of code, which stinks.

Combined with the last commit, this changes source files, code units, positions and ranges to be one-based, and we convert from the zero-based language server at the boundaries. Now, the only code that adds or subtracts will be string manipulation code, which we'll likely put behind some kind of interface.

Get ready for a series of off-by-one errors

Fixes #81

### DIFF
--- a/apps/common_protocol/lib/lexical/protocol/conversions.ex
+++ b/apps/common_protocol/lib/lexical/protocol/conversions.ex
@@ -8,6 +8,7 @@ defmodule Lexical.Protocol.Conversions do
   are the same in both utf-8 and utf-16, since they reference characters and not bytes.
   """
   alias Lexical.CodeUnit
+  alias Lexical.Math
   alias Lexical.Protocol.Types.Position, as: LSPosition
   alias Lexical.Protocol.Types.Range, as: LSRange
   alias Lexical.SourceFile
@@ -18,10 +19,7 @@ defmodule Lexical.Protocol.Conversions do
 
   import Line
 
-  def to_elixir(
-        %LSRange{} = ls_range,
-        %SourceFile{} = source
-      ) do
+  def to_elixir(%LSRange{} = ls_range, %SourceFile{} = source) do
     with {:ok, start_pos} <- to_elixir(ls_range.start, source.document),
          {:ok, end_pos} <- to_elixir(ls_range.end, source.document) do
       {:ok, %ElixirRange{start: start_pos, end: end_pos}}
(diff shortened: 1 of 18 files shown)


apps/common_protocol/lib/lexical/protocol/conversions.ex
```diff
@@ -8,6 +8,7 @@ defmodule Lexical.Protocol.Conversions do
   are the same in both utf-8 and utf-16, since they reference characters and not bytes.
   """
   alias Lexical.CodeUnit
+  alias Lexical.Math
```
scottming, line 11: Unused.

apps/server/lib/lexical/server/project/diagnostics.ex
```diff
@@ -133,20 +135,23 @@ defmodule Lexical.Server.Project.Diagnostics do
...
     defp position_to_range(%SourceFile{} = source_file, {line_number, column}) do
-      line_number = Math.clamp(line_number - 1, 0, SourceFile.size(source_file) - 1)
-      column = max(column - 1, 0)
-
-      with {:ok, line_text} <- SourceFile.fetch_text_at(source_file, line_number),
-           {:ok, character} <- CodeUnit.to_utf16(line_text, column) do
-        Range.new(
-          start: Position.new(line: line_number, character: character),
-          end: Position.new(line: line_number + 1, character: 0)
+      line_number = Math.clamp(line_number, 1, SourceFile.size(source_file) + 1)
```
scottming, line 138: why changed `SourceFile.size(source_file) - 1` to `SourceFile.size(source_file) + 1)`?

We need to ensure that the line_number does not exceed the actual maximum number of lines, so we need to subtract 1.

apps/server/test/support/cursor_support.ex
```diff
@@ -51,4 +51,12 @@ defmodule Lexical.Test.CursorSupport do
...
     |> IO.iodata_to_binary()
   end
+
+  defp starting_line do
+    1
+  end
+
+  defp starting_column do
+    1
+  end
```
Sleepful, line 61: clear naming 👍

Sleepful (approved):
all of these numbers are making me dizzy, is this what it feels like to be an accountant? 

I have a feeling that having `1`s and `0`s on `remote_control` and `server` is evidence of leaky abstraction that should not leave the `common` and `protocol` apps. Tell me what you think!

apps/server/test/support/cursor_support.ex
Sleepful: I would not dislike it at all if this approach was taken for most `1`s and `0`s across the codebase

scohen:
It’s not a leaky abstraction, the lsp defines positions as being zero based, elixir compilation errors are one~based, but elixir strong operations are zero based. The problem isn’t a leaky abstraction so much as a lack of an abstraction

apps/common_protocol/lib/lexical/protocol/conversions.ex
scohen: Strange, lexical didn't highlight this.

PR #88, opened 2023-04-05
Use multiple diagnostics instead of one

### LINKED ISSUE #25: Should generate two diagnostics for `missing terminator` error
```elixir
defmodule Dummy do
  @moduledoc """
  Documentation for `Dummy`.
  """

  @doc """
  Hello world.

  ## Examples

      iex> Dummy.hello()
      :world

  """
  def hello do
    :world
  end

  def foo(foo) do
    .
  end
end
```

```bash
$ mix compile
Compiling 1 file (.ex)

== Compilation error in file lib/dummy.ex ==
** (TokenMissingError) lib/dummy.ex:23:1: missing terminator: end (for "do" starting a
t line 1)
    (elixir 1.14.0) lib/kernel/parallel_compiler.ex:346: anonymous fn/5 in Kernel.Para
llelCompiler.spawn_workers/7
```

scottming:
Hint is not always available, but when there is a hint, its information is more precise than the context. Currently, when there is a hint, there will be three diagnostics provided. However, I prefer to have only **one diagnostic**.

for example:

```elixir
defmodule ElixirNav do
  def foo_middle() do
  end






  def foo_end() do

  def foo() do
  end
end
```

image

The diagnostic of line 10 is enough.

### another change

Also, we cannot use the method of adding 1 to highlight the entire line. This can easily cause the range to exceed the total number of lines in the code and highlight the whole file. Instead, we should use:

```
      elixir_range =
        ExRange.new(
          ExPosition.new(line_number, column),
          ExPosition.new(min(line_number + 1, file_size), 1)
        )
```
This is the preview in VScode.(I verified this in vscode and neovim.)

image

And this is the preview of this previous code

```elixir
          ExPosition.new(line_number, column),
          ExPosition.new(line_number + 1, 1)
```

image

Fixes #25

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/build/error.ex
+++ b/apps/remote_control/lib/lexical/remote_control/build/error.ex
@@ -16,15 +16,65 @@ defmodule Lexical.RemoteControl.Build.Error do
   end
 
   def parse_error_to_diagnostics(context, message_info, token) do
-    diagnostic = %Diagnostic{
-      file: nil,
-      severity: :error,
-      position: context_to_position(context),
-      compiler_name: "Elixir",
-      message: message_info_to_binary(message_info, token)
-    }
+    parse_error_diagnostic_functions = [
+      &build_end_line_diagnostics/3,
+      &build_start_line_diagnostics/3,
+      &build_hint_diagnostics/3
+    ]
 
-    [diagnostic]
+    Enum.flat_map(parse_error_diagnostic_functions, & &1.(context, message_info, token))
+  end
+
+  defp build_end_line_diagnostics(context, message_info, token) do
+    [end_line_message | _] = String.split(message_info, "\n")
+
+    [
+      %Diagnostic{
+        file: nil,
+        severity: :error,
+        position: context_to_position(context),
+        compiler_name: "Elixir",
+        message: "#{end_line_message}#{token}"
+      }
+    ]
+  end
+
+  @start_line_regex ~r/(\w+) \(for (.*) starting at line (\d+)\)/
+  defp build_start_line_diagnostics(_context, message_info, _token) do
+    case Regex.run(@start_line_regex, message_info) do
+      [_, missing, token, start_line] ->
+        diagnostic = %Diagnostic{
+          file: nil,
+          severity: :error,
+          position: String.to_integer(start_line),
+          compiler_name: "Elixir",
+          message: "The #{token} here is missing a terminator: #{inspect(missing)}"
+        }
+
+        [diagnostic]
+
+      _ ->
+        []
+    end
+  end
+
+  @hint_regex ~r/HINT: .*on line (\d+).*/m
+  defp build_hint_diagnostics(_context, message_info, _token) do
+    case Regex.run(@hint_regex, message_info) do
+      [message, hint_line] ->
+        diagnostic = %Diagnostic{
+          file: nil,
+          severity: :error,
+          position: String.to_integer(hint_line),
+          compiler_name: "Elixir",
+          message: String.replace(message, ~r/on line \d+/, "here")
+        }
+
+        [diagnostic]
+
+      _ ->
+        []
+    end
   end
 
   def error_to_diagnostic(%CompileError{} = compile_error, _stack, _quoted_ast) do
--- a/apps/remote_control/lib/lexical/remote_control/build/error.ex
+++ b/apps/remote_control/lib/lexical/remote_control/build/error.ex
@@ -12,7 +12,7 @@ defmodule Lexical.RemoteControl.Build.Error do
     detail_diagnostics = detail_diagnostics(detail)
     error = message_info_to_binary(message_info, token)
     error_diagnostics = parse_error_to_diagnostics(context, error, token)
-    uniq(error_diagnostics ++ detail_diagnostics)
+    extract_line(error_diagnostics ++ detail_diagnostics)
   end
 
   def parse_error_to_diagnostics(context, message_info, token) do
@@ -25,18 +25,15 @@ defmodule Lexical.RemoteControl.Build.Error do
     Enum.flat_map(parse_error_diagnostic_functions, & &1.(context, message_info, token))
   end
 
-  defp uniq(diagnostics) do
+  defp extract_line(diagnostics) do
     # We need to uniq by position because the same position can be reported
     # and the `end_line_diagnostic` is always the precise one
     position_key = fn
-      %Diagnostic{position: position} when is_tuple(position) ->
-        elem(position, 0)
-
-      %Diagnostic{position: position} ->
-        position
+      %Diagnostic{position: {line, _column}} -> line
+      %Diagnostic{position: line} -> line
     end
 
-    Enum.uniq_by(diagnostics, &position_key.(&1))
+    Enum.uniq_by(diagnostics, position_key)
   end
 
   defp build_end_line_diagnostics(context, message_info, token) do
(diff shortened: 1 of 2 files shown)


scohen:
> However, I prefer to have only one diagnostic.

I prefer to have several, and here's why. 
Imagine you're editing a 300 line file, and you introduce an error in the middle. With a single diagnostic, it's very likely that you'll miss the tiny red squiggly lines somewhere in the middle of the file, but with diagnostics surrounding it, it's much more clear --especially if you're looking at the beginning or the end.

scohen:
> Also, we cannot use the method of adding 1 to highlight the entire line. This can easily cause the range to exceed the total number of lines in the code and highlight the whole file. Instead, we should use:

Your reasoning is incorrect; it is absolutely allowable to exceed the number of lines in the file, the problem isn't that this happens (and it absolutely will happen; elixir does this by itself, so we need to account for it) it's that the error handling case explicitly highlights the whole file:

If you want to fix that, then I'm on board.
The code below highlights the whole file
```elixir
        _error ->
          Range.new(
            start: Position.new(line: 0, character: 0),
            end: Position.new(line: line_number, character: 0)
          )
```

If you change it to this, things work better
```elixir
        _error ->
          Range.new(
            start: Position.new(line: line_number, character: 0),
            end: Position.new(line: line_number + 1, character: 0)
          )
```

apps/remote_control/lib/lexical/remote_control/build/error.ex
```diff
@@ -16,15 +16,74 @@ defmodule Lexical.RemoteControl.Build.Error do
   end
 
   def parse_error_to_diagnostics(context, message_info, token) do
-    diagnostic = %Diagnostic{
+    if missing_terminator?(message_info) do
```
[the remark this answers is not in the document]
scohen: In other words, why do we need this if? Why can't we just do what `build_missing_terminator_diagnostics` does?
```elixir
parse_error_diagnostic_functions = [&build_end_line_diagnostic/3, &build_start_line_diagnostic/3, &build_hint_diagnostic/3, &build_fallback_diagnostic/3]

Enum.flat_map(parse_error_diagnostic_functions, & &1.(context, message_info, token)
```
The only change is that each function needs to take 3 args and returns an empty list if it can't make a diagnostic.

apps/remote_control/lib/lexical/remote_control/build/error.ex
```diff
@@ -16,15 +16,74 @@ defmodule Lexical.RemoteControl.Build.Error do
...
+      severity: :error,
+      position: String.to_integer(start_line),
+      compiler_name: "Elixir",
+      message: "The #{token} here is missing a terminator: #{inspect(missing)}"
     }
+  end
+
+  @hint_regex ~r/HINT: .*on line (\d+).*/m
+  defp build_hint_diagnostic(message_info) do
+    hint_result = Regex.run(@hint_regex, message_info)
```
scohen, line 70: I tihnk this would be better represented via a case
```elixir 
case Regex.run(@hint_regex, message_info) do 
  [message, hint_line] -> 
      ...
      
  _ -> nil
end
```

apps/remote_control/lib/lexical/remote_control/build/error.ex
```diff
@@ -16,15 +16,65 @@ defmodule Lexical.RemoteControl.Build.Error do
...
 
   def parse_error_to_diagnostics(context, message_info, token) do
-    diagnostic = %Diagnostic{
-      file: nil,
-      severity: :error,
-      position: context_to_position(context),
-      compiler_name: "Elixir",
-      message: message_info_to_binary(message_info, token)
-    }
+    parse_error_diagnostic_functions = [
```
scohen, line 19: much nicer

apps/remote_control/lib/lexical/remote_control/build/error.ex
```diff
@@ -12,19 +12,83 @@ defmodule Lexical.RemoteControl.Build.Error do
...
+    # and the `end_line_diagnostic` is always the precise one
+    position_key = fn
+      %Diagnostic{position: position} when is_tuple(position) ->
+        elem(position, 0)
+
+      %Diagnostic{position: position} ->
+        position
+    end
+
+    Enum.uniq_by(diagnostics, &position_key.(&1))
```
scohen, line 39: You can just pass `position_key` to unique_by without capturing

apps/remote_control/lib/lexical/remote_control/build/error.ex
```diff
@@ -12,19 +12,83 @@ defmodule Lexical.RemoteControl.Build.Error do
...
 
-    [diagnostic]
+    Enum.flat_map(parse_error_diagnostic_functions, & &1.(context, message_info, token))
+  end
+
+  defp uniq(diagnostics) do
+    # We need to uniq by position because the same position can be reported
+    # and the `end_line_diagnostic` is always the precise one
+    position_key = fn
+      %Diagnostic{position: position} when is_tuple(position) ->
```
scohen, line 32: Technically, this means you will eliminate multiple issues on the same line. Not sure if the compiler can handle that, but calling it out nonetheless.

scohen: Also, why not just match on {line, _column}?

apps/remote_control/lib/lexical/remote_control/build/error.ex
```diff
@@ -12,19 +12,83 @@ defmodule Lexical.RemoteControl.Build.Error do
...
+        elem(position, 0)
+
+      %Diagnostic{position: position} ->
```
scohen, line 35: position: line

apps/remote_control/lib/lexical/remote_control/build/error.ex
```diff
@@ -12,19 +12,83 @@ defmodule Lexical.RemoteControl.Build.Error do
...
+        position
```
scohen, line 36: line

apps/remote_control/lib/lexical/remote_control/build/error.ex
```diff
@@ -12,19 +12,83 @@ defmodule Lexical.RemoteControl.Build.Error do
...
+    ]
 
-    [diagnostic]
+    Enum.flat_map(parse_error_diagnostic_functions, & &1.(context, message_info, token))
+  end
+
+  defp uniq(diagnostics) do
+    # We need to uniq by position because the same position can be reported
+    # and the `end_line_diagnostic` is always the precise one
+    position_key = fn
```
scohen, line 31: This function should be called `extract_line`

apps/remote_control/lib/lexical/remote_control/build/error.ex
scottming: > Technically, this means you will eliminate multiple issues on the same line. Not sure if the compiler can handle that, but calling it out nonetheless.


Yes, but what I eliminated was the information generated by ourselves. In fact, there is only one piece of original information provided by the compiler.

apps/remote_control/lib/lexical/remote_control/build/error.ex
```diff
@@ -12,19 +12,80 @@ defmodule Lexical.RemoteControl.Build.Error do
...
+      &build_hint_diagnostics/3
+    ]
+
+    Enum.flat_map(parse_error_diagnostic_functions, & &1.(context, message_info, token))
+  end
+
+  defp extract_line(diagnostics) do
+    # We need to uniq by position because the same position can be reported
+    # and the `end_line_diagnostic` is always the precise one
+    position_key = fn
```
scohen, line 31: nit: name this `extract_line`
Also, don't mark things resolved if they're not resolved, it wastes time.

scottming: Sorry, I misunderstood the number of lines you referred to earlier. I changed, and I think It's all good now.

scohen (approved):

Issue #44, opened 2023-03-20
Time out (crash) for code-format when using `lexical` as project context


Sleepful:
example, using `method = "textDocument/formatting",` in file `lexical/apps/common/test/support/range.ex` gives me a `Timeout` warning message on editor when `/Users/jose/Code/lexical` is used as root.

Client's log:

```
[DEBUG][2023-03-20 03:36:29] .../lua/vim/lsp.lua:1397	"LSP[lexical]"	"client.request"	1	"textDocument/formatting"	{  options = {    insertSpaces = true,    tabSize = 2  },  textDocument = {    uri = "file:///Users/jose/Code/lexical/apps/common/test/support/range.ex"  }}	<function 1>	161
[DEBUG][2023-03-20 03:36:29] .../vim/lsp/rpc.lua:284	"rpc.send"	{  id = 7,  jsonrpc = "2.0",  method = "textDocument/formatting",  params = {    options = {      insertSpaces = true,      tabSize = 2    },    textDocument = {      uri = "file:///Users/jose/Code/lexical/apps/common/test/support/range.ex"    }  }}
[DEBUG][2023-03-20 03:36:30] .../vim/lsp/rpc.lua:284	"rpc.send"	{  jsonrpc = "2.0",  method = "$/cancelRequest",  params = {    id = 7  }}
```

In a smaller project I can use `textDocument/formatting` without issue.

PR #89, opened 2023-04-05
Improve namespace for .app files & its references.

Sleepful:
This should be it for namespacing everything that requires namespacing.

Just added some sneaky edge cases that were not covered by previous code, the general scheme of the task stays very similar.

This fixes diagnostics that look like:

```ex
 %Mix.Task.Compiler.Diagnostic{file: "/Users/steve/Projects/lexical/apps/server/lib/lexical/server/project/diagnostics.ex", severity: :warning, message: "Lexical.SourceFile.Store.open?/1 defined in application :common is used by the current application but the current application does not depend on :common. To fix this, you must do one of:\n\n  1. If :common is part of Erlang/Elixir, you must include it under :extra_applications inside \"def application\" in your mix.exs\n\n  2. If :common is a dependency, make sure it is listed under \"def deps\" in your mix.exs\n\n  3. In case you don't want to add a requirement to :common, you may optionally skip this warning by adding [xref: [exclude: [Lexical.SourceFile.Store]]] to your \"def project\" in mix.exs\n", position: 36, compiler_name: "Elixir", details: nil}
```

- Closes https://github.com/lexical-lsp/lexical/issues/44 
- Removes the `unless` special case here, doesn't seem to be a problem anymore: https://github.com/lexical-lsp/lexical/commit/6ab6efd2601a04a88b14e430a82108d770a054bd#diff-64a6279b1684f2330338b7564bdcb9506b6bfcf3aba3f42b9461a9af889da367R22-R27

### DIFF
--- a/apps/proto/lib/mix/tasks/namespace/release.ex
+++ b/apps/proto/lib/mix/tasks/namespace/release.ex
@@ -3,44 +3,54 @@ defmodule Mix.Tasks.Namespace.Release do
   `remote_control.app` must be namespaced too
   """
   use Mix.Task
-  @release_files ~w(start.script start_clean.script lexical.rel)
-  @apps_to_rewrite ~w(remote_control)
+  @release_files ~w(start.script lexical.rel)
+  @apps_to_rewrite ~w(remote_control common common_protocol server protocol proto)
 
-  def run([release_path, release_version_path]) do
-    Enum.map(@apps_to_rewrite, &update_app(release_path, release_version_path, &1))
+  def run(_) do
+    release = Mix.Release.from_config!(:lexical, Mix.Project.config(), [])
+    Enum.map(@apps_to_rewrite, &update_app(release.path, release.version_path, &1))
+    # namespace .app filenames because the filename is used as identifier by BEAM
+    Enum.map(@apps_to_rewrite, &namespace_app_file(release.path, &1))
+    Mix.Shell.IO.info("\nApplied namespace to release app.")
   end
 
-  defp update_app(release_path, release_version_path, app_name) do
-    app_file_name = "#{app_name}.app"
-
+  defp path(:ebin, release_path, app_name) do
     [ebin_path] =
       [release_path, "lib", "#{app_name}-*", "ebin"]
       |> Path.join()
       |> Path.wildcard()
 
-    app_file_path = Path.join([ebin_path, app_file_name])
+    ebin_path
+  end
+
+  defp path(:app_file, release_path, app_name) do
+    Path.join([path(:ebin, release_path, app_name), "#{app_name}.app"])
+  end
+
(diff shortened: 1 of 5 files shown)


apps/proto/lib/mix/tasks/namespace/release.ex
```diff
@@ -3,44 +3,54 @@ defmodule Mix.Tasks.Namespace.Release do
...
       |> Path.wildcard()
 
-    app_file_path = Path.join([ebin_path, app_file_name])
-    release_file_paths = Enum.map(@release_files, &Path.join([release_version_path, &1]))
-    paths = [app_file_path | release_file_paths]
+    ebin_path
+  end
 
-    Enum.each(paths, &update_file_contents(&1, "remote_control"))
+  defp path(:app_file, release_path, app_name) do
```
scohen, line 26: slight preference to naming these functions `app_file_path` and `ebin_path`

apps/remote_control/lib/lexical/remote_control/build.ex
```diff
@@ -16,10 +16,7 @@ defmodule Lexical.RemoteControl.Build do
   end
 
   def compile_source_file(%Project{} = project, %SourceFile{} = source_file) do
-    # Don't format mix.exs files until we come up with with a way to prevent them from causing errors
-    unless Path.absname(source_file.path) == "mix.exs" do
-      RemoteControl.call(project, GenServer, :cast, [__MODULE__, {:compile_file, source_file}])
-    end
+    RemoteControl.call(project, GenServer, :cast, [__MODULE__, {:compile_file, source_file}])
```
scohen, line 19: did you fix this? Prior, this broke pretty hard when you'd edit the `mix.exs` in the root directory, and it applied to every `mix.exs`, not just lexical's.

Sleepful: hm, looks like this was already fixed and unrelated to this PR. Removing this `unless` from the `main` branch doesn't cause any issues when formatting `mix.exs` files.

Ok to leave in the PR?

mix.exs
```diff
@@ -94,7 +94,7 @@ defmodule Lexical.MixProject do
 
   defp maybe_namespace_release(%Mix.Release{} = release) do
     if System.get_env("NAMESPACE") do
-      Mix.Task.run("namespace.release", [release.path, release.version_path])
+      Mix.Task.run("namespace.release")
```
scohen, line 97: nice, passing the arguments was a little janky

apps/remote_control/lib/lexical/remote_control/build.ex
scohen: sure, we can always add it back.

scohen (approved):

PR #99, opened 2023-04-07
Fix code actions line number issue

scottming:
```elixir
defmodule MyModule do
  def foo(foo) do
  end
end
```

The line_number extracted [here](https://github.com/lexical-lsp/lexical/blob/main/apps/server/lib/lexical/server/provider/code_action/replace_with_underscore.ex#L21) is `2`, but the lsp is 0-based.


scohen:
this is an example of the `n + 1`, `n - 1` thing I was complaining about in my previous refactor. I think the correct fix here is to use `Lexical.Position` structs in the TextEdit, and they'll be converted during encoding.

scottming:
in favor of #109

PR #102, opened 2023-04-10
Enhance the diagnostic column

scottming:
### before

image

image

### After

image

image

The cursor color and highlighting(no highlighting spaces) look better, and we can reach the exact place we want to modify with fewer taps w when using something like: `Lspsaga diagnostic_jump_next`.

### DIFF
--- a/apps/common/lib/lexical/math.ex
+++ b/apps/common/lib/lexical/math.ex
@@ -2,11 +2,4 @@ defmodule Lexical.Math do
   def clamp(number, min_number, max_number) do
     number |> max(min_number) |> min(max_number)
   end
-
-  def count_leading_spaces(str), do: count_leading_spaces(str, 0)
-
-  def count_leading_spaces(<<c, rest::binary>>, count) when c in [?\s, ?\t],
-    do: count_leading_spaces(rest, count + 1)
-
-  def count_leading_spaces(_, count), do: count
 end
--- a/apps/common/lib/lexical/math.ex
+++ b/apps/common/lib/lexical/math.ex
@@ -2,4 +2,11 @@ defmodule Lexical.Math do
   def clamp(number, min_number, max_number) do
     number |> max(min_number) |> min(max_number)
   end
+
+  def count_leading_spaces(str), do: count_leading_spaces(str, 0)
+
+  def count_leading_spaces(<<c::utf8, rest::binary>>, count) when c in [?\s, ?\t],
+    do: count_leading_spaces(rest, count + 1)
+
+  def count_leading_spaces(_, count), do: count
 end
--- a/apps/common/lib/lexical/math.ex
+++ b/apps/common/lib/lexical/math.ex
@@ -5,7 +5,7 @@ defmodule Lexical.Math do
 
   def count_leading_spaces(str), do: count_leading_spaces(str, 0)
 
-  def count_leading_spaces(<<c::utf8, rest::binary>>, count) when c in [?\s, ?\t],
+  def count_leading_spaces(<<c, rest::binary>>, count) when c in [?\s, ?\t],
     do: count_leading_spaces(rest, count + 1)
 
   def count_leading_spaces(_, count), do: count
--- a/apps/common/test/lexical/math_test.exs
+++ b/apps/common/test/lexical/math_test.exs
@@ -40,7 +40,7 @@ defmodule Lexical.MathTest do
     end
   end
 
-  defp non_leading_space_string() do
+  defp non_leading_space_string do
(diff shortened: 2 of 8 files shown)


apps/common/lib/lexical/math.ex
```diff
@@ -2,4 +2,11 @@ defmodule Lexical.Math do
   def clamp(number, min_number, max_number) do
     number |> max(min_number) |> min(max_number)
   end
+
+  def count_leading_spaces(str), do: count_leading_spaces(str, 0)
+
+  def count_leading_spaces(<<c::utf8, rest::binary>>, count) when c in [?\s, ?\t],
```
scohen, line 8: I don't think you need the utf8 specificer here, but it's not hurting things.

apps/common/lib/lexical/math.ex
```diff
@@ -2,4 +2,11 @@ defmodule Lexical.Math do
   def clamp(number, min_number, max_number) do
     number |> max(min_number) |> min(max_number)
   end
+
+  def count_leading_spaces(str), do: count_leading_spaces(str, 0)
```
scohen, line 6: This function doesn't seem like it belongs in the `Math` module, as it's more of a String thing.

apps/common/test/lexical/math_test.exs
```diff
@@ -28,4 +28,22 @@ defmodule Lexical.MathTest do
...
+      assert Math.count_leading_spaces("hello") == 0
+    end
+
+    test "returns 2 when string has two leading spaces" do
+      assert Math.count_leading_spaces("  hello") == 2
+    end
+
+    test "return 1 when string has one leading tab" do
+      assert Math.count_leading_spaces("\thello") == 1
+    end
```
scohen, line 47: i feel like property tests would be appropriate here.

apps/server/lib/lexical/server/project/diagnostics.ex
```diff
@@ -137,11 +137,23 @@ defmodule Lexical.Server.Project.Diagnostics do
...
+      end
+    end
+
+    defp to_lsp_range(line, column, source_file) do
       elixir_range =
         ExRange.new(
-          ExPosition.new(line_number, column),
-          ExPosition.new(line_number + 1, 1)
+          ExPosition.new(line, column),
+          ExPosition.new(line + 1, 1)
```
scohen, line 156: line_number is more descriptive; In this codebase, the word `line` refers to a specific data structure.

apps/server/lib/lexical/server/project/diagnostics.ex
```diff
@@ -150,18 +162,11 @@ defmodule Lexical.Server.Project.Diagnostics do
 
         _error ->
           Range.new(
-            start: Position.new(line: line_number, character: 0),
-            end: Position.new(line: line_number + 1, character: 0)
+            start: Position.new(line: line, character: 0),
+            end: Position.new(line: line + 1, character: 0)
```
scohen, line 166: same comment; change back to `line_number`

apps/common/lib/lexical/math.ex
scottming: In your opinion, which file in the 'common' folder do you think is more suitable for placing this `count_leading_spaces` function?

apps/common/test/lexical/math_test.exs
scottming: Added one via e767c11b7c6fb7798e8c9788830b3c741aef4b38

apps/common/lib/lexical/math.ex
scottming: Ok, I moved to the `Lexical.Text`

apps/server/lib/lexical/server/project/diagnostics.ex
```diff
@@ -137,7 +138,19 @@ defmodule Lexical.Server.Project.Diagnostics do
...
+    defp position_to_range(source_file, line_number) when is_integer(line_number) do
+      line_number = Math.clamp(line_number, 1, SourceFile.size(source_file))
+
+      with {:ok, line_text} <- SourceFile.fetch_text_at(source_file, line_number) do
+        column = Text.count_leading_spaces(line_text) + 1
+        to_lsp_range(line_number, column, source_file)
+      end
+    end
 
+    defp to_lsp_range(line_number, column, source_file) do
```
scohen, line 153: what happens if we don't make this an LSP range, and instead use a Lexical range?

scohen (approved):
try my suggestion and then merge.

apps/server/lib/lexical/server/project/diagnostics.ex
scottming: #103

PR #104, opened 2023-04-18
Moved store to ets

scohen:
While refactoring conversions to make them easier, I was getting deadlocks during get_and_update because the code neeed the context document in order to convert from lsp to native. However, it was inside the genserver process when it called `fetch`, which caused a deadlock. Moving read operations out of the genserver will fix this deadlock.

This commit moves all store read operations into an ETS table that has read concurrency, which should also speed up reads (though there is a small chance of getting stale data now).

### DIFF
--- a/apps/common/lib/lexical/source_file/store.ex
+++ b/apps/common/lib/lexical/source_file/store.ex
@@ -1,29 +1,40 @@
 defmodule Lexical.SourceFile.Store do
+  @moduledoc """
+  A backing store for source file documents
+
+  This implementation stores documents in ETS, and partitions read and write operations. Read operations are served
+  immediately by querying the ETS table, while writes go through a GenServer process (which is the owner of the ETS table).
+  """
   defmodule State do
     alias Lexical.SourceFile
     alias Lexical.SourceFile.Store
     require Logger
 
-    defstruct source_files: %{}, temp_files: %{}, temporary_open_refs: %{}
+    defstruct temporary_open_refs: %{}
+
+    @table_name SourceFile.Store
+
     @type t :: %__MODULE__{}
     def new do
+      :ets.new(@table_name, [:named_table, :set, :protected, write_concurrency: true])
+
       %__MODULE__{}
     end
 
-    @spec fetch(t, Store.uri()) :: {:ok, t()} | {:error, :not_open}
-    def fetch(%__MODULE__{} = store, uri) do
-      with :error <- Map.fetch(store.source_files, uri),
-           :error <- Map.fetch(store.temp_files, uri) do
-        {:error, :not_open}
+    @spec fetch(Store.uri()) :: {:ok, t()} | {:error, :not_open}
+    def fetch(uri) do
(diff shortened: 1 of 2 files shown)


apps/common/lib/lexical/source_file/store.ex
```diff
@@ -1,29 +1,40 @@
...
     require Logger
 
-    defstruct source_files: %{}, temp_files: %{}, temporary_open_refs: %{}
+    defstruct temporary_open_refs: %{}
+
+    @table_name SourceFile.Store
+
     @type t :: %__MODULE__{}
     def new do
+      :ets.new(@table_name, [:named_table, :set, :protected, write_concurrency: true])
```
Tuxified, line 19: Should we also set `read_concurrency` to true? It's false by default, the docs seem to suggest that's a good thing if one expects a fair amount of reads? Disclaimer: I'm no ETS expert ;)

apps/common/lib/lexical/source_file/store.ex
```diff
@@ -145,6 +146,48 @@ defmodule Lexical.SourceFile.Store do
...
+    @write_types [:sources, :temp]
+    defp ets_fetch(type, key) when type in @read_types do
+      case :ets.match(@table_name, {key, type_selector(type), :"$1"}) do
+        [[value]] -> {:ok, value}
+        _ -> :error
+      end
+    end
+
+    defp ets_put(type, key, value) when type in @write_types do
+      :ets.insert(@table_name, {key, type, value})
```
Tuxified, line 160: All internal function accept `type, key` as the order of arguments, yet they're stored in `{key , type ..}` in ETS. Perhaps it would be better to use the same order? The info in ETS could also be used by other, for example plugins?

apps/common/lib/lexical/source_file/store.ex
```diff
@@ -145,6 +146,48 @@ defmodule Lexical.SourceFile.Store do
...
+           :ok <- ets_delete(type, key) do
+        value
+      else
+        _ ->
+          nil
+      end
+    end
+
+    defp ets_delete(type, key) when type in @write_types do
+      match_spec = {key, type_selector(type), :_}
```
Tuxified, line 184: I think the type_selector only makes sense to search for :sources, :temp or both, but I think we can skip using it when writing, WDYT?

Tuxified (approved):
Looks great, sorry if I'm nitpicking too much

apps/common/lib/lexical/source_file/store.ex
scohen: Definitely. This was a mistake

apps/common/lib/lexical/source_file/store.ex
scohen: I did that because of a micro optimization with pattern matching on the first argument. I typically like to make the first arg the one matched against. It’s likely not necessary here

scohen:
Thank you for the thorough review.

PR #105, opened 2023-04-19
Fix script file diagnostics flushing

scottming:
The compilation diagnostics of a test file only come from the compilation of the file. The project compilation (i.e., `MIX_ENV=test mix compile`) will not produce any diagnostics for the test file. Therefore, when you save a problematic test file, it may seem strange that after saving, the error disappears but reappears when you edit it.

This fix ensures that the diagnostics of test files will not be cleared when saved.

### DIFF
--- a/apps/server/lib/lexical/server/project/diagnostics.ex
+++ b/apps/server/lib/lexical/server/project/diagnostics.ex
@@ -38,7 +38,7 @@ defmodule Lexical.Server.Project.Diagnostics do
         Map.new(state.diagnostics_by_uri, fn {uri, diagnostics} ->
           with true <- SourceFile.Store.open?(uri),
                {:ok, %SourceFile{} = source_file} <- SourceFile.Store.fetch(uri),
-               true <- source_file.dirty? or test_file?(source_file) do
+               true <- source_file.dirty? or script_file?(source_file) do
             {uri, diagnostics}
           else
             _ ->
@@ -169,8 +169,10 @@ defmodule Lexical.Server.Project.Diagnostics do
       end
     end
 
-    defp test_file?(source_file) do
-      String.ends_with?(source_file.path, "_test.exs")
+    defp script_file?(source_file) do
+      # The script file refers to test files, or Ecto migration files, and so on.
+      # When running `MIX_ENV mix compile` to compile the project, they will not be compiled.
+      String.ends_with?(source_file.path, ".exs")
     end
   end
 
--- a/apps/server/lib/lexical/server/project/diagnostics.ex
+++ b/apps/server/lib/lexical/server/project/diagnostics.ex
@@ -38,7 +38,7 @@ defmodule Lexical.Server.Project.Diagnostics do
         Map.new(state.diagnostics_by_uri, fn {uri, diagnostics} ->
           with true <- SourceFile.Store.open?(uri),
                {:ok, %SourceFile{} = source_file} <- SourceFile.Store.fetch(uri),
-               true <- source_file.dirty? do
+               true <- source_file.dirty? or test_file?(source_file) do
             {uri, diagnostics}
           else
             _ ->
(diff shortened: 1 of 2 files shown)


apps/server/lib/lexical/server/project/diagnostics.ex
```diff
@@ -168,6 +168,10 @@ defmodule Lexical.Server.Project.Diagnostics do
           )
       end
     end
+
+    defp test_file?(source_file) do
+      String.ends_with?(source_file.path, "_test.exs")
```
Tuxified, line 173: Small correction/amendment: this also happens with others script files (ending in .exs). `mix compile` compiles all .ex files that are in the given path. `.exs` files are scripts and not compiled but run directly by the elixir interpreter, hence *I think* it's better to change this check. For example Phoenix/Ecto also use .exs files for the migration definitions (usually found in priv/repo/migrations)

Tuxified (approved):
It looks good for the intended purpose (don't flush diagnostics for test files), but I feel we can extend it for other .exs (script) files as well

apps/server/lib/lexical/server/project/diagnostics.ex
scottming: Thanks for the suggestions, will change this later

apps/server/lib/lexical/server/project/diagnostics.ex
```diff
@@ -38,7 +38,7 @@ defmodule Lexical.Server.Project.Diagnostics do
         Map.new(state.diagnostics_by_uri, fn {uri, diagnostics} ->
           with true <- SourceFile.Store.open?(uri),
                {:ok, %SourceFile{} = source_file} <- SourceFile.Store.fetch(uri),
-               true <- source_file.dirty? do
+               true <- source_file.dirty? or script_file?(source_file) do
```
scohen, line 41: i'd replace this with `flushable?(source_file)`

apps/server/lib/lexical/server/project/diagnostics.ex
```diff
@@ -168,6 +168,12 @@ defmodule Lexical.Server.Project.Diagnostics do
           )
       end
     end
+
+    defp script_file?(source_file) do
+      # The script file refers to test files, or Ecto migration files, and so on.
+      # When running `MIX_ENV mix compile` to compile the project, they will not be compiled.
+      String.ends_with?(source_file.path, ".exs")
```
scohen, line 175: use `Path.extname` here.

apps/server/test/lexical/server/project/diagnostics/state_test.exs
```diff
@@ -22,8 +22,8 @@ defmodule Lexical.Project.Diagnostics.StateTest do
     Path.join([Project.root_path(project()), "lib", "project.ex"])
   end
 
-  def source_file(contents \\ "") do
-    file_uri = SourceFile.Path.to_uri(existing_file_path())
+  def source_file(contents \\ "", file_path \\ existing_file_path()) do
```
scohen, line 25: i really dislike functions with multiple default arguments, but it turns out that we always set the contents, so you can remove that one and have the file_path be the default.

apps/server/test/lexical/server/project/diagnostics/state_test.exs
```diff
@@ -160,6 +160,19 @@ defmodule Lexical.Project.Diagnostics.StateTest do
       assert [_] = diagnostics
     end
 
+    test "it should not clear a script file even if it is clean", %{state: state} do
+      script_file_path = Path.join([Project.root_path(project()), "test", "*.exs"])
+      source_file = source_file("assert f() == 0", script_file_path)
+
+      {:ok, state} =
+        State.add(state, compiler_diagnostic(message: "undefined function f/0"), source_file.uri)
+
```
scohen, line 169: for a more rigorous test:
```elixir
diagnostic = compiler_diagnostic(message: "undefined function f/0", source_file.uri)
...
state = State.clear_all_flushed(state)
assert [^diagnostic] = State.get(state, source_file.uri)
```

scohen (approved):
Address the comments then you're good.

PR #106, opened 2023-04-19
Fix the typo of ets_put args

scottming:
We can't save the file now.🤣


Tuxified (approved):
Good catch!

scohen (approved):
Can we have a unit test for this?

Issue #103, opened 2023-04-14
Remove conversion in diagnostics.


scottming:
Like these lines:

https://github.com/lexical-lsp/lexical/blob/main/apps/server/lib/lexical/server/project/diagnostics.ex#L160-L170

We can do conversions in a unified place like `Response.ex`

PR #109, opened 2023-04-20
Ranged to convertible

scohen:
The first commit here moves the `proto` module from `Lexical.Protocol.Proto` to `Lexical.Proto`. The second commit is the more important one, as it removes the `Ranged` protocol in favor of `Convertible` and `DocumentContainer`. 

The `Convertible` implementations live in the `protocol` app in `lib/lexical/protocol/convertibles` They all have tests.


scohen:
Fixes 
#103 and #3

apps/common/lib/lexical/code_unit.ex
```diff
@@ -87,6 +91,19 @@ defmodule Lexical.CodeUnit do
...
+  defp do_count_utf8(<<c, rest::binary>>, count) when c < 128 do
+    do_count_utf8(rest, count + 1)
+  end
+
+  defp do_count_utf8(<<c::utf8, rest::binary>>, count) do
+    increment = byte_size(<<c::utf8>>)
+    do_count_utf8(rest, count + increment)
+  end
+
   defp do_utf16_offset(_, 0, offset) do
```
scottming, line 107: What is this for? 

is it some special cases that cannot be converted to UTF-8?

apps/protocol/lib/lexical/protocol/convertibles/mix.task.compiler.diagnostic.ex
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  defp to_lexical_range(%SourceFile{} = source_file, line_number, column) do
+    if line_number < SourceFile.size(source_file) do
+      Range.new(
+        Position.new(line_number, column),
+        Position.new(line_number + 1, 1)
+      )
+    else
+      Range.new(
```
scottming, line 59: These two results are the same.

apps/remote_control/lib/lexical/remote_control/build/state.ex
```diff
@@ -184,7 +184,7 @@ defmodule Lexical.RemoteControl.Build.State do
     RemoteControl.Mix.in_project(fn _ ->
       Mix.Task.clear()
 
-      if connected_to_internet?() do
+      if connected_to_internet?() || true do
```
scottming, line 187: What is the `true` for? To make sure the below code aways will run?

apps/server/lib/lexical/server/state.ex
```diff
@@ -91,10 +91,12 @@ defmodule Lexical.Server.State do
     {:ok, state}
   end
 
-  def apply(%__MODULE__{} = state, %DidChange{lsp: event}) do
+  def apply(%__MODULE__{} = state, %DidChange{lsp: event} = did_change) do
     uri = event.text_document.uri
     version = event.text_document.version
 
+    Logger.info("Did change event #{inspect(did_change)}")
```
scottming, line 98: please delete this line

apps/common/lib/lexical/code_unit.ex
scohen: this is the case where the offset has been found and we return it.

apps/protocol/lib/lexical/protocol/convertibles/mix.task.compiler.diagnostic.ex
scohen: Good catch.

scottming (approved):

PR #110, opened 2023-04-21
Removed Access specifier

scohen:
It was unnecessary, since there is only one way to handle requests in lexical.


scottming:
I tried this branch tonight; the diagnostics, completion, and code actions work well, but need more time to review this huge pr.

scottming (approved):

PR #111, opened 2023-04-21
Move off the `:slave`

### LINKED ISSUE #93: Move off of `:slave` in remote control.
The erlang `:slave` module has the following drawbacks:

  1. It shares IO and file servers with the node that started it. This means the server's `cwd` is constantly changing
  2. It's started by the server node, which means the version of erlang is the same as the one on the server node. This means that projects that declare a different erlang version won't be running in the correct version
  3. It's deprecated
  4. It has a very regrettable name.

I'd suggest moving over to `:peer`, which fixes the first, third and fourth problems, but it still will have the second and it's new in erlang 25. I'd like to have at least two erlang versions of compatibility.

We should investigate starting up our own node and bootstrapping it. We already do this to some extent with `remote_control` so I wonder how hard it would be to launch a blank node and then add the libraries to it.

scottming:
Co-authored-by: Steve Cohen <scohen@scohen.org>

This PR is a significant change:

* We have removed the dependency of `:slave`, which means we have solved all four issues in #93.
* We explicitly use `Port`, and the `ProjectNode` process collects all `stdio/stderr` information. This makes it easier for us to debug some problems.
* The project node and server node are also **half-linked**. This means that when the project node dies, it will be restarted by the server node. However, if the server node dies, the project node will die.

Fixes #93


apps/remote_control/lib/lexical/remote_control.ex
```diff
@@ -49,6 +40,21 @@ defmodule Lexical.RemoteControl do
     end
   end
 
+  def wait_until_connected(node, timeout \\ 5_000)
```
scottming, line 43: I know using `wait_until_connected/2` is a little bit strange here, but I have no idea without it.

```elixir
    receive do
      {:nodeup, ^node, _} ->
        :ok
    after
```
    
I tried monitor_nodes, and tried to receive the `nodeup` message, then `Node.connect(node)`, but it won't work, because the `nodeup` message only appears when `true = Node.connect`

apps/remote_control/lib/lexical/remote_control.ex
```diff
@@ -78,9 +82,18 @@ defmodule Lexical.RemoteControl do
...
-    |> :slave.stop()
+    node = node_name(project)
+    :ok = :net_kernel.monitor_nodes(true, node_type: :visible)
+    :ok = :rpc.call(node, System, :stop, [])
+
+    receive do
+      {:nodedown, ^node, _} ->
+        :ok
+    after
+      5_000 ->
```
scohen, line 93: this is a long while to wait. 
We could do a gentle `System.stop` followed by a `System.halt` after a specified time.

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,28 @@
+defmodule Lexical.RemoteControl.ProjectNode do
+  use GenServer
+
+  def start_link(options) do
+    GenServer.start_link(__MODULE__, options, [])
+  end
+
+  def init(options) do
+    wait_boot = options[:wait_boot] || 5_000
```
scohen, line 9: nit: `boot_timeout` 
```suggestion
    boot_timeout = Keyword.get(options, :boot_timeout, 5000)
```

apps/remote_control/lib/lexical/remote_control.ex
```diff
@@ -49,6 +42,17 @@ defmodule Lexical.RemoteControl do
     end
   end
 
+  def wait_until_be_connected(timeout \\ 5_000) do
```
scohen, line 45: nit: `wait_for_connection`

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,28 @@
...
+    send(self(), :start_remote_control)
+    {:ok, %{options: options, wait_boot: wait_boot}}
```
scohen, line 11: make a state module, not a map.

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,28 @@
+defmodule Lexical.RemoteControl.ProjectNode do
+  use GenServer
+
+  def start_link(options) do
+    GenServer.start_link(__MODULE__, options, [])
+  end
+
+  def init(options) do
+    wait_boot = options[:wait_boot] || 5_000
+    send(self(), :start_remote_control)
```
scohen, line 10: use `handle_continue` here:
```elixir
{:ok, state, {:continue, :start_remote_control}}
end

# then 
def handle_continue(:start_remote_control, state) do 
  ...
end
```

This genserver should probably go into a dynamic supervisor, and it should do the monitoring of the remote node, right?

Then you can do something like: 
```elixir
ProjectNode.wait_until_started(project) 
```

scohen (commented):
This is a nice POC, but it needs some work before it's merged.
Overall, remote control should keep the same interface and guarantees that it had previously. Namely:
  1. `start_link` starts the other node and when it returns the node has started. 
  2. If the manager node dies, so does the project node
  3. If the project node dies, it will be restarted by the manager node.

But the proof of concept works, and gives us everything we wanted, so keep on going!

apps/remote_control/lib/lexical/remote_control.ex
```diff
@@ -7,36 +7,29 @@ defmodule Lexical.RemoteControl do
...
-      ])
-
     apps_to_start = [:elixir | @allowed_apps] ++ [:runtime_tools]
     remote_control_config = Application.get_all_env(:remote_control)
 
-    with {:ok, node} <- :slave.start_link(@localhost_charlist, node_name, erl_args),
-         :ok <- :rpc.call(node, :code, :add_paths, [glob_paths()]),
+    node = node_name(project)
+    paths = format_prepending_paths(glob_paths())
+    start_options = [paths: paths, name: Atom.to_string(node), cookie: Node.get_cookie()]
```
scohen, line 27: i'd prefer these to be actual arguments to start_link

scohen, line 27: you don't need to convert name into a string; atoms convert without the leading colon.

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,28 @@
...
+    {:ok, %{options: options, wait_boot: wait_boot}}
+  end
+
+  def handle_info(:start_remote_control, state) do
+    options = state[:options]
+
+    cmd =
+      "elixir -pa #{options[:paths]} --name #{options[:name]} --cookie #{options[:cookie]} --no-halt -e 'Node.connect(#{inspect Node.self()})'"
```
scohen, line 18: you'll likely need to use the function defined in `remote_control.ex` to find the elixir executable.

apps/remote_control/lib/lexical/remote_control.ex
scottming: > this is a long while to wait.

How long do you think is better?

apps/remote_control/lib/lexical/remote_control.ex
```diff
@@ -7,44 +7,33 @@ defmodule Lexical.RemoteControl do
...
          :ok <-
            :rpc.call(node, RemoteControl.Bootstrap, :init, [
              project,
              project_listener,
              remote_control_config
            ]),
          :ok <- ensure_apps_started(node, apps_to_start) do
-      supervisor_pid = :rpc.call(node, Process, :whereis, [Lexical.RemoteControl.Supervisor])
+      # supervisor_pid = :rpc.call(node, Process, :whereis, [Lexical.RemoteControl.Supervisor])
+      supervisor_pid = Process.whereis(Lexical.RemoteControl.ProjectNodeSupervisor)
```
scottming, line 36: we should use this supervisor_id, right?

apps/remote_control/lib/lexical/remote_control/project_node.ex
scottming: I think you mean: df3293e50fc940a92274e1055ab2b3a8883afdae

However, I still don't know how to halt the project node quickly when the server node halts.

scottming:
> 1. start_link starts the other node and when it returns the node has started.

I applied your another suggestion, use the name: `wait_until_started`, and moved the `:rpc.call(node, Bootstrap, :init, [...])` there

> 2. If the manager node dies, so does the project node

https://github.com/lexical-lsp/lexical/pull/111/files#diff-2f245f89b68569539cdbbd0db39b246705e81aee5189bc6fb3a5d85e4723c7bcR70

> 3. If the project node dies, it will be restarted by the manager node.

I think the node tests already contain these cases.

https://github.com/lexical-lsp/lexical/pull/111/files#diff-3e0ae6b7d322efd64bb87732650edfcd4b7f777086a939401cf3989fdc66c9bcR15

scottming:
sometimes, I'll encounter this error:


```elixir
2023-04-23T16:40:16.852030+08:00 info: Supervisor: {<0.283.0>,tls_dyn_connection_sup}. Started: id=receiver,pid=<0.303.0>.
2023-04-23T16:40:16.852061+08:00 info: Supervisor: {<0.287.0>,tls_dyn_connection_sup}. Started: id=receiver,pid=<0.305.0>.
2023-04-23T16:40:27.909067+08:00 error: Supervisor: {local,'Elixir.Lexical.RemoteControl.Supervisor'}. Context: child_terminated. Reason: killed. Offender: id='Elixir.Lexical.RemoteControl.Build',pid=<0.143.0>.
2023-04-23T16:40:27.909371+08:00 info: Supervisor: {local,'Elixir.Lexical.RemoteControl.Supervisor'}. Started: id='Elixir.Lexical.RemoteControl.Build',pid=<0.5308.0>.
2023-04-23T16:40:27.924909+08:00 error: Error in process <0.5311.0> on node 'lexical@127.0.0.1' with exit value:, {noproc,[{erlang,link,[<0.143.0>],[{error_info,#{module => erl_erts_errors}}]},{'Elixir.Module.ParallelChecker','-spawn/3-fun-0-',5,[{file,"lib/module/parallel_checker.ex"},{line,56}]}]}
2023-04-23T16:40:27.954564+08:00 error: Error in process <0.5313.0> on node 'lexical@127.0.0.1' with exit value:, {noproc,[{erlang,link,[<0.143.0>],[{error_info,#{module => erl_erts_errors}}]},{'Elixir.Module.ParallelChecker','-spawn/3-fun-0-',5,[{file,"lib/module/parallel_checker.ex"},{line,56}]}]}
2023-04-23T16:40:27.957158+08:00 error: Error in process <0.5314.0> on node 'lexical@127.0.0.1' with exit value:, {noproc,[{erlang,link,[<0.143.0>],[{error_info,#{module => erl_erts_errors}}]},{'Elixir.Module.ParallelChecker','-spawn/3-fun-0-',5,[{file,"lib/module/parallel_checker.ex"},{line,56}]}]}
2023-04-23T16:40:27.973821+08:00 error: Error in process <0.5315.0> on node 'lexical@127.0.0.1' with exit value:, {noproc,[{erlang,link,[<0.143.0>],[{error_info,#{module => erl_erts_errors}}]},{'Elixir.Module.ParallelChecker','-spawn/3-fun-0-',5,[{file,"lib/module/parallel_checker.ex"},{line,56}]}]}
2023-04-23T16:40:27.976863+08:00 error: Error in process <0.5316.0> on node 'lexical@127.0.0.1' with exit value:, {noproc,[{erlang,link,[<0.143.0>],[{error_info,#{module => erl_erts_errors}}]},{'Elixir.Module.ParallelChecker','-spawn/3-fun-0-',5,[{file,"lib/module/parallel_checker.ex"},{line,56}]}]}
2023-04-23T16:40:27.997368+08:00 error: Error in process <0.5317.0> on node 'lexical@127.0.0.1' with exit value:, {noproc,[{erlang,link,[<0.143.0>],[{error_info,#{module => erl_erts_errors}}]},{'Elixir.Module.ParallelChecker','-spawn/3-fun-0-',5,[{file,"lib/module/parallel_checker.ex"},{line,56}]}]}
```

scohen:
> If the manager node dies, so does the project node

This means dies for any reason, like a crash and not just being stopped gracefully. I didn't see and monitoring across nodes or anything.

apps/remote_control/lib/lexical/remote_control.ex
scohen: one second?

apps/remote_control/lib/lexical/remote_control.ex
```diff
@@ -77,10 +57,25 @@ defmodule Lexical.RemoteControl do
...
+    :ok = :rpc.call(node, System, :stop, [])
+
+    receive do
+      {:nodedown, ^node, _} ->
+        :ok
+    after
+      5_000 ->
+        Logger.warn("Node #{inspect(node)} did not go down after 5 seconds")
+        halt(project)
+        {:error, :timeout}
```
scohen, line 77: what now? This is an error? What should be done about it?

apps/remote_control/lib/lexical/remote_control.ex
```diff
@@ -89,7 +84,7 @@ defmodule Lexical.RemoteControl do
     |> :erpc.call(m, f, a)
   end
 
-  defp node_name(%Project{} = project) do
+  def node_name(%Project{} = project) do
```
scohen, line 87: why's this public now?

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,83 @@
...
+
+  use GenServer
+
+  def wait_until_started(project, project_listener, boot_timeout \\ 5_000) do
+    :ok = :net_kernel.monitor_nodes(true, node_type: :visible)
+
+    {:ok, node_pid} =
+      DynamicSupervisor.start_child(
+        ProjectNodeSupervisor,
+        {RemoteControl.ProjectNode, project}
```
scohen, line 30: the dynamic supervisor needs to go in an application's supervision tree, or it's not particularly useful.

apps/remote_control/lib/lexical/remote_control.ex
```diff
@@ -77,10 +57,25 @@ defmodule Lexical.RemoteControl do
...
-    |> :slave.stop()
+    node = node_name(project)
+    :ok = :net_kernel.monitor_nodes(true, node_type: :visible)
+    :ok = :rpc.call(node, System, :stop, [])
+
+    receive do
+      {:nodedown, ^node, _} ->
+        :ok
+    after
+      5_000 ->
```
scohen, line 74: Halting the project is really the responsibility of the ProjectNode genserver, though to do it correctly, you're going to need to make use of `GenServer.reply`

what should happen is that you define a function like this in the `ProjectNode` genserver:
```elixir
def stop(%Project{} = project) do 
  project
  |> name()
  |> GenServer.call(:stop)
```

and then the handle_call would look something like this:

```elixir
def handle_call(:stop, from, %State{} = state) do 
   # do the stuff to stop the node
   {:noreply, State.set_stopped_by(state, from), @stop_timeout}
end 
```
and then wait for the message and handle the timeouts.
```elixir
def handle_info({:nodedown, ^name}, state) do 
  GenServer.reply(state.stopped_by, :ok)
  {:stop, :normal, nil}
end

def handle_info(:timeout, state) do 
  # how halt the node
  {:noreply, state}
end
```

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,83 @@
...
+  require Logger
+
+  defmodule State do
+    defstruct [:project, :paths, :cookie]
+
+    def new(project) do
+      paths = format_prepending_paths(RemoteControl.glob_paths())
+      cookie = Node.get_cookie()
+
+      struct!(__MODULE__, paths: paths, cookie: cookie, project: project)
```
scohen, line 12: why use struct! here?
I've always seen `%__MODULE__{paths: paths, cookie: cookie, project: project}`

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,83 @@
+defmodule Lexical.RemoteControl.ProjectNode do
+  alias Lexical.RemoteControl
+  require Logger
+
+  defmodule State do
+    defstruct [:project, :paths, :cookie]
+
+    def new(project) do
```
scohen, line 8: nit:
```suggestion
    def new(%Project{} = project) do
```

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,83 @@
...
+    state = State.new(project)
+    GenServer.start_link(__MODULE__, state, [])
+  end
+
+  def init(state) do
+    {:ok, state, {:continue, :start_remote_control}}
+  end
+
+  def handle_continue(:start_remote_control, state) do
+    name = RemoteControl.node_name(state.project)
```
scohen, line 68: the responsibility of the node name really belongs this module, and not `RemoteControl`.

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,83 @@
...
+
+    def new(project) do
+      paths = format_prepending_paths(RemoteControl.glob_paths())
+      cookie = Node.get_cookie()
+
+      struct!(__MODULE__, paths: paths, cookie: cookie, project: project)
+    end
+
+    defp format_prepending_paths(paths_as_charlists) do
+      Enum.map_join(paths_as_charlists, " -pa ", &Path.expand/1)
```
scohen, line 16: this is really not a use case for map_join, as you're going to miss the initial `-pa` 
Similarly, the name of "paths" is incorrect, they're not paths, they're actually path append arguments. 
For naming, i suggest keeping the paths as actual binary paths and making the following function:

```elixir
def path_append_arguments(%__MODULE__{} = state) do 
   Enum.map(state.paths, fn path -> 
    expanded_path = Path.expand(path)
     "-pa #{expanded_path} "
   end  
  |> IO.iodata_to_binary()
end
```

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,83 @@
...
+
+    with :ok <- wait_until(boot_timeout),
+         :ok <-
+           :rpc.call(node, RemoteControl.Bootstrap, :init, [
+             project,
+             project_listener,
+             remote_control_config
+           ]) do
+      {:ok, node_pid}
+    end
```
scohen, line 44: this should all be done inside the genserver to ensure requests are serialized. What would happen if this is called twice in rapid succession?

Additionally, all the waits would be eliminated in favor of responding to `nodeup` messages and doing a delayed reply with `GenServer.reply`

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,83 @@
...
+      Enum.map_join(paths_as_charlists, " -pa ", &Path.expand/1)
+    end
+  end
+
+  alias Lexical.RemoteControl.ProjectNodeSupervisor
+
+  use GenServer
+
+  def wait_until_started(project, project_listener, boot_timeout \\ 5_000) do
+    :ok = :net_kernel.monitor_nodes(true, node_type: :visible)
```
scohen, line 25: the genserver should be monitoring nodes when it inits, otherwise messages will be delivered to whatever process calls this function.

apps/remote_control/lib/lexical/remote_control/project_node_supervisor.ex
```diff
@@ -0,0 +1,12 @@
...
+
+  def start_link(init_arg) do
+    DynamicSupervisor.start_link(__MODULE__, init_arg, name: __MODULE__)
+  end
+
+  @impl true
+  def init(_init_arg) do
+    DynamicSupervisor.init(strategy: :one_for_one)
+  end
+end
```
scohen, line 12: this should have a function that starts project nodes, no?

apps/server/lib/lexical/server/state.ex
```diff
@@ -66,6 +67,7 @@ defmodule Lexical.Server.State do
 
   def apply(%__MODULE__{shutdown_received?: true} = state, %Exit{}) do
     Logger.warn("Received an Exit notification. Halting the server in 150ms")
+    RemoteControl.halt(state.configuration.project)
```
scohen, line 70: the deadline on `RemoteControl.halt` is over five seconds now, but the prior deadline was 50ms.

apps/server/test/lexical/server/project/node_test.exs
```diff
@@ -12,6 +12,8 @@ defmodule Lexical.Server.Project.NodeTest do
   setup do
     project = project()
 
+    {:ok, _} = start_supervised(Lexical.RemoteControl.ProjectNodeSupervisor)
```
scohen, line 15: this would not be necessary if the project's supervisor had the project node supervisor in it.

apps/server/test/lexical/server/provider/handlers/go_to_definition_test.exs
```diff
@@ -22,13 +22,16 @@ defmodule Lexical.Server.Provider.Handlers.GoToDefinitionTest do
...
 
+    {:ok, _} = start_supervised(Lexical.RemoteControl.ProjectNodeSupervisor)
     {:ok, _} = start_supervised({Server.Project.Supervisor, project})
 
     Dispatch.register(project, [project_compiled()])
     RemoteControl.Api.schedule_compile(project, true)
 
     assert_receive project_compiled(), 5000
 
+    on_exit(fn -> RemoteControl.stop(project) end)
```
scohen, line 33: this is an indication that bidirectional monitoring isn't working properly. You should not have to stop anything manually that was started via `start_supervised`.

apps/server/test/support/completion_case.ex
```diff
@@ -23,10 +23,14 @@ defmodule Lexical.Test.Server.CompletionCase do
         {DynamicSupervisor, name: Server.Project.Supervisor.dynamic_supervisor_name()}
       )
 
+    {:ok, _} = start_supervised(Lexical.RemoteControl.ProjectNodeSupervisor)
     {:ok, _} = start_supervised({Server.Project.Supervisor, project})
     Dispatch.register(project, [project_compiled()])
     RemoteControl.Api.schedule_compile(project, true)
     assert_receive project_compiled(), 5000
+
+    on_exit(fn -> RemoteControl.stop(project) end)
```
scohen, line 32: same comment; this should not have to be stopped manually if monitoring was working correctly.

apps/remote_control/lib/lexical/remote_control/project_node.ex
scottming: Many areas use the `node_name`, should I put this in the `ProjectNode` module instead of `ProjectNode.State`?

apps/remote_control/lib/lexical/remote_control.ex
```diff
@@ -77,10 +55,8 @@ defmodule Lexical.RemoteControl do
     :persistent_term.put({__MODULE__, :project}, project)
   end
 
-  def stop(%Project{} = project) do
-    project
-    |> node_name()
-    |> :slave.stop()
+  def stop(%Project{} = project, stop_timeout \\ 1_000) do
+    ProjectNode.stop(project, stop_timeout)
   end
```
scohen, line 60: this feels like it could be a defdelegate now.

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,188 @@
...
+      :started_by,
+      :status
+    ]
+
+    def new(%Project{} = project) do
+      cookie = Node.get_cookie()
+
+      %__MODULE__{
+        project: project,
+        node: node_name(project),
```
scohen, line 23: why do we need this field? it seems like it's easy to calculate given the project.

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,188 @@
...
+        paths: RemoteControl.glob_paths(),
```
scohen, line 24: I don't like this circular dependency, paths should be passed in.

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,188 @@
...
+            state.node,
+            "--cookie",
+            state.cookie,
+            "--no-halt",
+            "-e",
+            "Node.connect(#{inspect(Node.self())})"
+          ] ++ path_append_arguments(state)
+      )
+
+    state = State.set_started_by(state, from)
```
scohen, line 119: this is all business logic that should move to the state module.
all this function needs to be is 
```elixir 
case State.start(state, from) do 
  {:ok, new_state} -> {:noreply, state}
  {:error, reason} -> {:stop, reason, nil}
end
```

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,188 @@
...
+    {:noreply, state}
+  end
+
+  @impl true
+  def handle_call({:stop, stop_timeout}, from, state) do
+    state = State.set_stopped_by(state, from, stop_timeout)
+    :rpc.call(state.node, System, :stop, [])
```
scohen, line 126: this is business logic that belongs in the state
```elixir
new_state = State.stop(state, from)
{:noreply, state, stop_timeout}
```

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,188 @@
...
+    {:noreply, state, stop_timeout}
+  end
+
+  @impl true
+  def handle_info({:nodeup, _node, _}, state) do
+    GenServer.reply(state.started_by, :ok)
+    {:noreply, State.set_started(state)}
```
scohen, line 133: i'd call this function `State.handle_nodeup`

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,188 @@
...
+    ]
+
+    def new(%Project{} = project) do
+      cookie = Node.get_cookie()
+
+      %__MODULE__{
+        project: project,
+        node: node_name(project),
+        paths: RemoteControl.glob_paths(),
+        cookie: cookie
```
scohen, line 25: there's no default for status.

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,188 @@
...
+  end
+
+  @impl true
+  def handle_info({:nodeup, _node, _}, state) do
+    GenServer.reply(state.started_by, :ok)
+    {:noreply, State.set_started(state)}
+  end
+
+  @impl true
+  def handle_info(:maybe_start_timeout, %{status: status} = state) when status != :started do
```
scohen, line 137: why the maybe?

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,188 @@
...
+    GenServer.reply(state.started_by, {:error, :start_timeout})
+    {:stop, :shutdown, state}
+  end
+
+  @impl true
+  def handle_info(:maybe_start_timeout, state) do
```
scohen, line 143: this would be clearer if the positive case was handled
```elixir
def handle_info(:maybe_start_timeout, %State{status: :started}) do 
  {:noreply, state}
end

def handle_info(:maybe_start_timeout, %State{} = state) do 
  GenServer.reply(state.started_by, {:error, :start_timeout})
  {:stop, :start_timeout, nil}
end

File: apps/remote_control/lib/lexical/remote_control/project_node.ex [hunk 34]
```diff
@@ -0,0 +1,188 @@
...
+    {:noreply, state}
+  end
+
+  @impl true
+  def handle_info({:nodedown, _, _}, state) do
+    GenServer.reply(state.stopped_by, :ok)
```
scohen, line 149: same deal, this is business logic
```elixir
new_state = State.on_nodedown(state)
{:stop, :normal, state}
```

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,188 @@
...
+    {:stop, :shutdown, state}
+  end
+
+  @impl true
+  def handle_info(:timeout, state) do
+    :rpc.call(state.node, System, :halt, [])
```
scohen, line 155: This is business logic. same deal.

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,188 @@
...
+    GenServer.reply(state.stopped_by, :ok)
+    {:stop, :shutdown, state}
```
scohen, line 157: make the reason more descriptive of why this is exiting.

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,188 @@
+defmodule Lexical.RemoteControl.ProjectNode do
+  alias Lexical.Project
+  alias Lexical.RemoteControl
+  require Logger
+
+  defmodule State do
+    defstruct [
+      :project,
```
scohen, line 8: shouldn't you store the port here?

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,188 @@
...
+    {:noreply, state}
+  end
+
+  defp path_append_arguments(%State{} = state) do
+    Enum.map(state.paths, fn path ->
+      expanded_path = Path.expand(path)
+      ["-pa", "#{expanded_path}"]
+    end)
+    |> List.flatten()
+  end
```
scohen, line 177: this belongs in state

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,188 @@
...
+
+  def name(%Project{} = project) do
+    :"#{Project.name(project)}::node_process"
+  end
+
+  defp port_wrapper_executable(project) do
+    # Use the project_dir to detect the correct erl version
+    System.put_env("PROJECT_DIR", Project.root_path(project))
+    Path.join(:code.priv_dir(:remote_control), "port_wrapper.sh")
+  end
```
scohen, line 187: this belongs in state, and why are you setting an environment variable?

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,188 @@
...
+
+  def name(%Project{} = project) do
+    :"#{Project.name(project)}::node_process"
+  end
+
+  defp port_wrapper_executable(project) do
+    # Use the project_dir to detect the correct erl version
+    System.put_env("PROJECT_DIR", Project.root_path(project))
+    Path.join(:code.priv_dir(:remote_control), "port_wrapper.sh")
```
scohen, line 186: nitpick:
```elixir
:remote_control
|> :code.priv_dir()
|> Path.join("port_wrapper.sh")
```

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,188 @@
...
+    GenServer.start_link(__MODULE__, state, name: name(project))
+  end
+
+  @impl GenServer
+  def init(state) do
+    {:ok, state}
+  end
+
+  @impl true
+  def handle_call(:start, from, state) do
```
scohen, line 97: please add a `%State{} = ` pattern match on the genserver callbacks to be clear what state is.

apps/remote_control/priv/port_wrapper.sh
```diff
@@ -0,0 +1,45 @@
...
+
+set_up_version_manager() {
+    if [ -e $HOME/.asdf ]; then
+        VERSION_MANAGER="asdf"
+    elif [ -e $HOME/.rtx ]; then
+        VERSION_MANAGER="rtx"
+    else
+        VERSION_MANAGER="none"
+    fi
+}
```
scohen, line 11: i also think this needs to be removed from `start_lexical.sh`

apps/server/test/support/completion_case.ex
```diff
@@ -25,6 +25,7 @@ defmodule Lexical.Test.Server.CompletionCase do
     Dispatch.register(project, [project_compiled()])
     RemoteControl.Api.schedule_compile(project, true)
     assert_receive project_compiled(), 5000
+
```
scohen, line 28: remove?

apps/remote_control/lib/lexical/remote_control/project_node.ex
scottming: set to `initializing`

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,210 @@
...
+      :rpc.call(node_name(state.project), System, :stop, [])
+      %{state | stopped_by: from, stop_timeout: stop_timeout, status: :stopping}
+    end
+
+    def halt(%__MODULE__{} = state) do
+      :rpc.call(node_name(state.project), System, :halt, [])
+      on_nodedown(state)
+    end
+
+    def handle_nodeup(%__MODULE__{} = state) do
```
scohen, line 65: nit: make the naming consistent with `on_nodedown`

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,210 @@
...
+  end
+
+  alias Lexical.RemoteControl.ProjectNodeSupervisor
+  use GenServer
+
+  def start(project, project_listener, paths) do
+    node_name = State.node_name(project)
+    remote_control_config = Application.get_all_env(:remote_control)
+
+    {:ok, node_pid} = ProjectNodeSupervisor.start_project_node(project)
```
scohen, line 97: Should this go into the with? Do you want to crash if this fails?

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,210 @@
...
+  def handle_info({:nodedown, _, _}, %State{} = state) do
+    GenServer.reply(state.stopped_by, :ok)
+    state = State.on_nodedown(state)
+    {:stop, :shutdown, state}
+  end
+
+  @impl true
+  def handle_info({:DOWN, _ref, :process, _object, _reason}, %State{} = state) do
+    # The launcher has died. Usually, the launcher is Server.Project.Node,
+    # and when it dies, both the process and project node vm should die.
```
scohen, line 184: how can you tell who this is? it could be the starter, the stopper or the node process, correct?

rel/deploy/overlays/start_lexical.sh
```diff
@@ -1,15 +1,5 @@
...
-set_up_version_manager() {
-    if [ -e $HOME/.asdf ]; then
-        VERSION_MANAGER="asdf"
-    elif [ -e $HOME/.rtx ]; then
-        VERSION_MANAGER="rtx"
-    else
-        VERSION_MANAGER="none"
-    fi
-}
-
```
scohen, line 12: if `include_erts` is false (and it should be), we'll still need all this.

mix.exs
```diff
@@ -67,7 +67,7 @@ defmodule Lexical.MixProject do
           mix: :load
         ],
         include_executables_for: [:unix],
-        include_erts: false,
+        include_erts: true,
```
scohen, line 70: this should go back to false.

scohen (commented):
this is a vast improvement. Looks almost ready.

apps/remote_control/lib/lexical/remote_control/project_node.ex
scottming: But I only monitored the starter.

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,210 @@
...
+  @impl true
+  def handle_call({:stop, stop_timeout}, from, %State{} = state) do
+    state = State.stop(state, from, stop_timeout)
+    {:noreply, state, stop_timeout}
+  end
+
+  @impl true
+  def handle_info({:nodeup, _node, _}, %State{} = state) do
+    GenServer.reply(state.started_by, :ok)
+    {pid, _ref} = state.started_by
```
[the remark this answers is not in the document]
scottming: Why do we need to ignore it?

mix.exs
scottming: Are you sure?

> You may also set this option to false if you desire to use the ERTS version installed on the target. Note, however, that the ERTS version on the target must have the exact version as the ERTS version used when the release is assembled. Setting it to false also disables hot code upgrades. Therefore, :include_erts should be set to false with caution and only if you are assembling the release on the same server that runs it.

use `false`, we can't start the `start_lexical.sh` when version not matching.

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -0,0 +1,210 @@
...
+    def on_nodedown(%__MODULE__{} = state) do
+      %{state | status: :stopped}
+    end
+
+    def node_name(%Project{} = project) do
+      :"#{Project.name(project)}@127.0.0.1"
+    end
+
+    defp path_append_arguments(paths) do
+      Enum.flat_map(paths, fn path ->
```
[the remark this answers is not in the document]
scottming: I prefer leave that, `(&1)])`is not clear.

apps/remote_control/lib/lexical/remote_control/project_node.ex
scottming: Should rename this, because     
```elixir
with {:ok, node_pid} <- ProjectNodeSupervisor.start_project_node(project),
         :ok <- start_node(project, paths),
 ```
` ProjectNodeSupervisor.start_project_node` followed by a `start_node` looks not good

File: apps/remote_control/lib/lexical/remote_control/project_node.ex (hunk 49 above)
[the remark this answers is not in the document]
Comment (scottming) [reply apps/remote_control/lib/lexical/remote_control/project_node.ex]: Also, I prefer to put `GenServer.reply` in the callback function as it seems strange to have lifecycle functions in the state.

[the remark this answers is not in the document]
Comment (scottming) [reply apps/remote_control/lib/lexical/remote_control/project_node.ex]: > link to it and do your logic in the exit message handler.

I modified the code via this way.

File: apps/remote_control/lib/lexical/remote_control/project_node.ex [hunk 51]
```diff
@@ -0,0 +1,210 @@
...
+
+  @impl true
+  def handle_info({:nodedown, _, _}, %State{} = state) do
+    GenServer.reply(state.stopped_by, :ok)
+    state = State.on_nodedown(state)
+    {:stop, :shutdown, state}
+  end
+
+  @impl true
+  def handle_info({:DOWN, _ref, :process, _object, _reason}, %State{} = state) do
```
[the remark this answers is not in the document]
scottming: However, I am inquisitive. Since `RemoteControl.ProjectNode` and `Server.Project.Node` have already been linked, why doesn't the death of the project node vm cause the death of the `Server.Project.Node` process? Normally, both sides need to use a flag(:trap_exit).

mix.exs
scohen: you're not understanding what the docs are saying. Including ERTS means that we include the entire runtime from the build environment, which means that this will not run on your computer if you download the build from githun, like vscode users do. We need to use the default runtime that users have on their computers. 
You also need to revert your changes to `start_lexical.sh`

apps/remote_control/lib/lexical/remote_control/project_node.ex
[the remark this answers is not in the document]
scottming: yes, I just mean `&[`

scottming:
@wingyplus I use these commands can't reproduce:

```bash
$ rm -rf _build deps .lexical && mix deps.get  
$ env MIX_ENV=prod mix release lexical             
$ env MIX_ENV=prod NAMESPACE=true mix release lexical --overwrite  
```

mix.exs
scottming: > Including ERTS means that we include the entire runtime from the build environment
I know this.

but the doc said: 

> that the ERTS version on the target must have the exact version as the ERTS version used when the release is assembled

scottming: I think it means that `include_erts: false` requires the runtime version of erl to be consistent with the release version.

apps/remote_control/lib/lexical/remote_control/project_node.ex
scohen: Yes, but you're trapping exits, and exits can happen all the time.

mix.exs
scohen: I still thing this needs to be false. What we need to do is build a version for every major elixir / erlang release that we support, and the start_lexical script needs to be made smart about which environment it uses. 
So we need to build a matrix like:

Erlang:     Elixir
24            1.13, 1.14
25           1.13, 1.14
26           1.13, 1.14, (1.15)

scohen: If this is true, the situation is even worse, we'll need to build releases for each elixir, erlang, O/S, *and* architecture we support.

scohen (approved):

Issue #112, opened 2023-04-21
Rebuild project node on deps change


scohen:
When a dependency is added or removed, we need to rebuild the project in the project node. I think this can be accomplished by issuing a build request with `force? = true`. 

To accomplish this, we need to monitor all the `mix.exs` files in the projejct, and if any has changed, check the  deps apps `Mix.Project.deps_apps` in the project for changes. This means we'll need to have the editor watch files for us.

PR #115, opened 2023-04-24
Sped up builds by moving steps to the force? path

scohen:
Project-wide builds were taking around 900ms on my laptop. A lot of this time was coming from tasks that I thought should have negligible run times, like fetching deps when there were none to fetch. It turns out that they were each contributing around 250ms to the build. By moving these tasks to only run if a build is being forced, the compilation now takes around 250ms.


scottming:
Since we don't have a solution for #112 now, after merging this PR, you'll need to restart your editor when you make changes to `mix.exs` dependencies.

scottming (approved):

PR #118, opened 2023-04-25
Fix neovim completion issue caused by incomplete single character

scottming:
based on the protocol

> for speed clients should be able to filter an already received completion list if the user continues typing. Servers can opt out of this using a [CompletionList](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionList) and mark it as isIncomplete.

and the cmp author's suggestions: https://github.com/hrsh7th/cmp-nvim-lsp/issues/55

image


lib/generated/lexical/protocol/types/completion/list.ex
```diff
@@ -0,0 +1,30 @@
+# This file's contents are auto-generated. Do not edit.
```
scohen, line 1: Was this generated with the mix task? If not, please generate it with:

`mix lsp.mappings.generate --roots CompletionList` 

I'm asking because I see parens around deftype.

scottming: good catch

apps/server/lib/lexical/server/code_intelligence/completion.ex
```diff
@@ -2,6 +2,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
   alias Lexical.Completion.Translatable
   alias Lexical.Project
   alias Lexical.Protocol.Types.Completion
+  alias Lexical.Protocol.Types.Completion.List, as: CompletionList
```
scohen, line 5: slight preference for `Completion.List`, and just alias `Types.Completion`

scohen (approved):

apps/server/lib/lexical/server/code_intelligence/completion.ex
scottming: Because in Elixir, the `List` gives the impression of `[]`, and `List.new` feels strange. Although, I know that in JavaScript all lists are objects.

scohen: you would do `Completion.LIst.new`, not `List.new`

PR #119, opened 2023-04-25
Make adding curlies more accurate

scottming:
image

Above is the current behavior, which is annoying.

### DIFF
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
@@ -44,7 +44,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
       end
 
     completion_kind =
-      if defines_struct? do
+      if add_curlies? do
         :struct
       else
         :module
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
@@ -18,7 +18,10 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
     detail = builder.fallback(module.summary, module.name)
     struct_reference? = Env.struct_reference?(env)
     defines_struct? = Intelligence.defines_struct?(env.project, module.full_name)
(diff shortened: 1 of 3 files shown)


apps/server/test/lexical/server/code_intelligence/completion/translations/struct_test.exs
```diff
@@ -37,7 +37,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.StructTest do
       assert completion.kind == :struct
     end
 
-    test "should complete module aliases after %", %{project: project} do
+    test "should complete module aliases and ends with {} after %", %{project: project} do
```
scohen, line 40: what about a test for not adding curlies when it already ends with {} ?

apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
```diff
@@ -18,7 +18,10 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
     detail = builder.fallback(module.summary, module.name)
     struct_reference? = Env.struct_reference?(env)
     defines_struct? = Intelligence.defines_struct?(env.project, module.full_name)
-    add_curlies? = defines_struct? and not String.contains?(env.suffix, "{")
+
+    add_curlies? =
```
scohen, line 22: I suggest changing line 47 to `if add_curlies? do`

apps/server/test/lexical/server/code_intelligence/completion/translations/struct_test.exs
scottming: o, there is already a test to do that; I need to reset this description.

apps/server/test/lexical/server/code_intelligence/completion/translations/struct_test.exs
```diff
@@ -51,6 +51,20 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.StructTest do
...
+        defmodule TestModule do
+        alias Project.Structs.User
+
+        Us|
+      ]
+
+      assert [completion] = complete(project, source)
+
+      assert completion.insert_text == "User"
+      assert completion.kind == :struct
```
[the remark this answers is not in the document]
scottming: right.

scohen (approved):

PR #120, opened 2023-04-25
Dialyzer analysis

scohen:
To prepare for adding dialyzer diagnostics to the project, it would be nice if we ran dialyzer clean. This PR fixes a number of issues that dialyzer found, and actually fixes a couple of bugs that were going unreported.
In addition, it enables dialyzer analysis in our github actions workflow, so we can keep the project clean.

I had to ignore a couple of things in the configuration code, as it would be impossible for it to dialyze clean as long as we don't have a way to enable dialyzer.


apps/common/lib/lexical/source_file/range.ex
```diff
@@ -7,8 +7,8 @@ defmodule Lexical.SourceFile.Range do
   defstruct start: nil, end: nil
 
   @type t :: %__MODULE__{
-          start: non_neg_integer(),
```
scohen, line 10: this was very, very wrong.

apps/common/lib/lexical/source_file.ex
```diff
@@ -109,14 +109,18 @@ defmodule Lexical.SourceFile do
...
-      Enum.reduce_while(changes, source, fn change, source ->
-        case apply_change(source, change) do
-          {:ok, new_source} ->
-            {:cont, new_source}
-
-          error ->
-            {:halt, error}
-        end
+      Enum.reduce_while(changes, source, fn
+        nil, source ->
```
scottming, line 113: It really caught bugs.

apps/protocol/test/support/lsp_protocol.ex
```diff
@@ -165,10 +165,9 @@ defmodule Lexical.Proto.Fixtures.LspProtocol do
           error -> {:halt, error}
         end
       end)
-      |> Map.new()
 
     case list_elements do
-      %{} = elements -> {:ok, elements}
+      %{} = elements -> {:ok, Map.new(elements)}
```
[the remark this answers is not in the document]
scohen: no, they're keyword lists. This module is used to build nested structures, and you pass in keywords.

apps/server/lib/lexical/server.ex
```diff
@@ -20,6 +20,8 @@ defmodule Lexical.Server do
     Requests.Shutdown
   ]
 
+  @dialyzer {:nowarn_function, apply_to_state: 2}
```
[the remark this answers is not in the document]
scohen: There are a couple errors caused by configuration not having dialyzer enabled that cause upstream problems here. It's just noise, and will be fixed once we have the ability to turn dialyzer on and off.

scottming (approved):

apps/common/lib/lexical/source_file.ex
scohen: this was handled before with a function head that I removed. This seems cleaner to me.

Issue #83, opened 2023-04-01
Protocol improvements


scohen:
Some ideas for improvements in the protocol:

 

- [x]  1. Have requests and notifications point directly to their params structs rather than having to spell them out.
- [ ]   2. Make it so requests know about and can generate their responses.
- [ ]   3. Maybe use ecto schemas for validation of parameters.
- [x]   4. Remove the `:exclusive` keyword on defrequest
- [x]   5. Have responses convert their elixir positions automatically.

scohen:
4 is fixed by #110 
5 is fixed by #109

scohen:
1 is fixed by #127

scohen:
The protocol is pretty stable right now, We can revisit more improvements if people want them, but closing for now.

PR #121, opened 2023-04-25
Simplified building requests and responses

scohen:
The protocol acutally defines params types for notifications and responses, and we were recapitulating them in their definitions. This PR adds a bunch of the auto-generated types and allows notification/request definitions to reference them. Then the types are consulted to build the request or notification's parameters.

The result is much simpler definitions and less manual work when we need to create new ones.

### DIFF
--- a/apps/proto/.formatter.exs
+++ b/apps/proto/.formatter.exs
@@ -4,6 +4,7 @@ proto_dsl = [
   defenum: 1,
   defnotification: 1,
   defnotification: 2,
+  defrequest: 1,
   defrequest: 2,
   defresponse: 1,
   deftype: 1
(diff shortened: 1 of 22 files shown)


scohen:
Note, the files in `apps/protocol/lib/generated` are auto-generated and you don't need to review them.

scohen:
Fixes the first bullet point of #83

apps/proto/mix.exs
```diff
@@ -25,6 +25,7 @@ defmodule Proto.MixProject do
   # Run "mix help deps" to learn about dependencies.
   defp deps do
     [
+      {:patch, "~> 0.12", only: [:test]},
```
scottming, line 28: Why we need `patch` in `proto` app?

scottming (approved):

apps/proto/mix.exs
scohen: We don't, good catch.

PR #130, opened 2023-04-27
Made the completion environment a builder

scohen:
I noticed that some of our suggestions weren't coming across and displaying in my editor. This is because the editor is performing logic (incorrectly) about what the edits should be, and deciding that they don't apply in this case. To fix this, I'm going to transition translations for completions to use text edits, which fixes the problem and takes changes out of the hands of the editor.

The first step is to require builders to take the environment, as it will be helpful for figuring out ranges semi-automatically. Then I thought, "why not just make the Env a builder so we don't need two things?" So here you go. The Env is an environment, and a builder.


scottming:
Does this pr fix something? or refactor?

image

This no-completion problem when `%Project|` still exists.

scohen:
It's just a refactor to move ourselves in the direction of using Edits.

scottming (approved):

PR #131, opened 2023-04-29
Correct arity of format_string

Nicd:
The function is called with only one argument. Using it like this leads to a crash of the server on save.


scottming (approved):

scottming:
#132

scohen (approved):

scohen:
The question I have is why fetch_formatter() fails

PR #134, opened 2023-04-29
ci: Remove build directory structure from released archives

Blond11516:
Changes the released archives to have the release files at the root rather than under the `_build/dev/rel/lexical` path.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,8 +56,10 @@ jobs:
 
       - name: Archive release
         run: |
-          zip -r lexical.zip _build/dev/rel/lexical
-          cp lexical.zip lexical-${{ steps.timestamp.outputs.result }}.zip
+          cd _build/dev/rel/lexical
+          zip -r lexical.zip *
+          cp lexical.zip ../../../../lexical.zip
+          cp lexical.zip ../../../../lexical-${{ steps.timestamp.outputs.result }}.zip
 
       - name: Create Git tag for commit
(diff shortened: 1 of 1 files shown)


scohen:
I've been thinking of making a `releases` directory in the current directory in the mix.release task, would that be better, @Blond11516 ?

Blond11516:
@scohen I guess that would be a bit easier yes, but honestly I don't think it would make much of a difference. I'd still want to navigate inside that directory to build the archive so there aren't any unnecessary directories in it.

.github/workflows/release.yml
```diff
@@ -56,8 +56,10 @@ jobs:
 
       - name: Archive release
         run: |
-          zip -r lexical.zip _build/dev/rel/lexical
-          cp lexical.zip lexical-${{ steps.timestamp.outputs.result }}.zip
+          cd _build/dev/rel/lexical
+          zip -r lexical.zip *
+          cp lexical.zip ../../../../lexical.zip
```
scohen, line 61: why's this here?

Blond11516: Because the step that creates the release expects the ZIPs to be at the root of the directory. It's either that or changing the release step to use `_build/dev/rel/lexical/lexical*.zip` as artifacts.

I figured this was simpler because it avoided having to modify another step, but come to think of it it might make more sense for the ZIPs to be in the same place as the release :thinking:

scohen (approved):

PR #135, opened 2023-05-02
Support work done progress during project compilation

scottming:
https://user-images.githubusercontent.com/12830256/235573820-41696dbe-a9dd-4af5-a7c0-d27e835c575c.mp4

### TODO

* [ ] make the *mix deps.get* more precise
* [ ] make the *mix deps.compile* more precise
* [ ] make the *mix compile* more precise


scottming:
A tricky issue is that it also takes time to start the **project node**. It would be great if we could report progress for it too, but the key is that we cannot use `RemoteControl.notify_listener` until the project node has started.

scottming:
The implementation currently is simple and ugly, not ready for review

apps/server/lib/lexical/server/project/progress.ex
```diff
@@ -0,0 +1,252 @@
...
+    def new(options) do
+      %__MODULE__{
+        title: options[:title],
+        message: options[:message],
+        percentage: options[:percentage]
+      }
+    end
+  end
+
+  defmodule State do
```
scohen, line 16: all functions in this module should have `%__MODULE__{} = state` as their first parameter.

apps/server/lib/lexical/server/project/progress.ex
```diff
@@ -0,0 +1,252 @@
...
+    end
+  end
+
+  alias Lexical.Project
+  alias Lexical.Protocol.Notifications.CreateWorkDoneProgress
+  alias Lexical.Protocol.Notifications.Progress, as: LSProgress
+  alias Lexical.Protocol.Types.WorkDone.ProgressBegin
+  alias Lexical.Protocol.Types.WorkDone.Progress.Report, as: ProgressReport
+  alias Lexical.Protocol.Types.WorkDone.ProgressEnd
+  alias Lexical.RemoteControl.Api.Messages
```
scohen, line 148: why the aliases?
Just alias `Notifications` and `Types.WorkDone` and you'll have nicer code with a lower cognitive buden.

apps/protocol/lib/generated/lexical/protocol/types/work_done/progress_begin.ex
```diff
@@ -0,0 +1,11 @@
+# This file's contents are auto-generated. Do not edit.
+defmodule Lexical.Protocol.Types.WorkDone.ProgressBegin do
```
scohen, line 2: this mapping, and the end mapping feel a bit wonky. 
Perhaps we change them to `WorkDone.Progress.Begin`,  and`Workdone.Progress.End` ?

It's easy to do, just go edit `type_mappings.json` and re-generate.

apps/server/lib/lexical/server/project/progress.ex
```diff
@@ -0,0 +1,252 @@
...
+    end
+
+    {:noreply, state}
+  end
+
+  defp name(%Project{} = project) do
+    :"#{Project.name(project)}::progress"
+  end
+
+  defp create_workdone_progress(state, label) do
```
scohen, line 201: All the functions below this belong in the state. They're all business logic.

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -158,7 +158,9 @@ defmodule Lexical.RemoteControl.ProjectNode do
   end
 
   @impl true
-  def handle_info({_port, {:data, _message}}, state) do
+  def handle_info({_port, {:data, message}}, state) do
+    message = message |> IO.iodata_to_binary() |> String.trim()
+    Logger.info("Port: #{message}")
```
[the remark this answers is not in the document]
scottming: I wanted to use this message to calculate the progress of `deps.get`, but it wasn't frequent, so it may not be necessary.

apps/server/lib/lexical/server/project/progress.ex
```diff
@@ -0,0 +1,252 @@
...
+  require Logger
+
+  defmodule ProgressValue do
+    defstruct [:title, :message, :percentage]
+
+    def new(options) do
+      %__MODULE__{
+        title: options[:title],
+        message: options[:message],
+        percentage: options[:percentage]
```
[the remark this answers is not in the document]
scottming: That's good

But the key is that it's really difficult for us to calculate these two values. I plan to implement this percentage function after I figure out the compilation relationship of Elixir in the future.

apps/server/lib/lexical/server/project/progress.ex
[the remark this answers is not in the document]
scottming: This `boundary` suggestions really helpful, thank you so much, the code looks clear now.

apps/server/lib/lexical/server/project/progress.ex
```diff
@@ -0,0 +1,252 @@
...
+        _ ->
+          # TODO: refactor and reduce complexity
+          percentages_by_label = percentages(state, label, message)
+          state = %{state | percentages_by_label: percentages_by_label}
+          progress = report_progress(state, label, message)
+          %{state | progress: progress}
+      end
+    end
+
+    defp percentages(state, "compile", file) do
```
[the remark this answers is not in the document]
scottming: Because the main time-consuming process during each startup is compilation.

apps/remote_control/lib/lexical/remote_control/compile_tracer.ex
```diff
@@ -7,15 +7,35 @@ defmodule Lexical.RemoteControl.CompileTracer do
...
+
+  defp progress_message(file) do
+    relative_path = Path.relative_to_cwd(file)
+
+    if String.starts_with?(relative_path, "deps") do
+      project_progress(label: "deps.compile", message: relative_path)
+    else
+      project_progress(label: "compile", message: relative_path)
+    end
+  end
```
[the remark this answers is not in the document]
scottming: I haven't done any specific measures, but the preparation time for opening the lexical project is about 10 seconds. Do you have any suggestions on how to measure it?

apps/protocol/lib/lexical/protocol/notifications.ex
```diff
@@ -87,5 +87,17 @@ defmodule Lexical.Protocol.Notifications do
     end
   end
 
+  defmodule CreateWorkDoneProgress do
```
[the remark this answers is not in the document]
scottming: In addition, most of the modules in this file start with verbs, such as `ShowMessage` and `PublishDiagnostics`.

apps/server/lib/lexical/server/project/progress.ex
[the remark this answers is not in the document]
scohen: yes, but why handle it differently than other things?

apps/remote_control/lib/lexical/remote_control/build/progress.ex
```diff
@@ -0,0 +1,16 @@
+defmodule Lexical.RemoteControl.Build.Progress do
+  alias Lexical.RemoteControl
+
+  import Lexical.RemoteControl.Api.Messages
+
+  def with_progress(label, func, opts \\ []) when is_function(func) do
```
scohen, line 6: note: you can specify the arity of the function in `is_function`.
```suggestion
  def with_progress(label, func, opts \\ []) when is_function(func, 0) do
```

apps/server/lib/lexical/server/project/progress/util.ex
```diff
@@ -0,0 +1,15 @@
+defmodule Lexical.Server.Project.Progress.Util do
+  alias Lexical.Server.Project.Progress
+
+  import Lexical.RemoteControl.Api.Messages
+
+  def with_progress(project, label, func) do
+    progress_pid = Progress.name(project)
+
+    send(progress_pid, project_progress(label: label <> ".begin"))
```
scohen, line 9: I think you can get away from string concatenation and splitting if you just use a more descriptive record
```elixir
project_progress(label: label, stage: :begin)
```

apps/server/lib/lexical/server/project/progress/state.ex
```diff
@@ -0,0 +1,108 @@
...
+    |> String.trim_trailing(".begin")
+    |> String.trim_trailing(".end")
+  end
+
+  def kind(label) do
+    kind = String.split(label, ".") |> List.last()
+    if kind in ["prepare", "begin", "end"], do: String.to_atom(kind), else: :report
+  end
+
+  defp create_workdone_progress(%__MODULE__{} = state, label) do
```
scohen, line 62: nit: `work_done`

apps/server/lib/lexical/server/project/node.ex
```diff
@@ -47,7 +49,7 @@ defmodule Lexical.Server.Project.Node do
 
   @impl GenServer
   def init(%Project{} = project) do
-    case start_node(project) do
+    case with_progress project, "Project Node", fn -> start_node(project) end do
```
scohen, line 52: this would look better with parens.

apps/server/lib/lexical/server/project/progress/state.ex
```diff
@@ -0,0 +1,108 @@
...
+
+  def complete(%__MODULE__{} = state, label, message \\ "") do
+    progress = end_progress_value(state, label, message)
+    write(progress)
+
+    state = clear_token_by_label(state, label)
+    %{state | progress: progress}
+  end
+
+  def trim(label) do
```
scohen, line 50: is this supposed to be public?

apps/server/lib/lexical/server/project/progress/util.ex
```diff
@@ -0,0 +1,15 @@
+defmodule Lexical.Server.Project.Progress.Util do
```
scohen, line 1: Note: A module named util almost is certainly not what you want ;)

apps/server/lib/lexical/server/project/progress/util.ex
```diff
@@ -0,0 +1,15 @@
...
+  alias Lexical.Server.Project.Progress
+
+  import Lexical.RemoteControl.Api.Messages
+
+  def with_progress(project, label, func) do
+    progress_pid = Progress.name(project)
+
+    send(progress_pid, project_progress(label: label <> ".begin"))
+    result = func.()
```
scohen, line 10: this should probably try/rescue to ensure the end progress is sent.

apps/server/lib/lexical/server/project/progress/value.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Lexical.Server.Project.Progress.Value do
+  alias Lexical.Protocol.Notifications
+  alias Lexical.Protocol.Types.WorkDone
+
+  @enforce_keys [:token, :kind]
+  defstruct [:token, :kind, :title, :message, :percentage]
```
scohen, line 6: is percentage used anywhere?

apps/server/lib/lexical/server/project/progress/state.ex
```diff
@@ -0,0 +1,108 @@
...
+    :progress
+  ]
+
+  def new(%Project{} = project) do
+    %__MODULE__{
+      project: project,
+      token_by_label: %{},
+      progress: nil
+    }
+  end
```
scohen, line 21:
```suggestion
  defstruct [project: nil, token_by_label: %{}, progress: nil]

  def new(%Project{} = project) do
    %__MODULE__{project: project}
  end
```

apps/server/lib/lexical/server/project/progress/util.ex
scottming: So, Any better names? Naming is so hard for me.😅

apps/server/lib/lexical/server/project/progress.ex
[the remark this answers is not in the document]
scottming: At that time, my idea was to only target the calculation of percentages for compilation. Now, this part has been removed.

apps/server/lib/lexical/server/project/progress/util.ex
scottming: This function is only called in the "start project node" function. If it cannot be launched, it still will send an end event. A try-catch block may not be necessary for now.

apps/server/lib/lexical/server/project/progress/value.ex
scottming: nope

apps/server/lib/lexical/server/project/progress/util.ex
scottming: Another point is the issue of user experience. I'm not sure if it's necessarily a good thing to be unable to send an end message when encountering an exception, as this would force users to check logs for issues. Personally, I like doing it this way.

apps/protocol/lib/lexical/protocol/notifications.ex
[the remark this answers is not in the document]
scohen: the entire LSP namespace is flat, we've added the depth ourselves, and is wildly, wildly inconsistent. I'm trying ot hammer out some consistency.

apps/remote_control/lib/lexical/remote_control/build/progress.ex
```diff
@@ -0,0 +1,16 @@
...
+
+  import Lexical.RemoteControl.Api.Messages
+
+  def with_progress(label, func, opts \\ []) when is_function(func, 0) do
+    mix? = Keyword.get(opts, :mix?, true)
+    label = if mix?, do: "mix " <> label, else: label
+
+    RemoteControl.notify_listener(project_progress(label: label, stage: :begin))
+    result = func.()
+    RemoteControl.notify_listener(project_progress(label: label, stage: :end))
```
scohen, line 12: you sure this shouldn't be wrapped with a try / after ? If func.() blows up for any reason, you'll never send the `end` message, despite the fact that the task has indeed ended (just not successfully)

```suggestion
try do
    RemoteControl.notify_listener(project_progress(label: label, stage: :begin))
    result = func.()
after
    RemoteControl.notify_listener(project_progress(label: label, stage: :end))
end
```

apps/remote_control/lib/lexical/remote_control/api/messages.ex
```diff
@@ -16,6 +16,8 @@ defmodule Lexical.RemoteControl.Api.Messages do
   defrecord :project_diagnostics, project: nil, diagnostics: []
   defrecord :file_diagnostics, project: nil, uri: nil, diagnostics: []
 
+  defrecord :project_progress, label: nil, message: nil, stage: nil
```
scohen, line 19: default the stage to `:report`, as it's the most used.

apps/server/lib/lexical/server/project/progress.ex
```diff
@@ -0,0 +1,55 @@
...
+          State.begin(state, label)
+
+        :report ->
+          State.update(state, label, message)
+
+        :end ->
+          State.complete(state, label, message)
+
+        _ ->
+          state
```
scohen, line 46: do you need this?

apps/server/lib/lexical/server/project/progress.ex
```diff
@@ -0,0 +1,55 @@
...
+    new_state =
+      case stage do
+        :begin ->
+          State.begin(state, label)
+
+        :report ->
+          State.update(state, label, message)
+
+        :end ->
+          State.complete(state, label, message)
```
scohen, line 43: you could also do something like
```elixir
def handle_info(project_progress(stage: stage) = message, %State{} = state) do 
  new_state = apply(State, stage, [state, message])
  {:noreply, new_state}
end
```

then the state would be:
```elixir
def begin(%__MODULE__{} = state, project_progress(label: label) do 
   ...
end
```

apps/server/lib/lexical/server/project/progress/state.ex
```diff
@@ -0,0 +1,88 @@
...
+
+  defp write(progress) do
+    if progress && progress.token do
+      progress |> Value.to_progress() |> Transport.write()
+    end
+  end
+
+  defp value(:begin, state, label, _message) do
+    token = get_token(state, label)
+    Value.new(kind: :begin, token: token, title: label)
```
scohen, line 55: heh, I might push this logic into value, so you have `begin`, `report` and `end` functions there

I'm a little wary of naming functions `end` though, how about `complete`  or `start` and `finish`?

apps/server/lib/lexical/server/project/progress/util.ex
scohen: won't that just cause a message to be left open? That's not great; I'd rather tell them that it failed.

apps/server/lib/lexical/server/project/progress/state.ex
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+  def complete(%__MODULE__{} = state, label, message) do
+    progress = value(:end, state, label, message)
+    write(progress)
+    state = clear_token_by_label(state, label)
+    %{state | progress: progress}
+  end
+
+  defp create_work_done(%__MODULE__{} = state, label) do
```
scohen, line 41: this does more than create the work done, it writes it.
`write_work_done` ?

apps/server/lib/lexical/server/project/progress/state.ex
```diff
@@ -0,0 +1,88 @@
...
+    token = get_token(state, label)
+    progress = Notifications.CreateWorkDoneProgress.new(token: token)
+    Transport.write(progress)
+  end
+
+  defp write(progress) do
+    if progress && progress.token do
+      progress |> Value.to_progress() |> Transport.write()
+    end
+  end
```
scohen, line 51: this could also be done in the function head
```elixir
def write(%Value{token: token} = progress) when not is_nil(token) do 
   ...
end
```

apps/server/lib/lexical/server/project/progress/state.ex
```diff
@@ -0,0 +1,88 @@
+defmodule Lexical.Server.Project.Progress.State do
+  alias Lexical.Project
+  alias Lexical.Protocol.Notifications
+  alias Lexical.Server.Project.Progress.Value
+  alias Lexical.Server.Transport
+
+  require Logger
+
+  defstruct project: nil, token_by_label: %{}, progress: nil
```
scohen, line 9: I think I said this before, but why not just store the values here? This seems easier than doing the token by label thing. 
Presumably you can have multiple progresses being updated at the same time, right?

apps/server/lib/lexical/server/project/progress/state.ex
```diff
@@ -0,0 +1,88 @@
...
+    token_by_label = Map.delete(state.token_by_label, label)
+    %{state | token_by_label: token_by_label}
+  end
+
+  defp get_token_when_update(state, label) do
+    case label do
+      "mix compile" ->
+        # NOTE: `mix compile` updating event is from the compile tracer
+        # Then this tracer can be triggered by both `mix deps.compile` and `mix compile`.
+        get_token(state, "mix compile") || get_token(state, "mix deps.compile")
```
scohen, line 78: you've lost me here.

apps/server/lib/lexical/server/project/progress/state.ex
```diff
@@ -0,0 +1,68 @@
...
+    %{state | progress_by_label: progress_by_label}
+  end
+
+  def complete(%__MODULE__{} = state, project_progress(label: label, message: message)) do
+    token = get_token(state, label)
+
+    progress = Value.new(token, :end, message)
+    write(progress)
+
+    clear_progress_by_label(state, label)
```
scohen, line 42: same comment here:
```suggestion
    {progress, progress_by_label} = 
      Map.get_and_update(state.progress_by_label, label, fn _ -> :pop end)
    progress
    |> Value.finish()
    |> write()
   %{state | progress_by_label: progress_by_label}
```

This will remove the need for the `clear_progress_by_label`

apps/server/lib/lexical/server/project/progress/state.ex
```diff
@@ -0,0 +1,68 @@
...
+    progress_by_label = Map.delete(state.progress_by_label, label)
+    %{state | progress_by_label: progress_by_label}
+  end
+
+  defp get_token(state, label) do
+    previous_progress = state.progress_by_label[label]
+
+    if previous_progress do
+      previous_progress.token
+    end
```
scohen, line 66: this function should go away, but it's easier to accomplish with `get_in`
```suggestion
   get_in(state.progress_by_label, [label, :token])
```

apps/server/test/lexical/server/project/progress/state_test.exs
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  test "clear the token_by_label after received a complete event", %{project: project} do
+    state = State.new(project) |> State.begin(message("mix compile"))
+
+    %{progress_by_label: progress_by_label} =
+      State.complete(state, message("mix compile", "in 2s"))
+
+    assert progress_by_label == %{}
+  end
```
scohen, line 52: the happy path is well tested, but what about these scenarios:

1. A report message is received before a begin
2. An exception happens during a report message (Patch is great for this)
3. A complete message happens before a report

apps/server/test/lexical/server/project/progress/state_test.exs
```diff
@@ -0,0 +1,53 @@
...
+
+    %{progress_by_label: progress_by_label} =
+      State.report(state, message(label, "lib/my_module.ex"))
+
+    assert %Value{token: ^previous_token, message: "lib/my_module.ex", kind: :report} =
+             progress_by_label[label]
+  end
+
+  test "clear the token_by_label after received a complete event", %{project: project} do
+    state = State.new(project) |> State.begin(message("mix compile"))
```
scohen, line 46: nit: commit to the pipeline:
```suggestion
    state = project |> State.new() |> State.begin(message("mix compile"))
```

apps/server/test/lexical/server/project/progress/state_test.exs
```diff
@@ -0,0 +1,53 @@
...
+    label = "mix deps.get"
+    state = State.new(project) |> State.begin(message(label))
+
+    assert %Value{token: token, title: ^label, kind: :begin} = state.progress_by_label[label]
+    assert token != nil
+  end
+
+  test "it should be able to add a report event use the begin event token", %{project: project} do
+    label = "mix compile"
+    state = State.new(project) |> State.begin(message(label))
```
scohen, line 34: nit: commit to the pipeline:
```suggestion
    state = project |> State.new() |> State.begin(message(label))
```

apps/server/test/lexical/server/project/progress/state_test.exs
```diff
@@ -0,0 +1,53 @@
...
+  setup :with_a_project
+  setup do: patch(Transport, :write, :ok)
+
+  def message(label, message \\ nil) do
+    project_progress(label: label, message: message)
+  end
+
+  test "it should be able to add a begin event and put the new token", %{project: project} do
+    label = "mix deps.get"
+    state = State.new(project) |> State.begin(message(label))
```
scohen, line 26: nit: commit to the pipeline
```suggestion
    state = project |> State.new() |> State.begin(message(label))
```

apps/server/test/lexical/server/project/progress/state_test.exs
```diff
@@ -0,0 +1,53 @@
...
+
+  def with_a_project(_) do
+    project = project()
+    {:ok, %{project: project}}
+  end
+
+  setup :with_a_project
+  setup do: patch(Transport, :write, :ok)
+
+  def message(label, message \\ nil) do
```
scohen, line 20: i'd name this `progress` rather than `message`

apps/server/test/lexical/server/project/progress/state_test.exs
```diff
@@ -0,0 +1,53 @@
...
+  use ExUnit.Case
+  use Patch
+
+  def with_a_project(_) do
+    project = project()
+    {:ok, %{project: project}}
+  end
+
+  setup :with_a_project
+  setup do: patch(Transport, :write, :ok)
```
scohen, line 18: I'd just make this a setup block
```elixir
setup do 
  project = project()
  patch(Transport, :write, :ok)
 {:ok, project: project}
end
```

scohen (commented):
Getting close

apps/server/lib/lexical/server/project/progress/state.ex
scottming: Yeah, I'm stupid; I just forgot `get_in` is for this.

apps/server/lib/lexical/server/project/progress/state.ex
```diff
@@ -0,0 +1,68 @@
...
+
+  defstruct project: nil, progress_by_label: %{}
+
+  def new(%Project{} = project) do
+    %__MODULE__{project: project}
+  end
+
+  def begin(%__MODULE__{} = state, project_progress(label: label)) do
+    token = System.unique_integer([:positive])
+    progress = Value.new(token, :begin, label)
```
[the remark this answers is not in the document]
scottming: Yeah, agree. modified

apps/protocol/lib/lexical/protocol/notifications.ex
```diff
@@ -87,5 +87,17 @@ defmodule Lexical.Protocol.Notifications do
     end
   end
 
+  defmodule WorkDone.Progress.Create do
+    use Proto
+
+    defnotification "window/workDoneProgress/create", Types.WorkDone.Progress.Create.Params
+  end
+
+  defmodule Progress do
```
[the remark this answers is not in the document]
scottming: I don't think so.

> The base protocol offers also support to report progress in a generic fashion. This mechanism can be used to report any kind of progress including [work done progress](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workDoneProgress) (usually used to report progress in the user interface using a progress bar) and partial result progress to support streaming of results.

That means `progress` not only for `workDoneProgress`

apps/server/test/lexical/server/project/progress/state_test.exs
scottming: I added tests for 1 and 3, and for `2`, I don't know how to test it.

And do you mean to test the `with_progress` logic? That is too simple, I don't think we need to test the `try...after` keyword.

```elixir
    try do
      send(progress_pid, project_progress(label: label, stage: :begin))
      func.()
    after
      send(progress_pid, project_progress(label: label, stage: :complete))
    end
```

apps/protocol/lib/lexical/protocol/notifications.ex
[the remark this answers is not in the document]
scohen: alright, this API is weak sauce.

apps/server/lib/lexical/server/project/progress/value.ex
```diff
@@ -0,0 +1,39 @@
...
+  def report(_, _) do
+    nil
+  end
+
+  def complete(%__MODULE__{token: token}, message) do
+    %__MODULE__{token: token, kind: :end, message: message}
+  end
+
+  def to_progress(%__MODULE__{kind: :report = kind} = value) do
+    p_value = WorkDone.Progress.Report.new(kind: kind, message: value.message)
```
scohen, line 26: i'm not a big fan of short prefixes like `p_` I'm ok with:
```elixir
Notifications.Progress.new(token: value.token,
  value: WorkDone.Progress.Report.new(kind: value.kind, message: value.message)
  )
  ```
Also, the `kind: :report = kind` is kinda clunky, you don't need to capture the kind, just use value.kind.

File: apps/server/lib/lexical/server/project/progress/value.ex [hunk 39]
```diff
@@ -0,0 +1,39 @@
...
+
+  def report(_, _) do
+    nil
+  end
+
+  def complete(%__MODULE__{token: token}, message) do
+    %__MODULE__{token: token, kind: :end, message: message}
+  end
+
+  def to_progress(%__MODULE__{kind: :report = kind} = value) do
```
scohen, line 25: I'd call this  function`to_protocol`

apps/server/lib/lexical/server/project/progress/value.ex
```diff
@@ -0,0 +1,39 @@
...
+    p_value = WorkDone.Progress.Report.new(kind: kind, message: value.message)
+    Notifications.Progress.new(token: value.token, value: p_value)
+  end
+
+  def to_progress(%__MODULE__{kind: :begin = kind} = value) do
```
scohen, line 30: same comments as above. 
`to_protocol`, don't capture kind, just build the whole struct rather than in pieces.

apps/server/lib/lexical/server/project/progress/value.ex
```diff
@@ -0,0 +1,39 @@
...
+    p_value = WorkDone.Progress.Begin.new(kind: kind, title: value.title)
+    Notifications.Progress.new(token: value.token, value: p_value)
+  end
+
+  def to_progress(%__MODULE__{kind: :end = kind} = value) do
```
scohen, line 35: same comments as above.

apps/server/test/lexical/server/project/progress/state_test.exs
scohen: why don't you need to test the try/after? You're assuming that the code exists, you should test functionality, not implementation. 

Someone could go in tomorrow and change it somehow so it'd break, without a test, how would we know?

scohen (approved):
Make the suggested changes and you're good.

apps/server/test/lexical/server/project/progress/state_test.exs
scottming: makes sense.

Issue #122, opened 2023-04-25
[Completion] if a module has struct and submodule, we should complete both struct and module 


scottming:
<img width="438" alt="image" src="https://user-images.githubusercontent.com/12830256/234432504-bffadf5b-68fa-443c-957b-d7eed258114a.png">

Currently, we only complete the struct, which is annoying.

scohen:
I'd argue that in your example, you shouldn't complete the module and only complete the struct, since your intention is to use a struct (the line starts with `%`) See if #137 works to fix this.

Issue #123, opened 2023-04-25
[Completion] when complete a struct, we should jump the cursor to the `{|`


scottming:
The priority of this one can be higher than #122  because when you are inside the curlies, you can quickly delete the `{` or complete a field.

like: `%Stru|` should complete as `%Struct{|}`

PR #137, opened 2023-05-02
Improve struct completions

scohen:
Two commits here. One improves struct completions, and the other makes them snippets to put the cursor inside the curly braces when you complete.

### DIFF
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
@@ -22,9 +22,10 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
       struct_reference? and defines_struct? ->
         struct_completion(builder, env, module.name)
 
-      struct_reference? and Intelligence.child_defines_struct?(env.project, module.full_name) ->
+      struct_reference? and
+          immediate_descentent_defines_struct?(env.project, module.full_name) ->
         env.project
-        |> Intelligence.child_struct_modules(module.full_name)
+        |> immediate_descendent_struct_modules(module.full_name)
         |> Enum.map(fn child_module_name ->
           local_name = local_module_name(module.full_name, child_module_name)
           struct_completion(builder, env, local_name)
@@ -88,4 +89,12 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
 
   defp strip_leading_period(<<".", rest::binary>>), do: rest
   defp strip_leading_period(string_without_period), do: string_without_period
+
+  defp immediate_descentent_defines_struct?(%Lexical.Project{} = project, module_name) do
+    Intelligence.descendent_defines_struct?(project, module_name, 1..2)
+  end
+
+  defp immediate_descendent_struct_modules(%Lexical.Project{} = project, module_name) do
+    Intelligence.descendent_struct_modules(project, module_name, 1..2)
+  end
 end
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
@@ -15,45 +15,73 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
   end
 
   defp do_translate(%_{} = module, builder, %Env{} = env) do
-    detail = builder.fallback(module.summary, module.name)
     struct_reference? = Env.struct_reference?(env)
     defines_struct? = Intelligence.defines_struct?(env.project, module.full_name)
 
-    add_curlies? =
-      defines_struct? and String.contains?(Env.last_word(env), "%") and
-        not String.contains?(env.suffix, "{")
+    cond do
+      struct_reference? and defines_struct? ->
+        struct_completion(builder, env, module.name)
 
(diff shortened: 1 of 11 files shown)


scohen:
Fixes #122 and #123

apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
```diff
@@ -15,45 +16,73 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
...
+    end
+  end
 
-        true ->
-          {module.name, ""}
-      end
+  defp struct_completion(builder, %Env{} = env, module_name) do
+    last_word = Env.last_word(env)
+
+    Logger.info("last word is: #{last_word}")
```
scohen, line 43: Remove

scohen, line 43:
```suggestion
```

scottming:
There is a bug when you want to complete the child module

```elixir
17:41:08.894 [info] last word is: %L
17:41:08.894 [info] add percent for exical.Project? true add curlies: true
17:41:08.895 [warn] Emitting completions: [#Protocol.Types.Completion.Item<[detail: "exical.Document (Struct)", insert_text: "%exical.Document{$1}", insert_text_format: :snippet, kind: :struct, label: "%exical.Document", sort_text: "exical.Document"]>, #Protocol.Types.Completion.Item<[detail: "exical.Project (Struct)", insert_text: "%exical.Project{$1}", insert_text_format: :snippet, kind: :struct, label: "%exical.Project", sort_text: "exical.Project"]>, #Protocol.Types.Completion.Item<[detail: "Linked lists hold zero, one, or more elements in the chosen order.", insert_text: "List", kind: :module, label: "List"]>, #Protocol.Types.Completion.Item<[detail: "A logger for Elixir applications.", insert_text: "Logger", kind: :module, label: "Logger"]>]
17:41:08.895 [info] Got [{"detail":"exical.Document (Struct)","insertText":"%exical.Document{$1}","insertTextFormat":2,"kind":22,"label":"%exical.Document","sortText":"exical.Document"},{"detail":"exical.Project (Struct)","insertText":"%exical.Project{$1}","insertTextFormat":2,"kind":22,"label":"%exical.Project","sortText":"exical.Project"},{"detail":"Linked lists hold zero, one, or more elements in the chosen order.","insertText":"List","kind":9,"label":"List"},{"detail":"A logger for Elixir applications.","insertText":"Logger","kind":9,"label":"Logger"}]
```

image

scottming:
2. when you type `%Lexic` still no completions

image

3. type `%Lexical.D` still no module completions, only one struct completion(and this is not a snippet)

image

It should to complete both the struct and module.

4. when my hint is starts with `%`, the completions should not show me the functions

image






BTW, completions are complicated.

scottming:
Another point to consider is whether you really need the cursor to be inside the braces when you are in the parameters of a function.

Because we may write a lot of `def func(%__MODULE__{} = module, _)` and `def handle_info(_, %State{}=state)` statements, compared to situations like `%Struct{|}`, seems like we use `%Struct{}|` more often when in parameters.

scohen:
@scottming i can't fix the no completions case, if elixir sense doesn't give me completions, there's nothing I can do.

scohen:
> Another point to consider is whether you really need the cursor to be inside the braces when you are in the parameters of a function

It's pretty common to pattern match on struct fields in function heads, and easy to tab out of the curly braces. I'd favor consistency here, since people will build up muscle memory about how to handle these things. Having a completion do one thing in one context and something else in another will be tough.

scohen:
> type %Lexical.D still no module completions, only one struct completion(and this is not a snippet)

if there's a leading `%`, no module completions are offered, this is intentional, since you're in a struct reference. It should be a snippet though. I think this is because this is the one case where elixir sense offers a struct.

scohen:
> There is a bug when you want to complete the child module

Yep, that's a bug. Can you verify that it only happens when completing the top level module with no dots? For example, `%Lexical|` shoes the bug, but `%Lexical.D|` works.

scohen:
So right now, struct completions only offer suggestions for the immediate children of a module. Would it be better if they offered suggestions for all descendants (children, grandchildren, great grandchildren, etc)?

scottming:
> if there's a leading %, no module completions are offered, this is intentional, since you're in a struct reference. It should be a snippet though. I think this is because this is the one case where elixir sense offers a struct.

not a snippet.

scottming:
> Yep, that's a bug. Can you verify that it only happens when completing the top level module with no dots? For example, %Lexical| shoes the bug, but %Lexical.D| works.

*     `%Lexical.Document.Store.St|` -> `    %Lexical.Document.Store.Store{|}` works
* `%Lexical.Docu` -> `%Lexical.Document{}|` There are cursor and grandson module issues.

scottming:
> So right now, struct completions only offer suggestions for the immediate children of a module. Would it be better if they offered suggestions for all descendants (children, grandchildren, great grandchildren, etc)?

right, And I think `%struc|` completions only offering struct is better, ignoring functions, fields, and so on.

scottming:
Should we also keep the icon of the snippet consistent? Its normal icon looks like this:

image

scohen:
The kind is based on the type of suggestion, not whether it's a snippet. All the items there are macros

scohen:
@scottming, I couldn't tell from your response, but do you think it's better to return all descendant structs, or just the immediate children?

scottming (approved):

PR #143, opened 2023-05-08
Improve intelligence api

scohen:
Refined generation specifiers

The previous specifiers were a bit opaque, with a range given to the functions. Hopefully this is a little clearer

### DIFF
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
@@ -91,10 +91,10 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
   defp strip_leading_period(string_without_period), do: string_without_period
 
   defp immediate_descentent_defines_struct?(%Lexical.Project{} = project, module_name) do
-    Intelligence.descendent_defines_struct?(project, module_name, 1..2)
+    Intelligence.defines_struct?(project, module_name, to: :grandchild)
   end
 
   defp immediate_descendent_struct_modules(%Lexical.Project{} = project, module_name) do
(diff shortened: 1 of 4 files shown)


apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
```diff
@@ -91,10 +91,10 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
   defp strip_leading_period(string_without_period), do: string_without_period
 
   defp immediate_descentent_defines_struct?(%Lexical.Project{} = project, module_name) do
-    Intelligence.descendent_defines_struct?(project, module_name, 1..2)
+    Intelligence.defines_struct?(project, module_name, to: :grandchild)
   end
 
   defp immediate_descendent_struct_modules(%Lexical.Project{} = project, module_name) do
-    Intelligence.descendent_struct_modules(project, module_name, 1..2)
+    Intelligence.collect_struct_modules(project, module_name, to: :grandchild)
```
scottming, line 98: After examining the implementation of this function in the State module, I found that it has no side effects, meaning that it will not change the state but only calculate the collected module information.

Therefore, would it be better to name it `collected_struct_modules` instead?

scottming (approved):

apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
scohen: I don't think so; -ed is past tense in english, which would imply that the collection has already happened, and if that's the case, it would be reasonable to assume that this was returning a cached value or something. Since it's doing work, I think collect is the right name here. This is very much like the `filter` function in Enum.

PR #144, opened 2023-05-09
Added completion for bitstring options

### LINKED ISSUE #142: Crash while editing typespecs
```
11:00:36.434 [error] ** (ErlangError) Erlang error: {:exception, :function_clause, [{Lexical.RemoteControl.Completion.Result, :from_elixir_sense, [%{name: "integer", type: :bitstring_option}], [file: 'lib/lexical/remote_control/completion/result.ex', line: 133]}, {Enum, :"-map/2-lists^map/1-0-", 2, [file: 'lib/enum.ex', line: 1658]}, {Enum, :"-map/2-lists^map/1-0-", 2, [file: 'lib/enum.ex', line: 1658]}, {Lexical.RemoteControl.Completion, :elixir_sense_expand, 2, []}]}
    (kernel 8.5.3) erpc.erl:702: :erpc.call/5
    (server 0.1.0) lib/lexical/server/code_intelligence/completion.ex:74: Lexical.Server.CodeIntelligence.Completion.completions/3
    (server 0.1.0) lib/lexical/server/code_intelligence/completion.ex:34: Lexical.Server.CodeIntelligence.Completion.complete/4
    (server 0.1.0) lib/lexical/server/provider/handlers/completion.ex:11: Lexical.Server.Provider.Handlers.Completion.handle/2
    (server 0.1.0) lib/lexical/server/provider/queue.ex:100: anonymous fn/2 in Lexical.Server.Provider.Queue.State.as_task/2
    (elixir 1.14.3) lib/task/supervised.ex:89: Task.Supervised.invoke_mfa/2
    (elixir 1.14.3) lib/task/supervised.ex:34: Task.Supervised.reply/4
    (stdlib 4.2) proc_lib.erl:240: :proc_lib.init_p_do_apply/3
```

scohen:
While fixing #142, I realized that we don't complete bitstring options. This PR fixes that omission and now bitstring completions are offered while inside of a binary pattern. This required doing a bit of tokenization, but the elixir tokenizer is available to us, which is nice.

Fixes #142


Tuxified (approved):
Looks good, the only concern I had was possible clashes with Bitwise operators ( <<< and >>> ) but since we're using Elixirs tokenizer, I think there's no problem

scohen:
I was unable to get this to work with `eglot` and vscode reliably, so I switched my emacs back to `lsp-mode`, and by using text edits, I was able to get this working. 
I thing eglot has serious issues regarding thinking too much for itself when it sees text edits. It was applying edits that were outside of its edit range, which is explicitly forbidden by the spec. 

Welp, back to lsp-mode with _its_ bugs.

PR #145, opened 2023-05-09
Improved function capture completions

scohen:
Function captures used to only complete `&Mod.fun/arity`, which is _ok_, but sometimes you want to do `&Mod.fun(&1)`. This PR adds two completion options for each function, one for the "complete" capture, and another that lets you add arguments.

I'm not sure if I like the UX on this yet, but I do like that I can pick between the two options. I'm open to suggestions about how to improve the UX.


scottming:
2. you missed the `kind` field

It should be both `function`, or `function`, `snippet`(only for [(Capture with arguments)])

scottming (approved):

PR #152, opened 2023-05-10
Add test matrix and bump OTP version

### LINKED ISSUE #146: Build matrix release
based on this comment: https://github.com/lexical-lsp/lexical/pull/111#discussion_r1189197244

> We need to build a matrix like this:

```
Erlang: Elixir
24 1.13, 1.14
25 1.13, 1.14
26 1.13, 1.14, (1.15)
```

wingyplus:
Closes #146


.github/workflows/elixir.yml
```diff
@@ -18,15 +18,52 @@ permissions:
...
+
+      - name: Install dependencies
+        run: |
+          mix deps.get
+
+      # Step: Check that the checked in code has already been formatted.
+      # This step fails if something was found unformatted.
+      # Customize this step as desired.
+      - name: Check Formatting
+        run: mix format --check-formatted
```
scohen, line 56: Why this change? We have a precommit hook and testing task that should prevent unformatted code from being merged into main. 

Also, I don't think you need to fetch and compile deps just to check formatting

wingyplus: @scohen I found that format style diff on Elixir version 1.13 T_T (https://github.com/lexical-lsp/lexical/actions/runs/4938201783/jobs/8827646922). So to avoid this I extract to a new job.

wingyplus: Or add some boolean in matrix (like `check-format: true | false`) and `if` at this step would be better?

.github/workflows/release.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: Set up Elixir
         uses: erlef/setup-beam@v1
         with:
-          otp-version: "25.2.1"
+          otp-version: "25.3.2"
```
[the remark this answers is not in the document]
wingyplus: I think it should be. Prefer edit in this PR or open a new one?

wingyplus:
I'll work again tomorrow. The action is unstable now (https://www.githubstatus.com/)

.github/workflows/release.yml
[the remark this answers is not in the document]
scohen: Do it here, though we should set it to the lowest value we support. 
1.13 / erl 24?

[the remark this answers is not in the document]
wingyplus: I think we should verify formatting on elixir 1.14 because we use 1.14 for development (reference from .tools-version).

.github/workflows/elixir.yml
scohen: how about we just format it so that 1.13 doesn't complain?

.github/workflows/release.yml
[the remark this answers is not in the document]
scohen: If we're going to support 1.13, we need to develop on that. Using a later version will only cause us pain.

wingyplus:
Just note on Elixir 1.13 issues:

1. Mix format diff on config/prod.exs (Fixed in 
2. Dialyzer failure on this version 
```
lib/mix/tasks/lsp/data_model.ex:82:no_return
Function collect_references/3 has no local return.
________________________________________________________________________________
lib/mix/tasks/lsp/data_model.ex:82:opaque_match
Attempted to pattern match against the internal structure of an opaque term.

Type:
MapSet.t(_)

Pattern:
%MapSet{}

This breaks the opaqueness of the term.

________________________________________________________________________________
lib/mix/tasks/lsp/data_model.ex:86:opaque_match
Attempted to pattern match against the internal structure of an opaque term.

Type:
MapSet.t(_)

Pattern:
%MapSet{}

This breaks the opaqueness of the term.

________________________________________________________________________________
lib/mix/tasks/lsp/mappings/generate.ex:78:call_to_missing
Call to missing or private function Keyword.replace_lazy/3.
________________________________________________________________________________
done (warnings were emitted)
Halting VM with exit status 2
``` 
3. Found warning on compiling proto apps due to  `Keyword.replace_lazy` introduced in Elixir 1.14

config/prod.exs
```diff
@@ -0,0 +1 @@
+
```
wingyplus, line 1: Elixir 1.13 format like this. On Elixir 1.14, it removed this line.

wingyplus:
@scohen I found a lot of test failures on Elixir 1.13 https://github.com/lexical-lsp/lexical/actions/runs/4950410802/jobs/8853977395?pr=152.  Need to fix the test first before adding Elixir 1.13 to the CI.

PR #154, opened 2023-05-11
Enable --warnings-as-errors in CI

scottming:


apps/server/mix.exs
```diff
@@ -56,6 +56,7 @@ defmodule Lexical.Server.MixProject do
       {:protocol, in_umbrella: true},
       {:remote_control, in_umbrella: true, runtime: false},
       {:sourceror, "~> 0.11"},
+      {:recode, "~> 0.4", only: :dev},
```
scohen, line 59: Why's this here?

scottming: > Try append a temporary deps to clear CI cache

Wrote that in a git message, just temporary

scohen (approved):

PR #158, opened 2023-05-11
Normalize naming

scohen:


apps/common/test/lexical/format_test.exs
scottming, line 1: you need to change the file name too.

scottming (approved):

PR #160, opened 2023-05-12
Addressed deadlocks in Document Store

scohen:
I was getting sporadic deadlocks in the document store since I moved to lsp-mode (it sends a *lot* more didChange updates than eglot), so I'm trying moving writes out of the genserver to see if that helps (it seems to).

I also noticed some troubling messages about not being able to apply didChange events, but that was due to an incorrect log message in the server state, which is also fixed here.

### DIFF
--- a/apps/common/lib/lexical/document/store.ex
+++ b/apps/common/lib/lexical/document/store.ex
@@ -2,8 +2,8 @@ defmodule Lexical.Document.Store do
   @moduledoc """
   A backing store for source file documents
 
-  This implementation stores documents in ETS, and partitions read and write operations. Read operations are served
-  immediately by querying the ETS table, while writes go through a GenServer process (which is the owner of the ETS table).
+  This implementation stores documents in a public ETS table. Since documents are versioned, it's OK if we get writes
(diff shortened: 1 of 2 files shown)


apps/common/lib/lexical/document/store.ex
```diff
@@ -70,21 +76,21 @@ defmodule Lexical.Document.Store do
...
           normalize_error(error)
       end
     end
 
-    def update(%__MODULE__{} = store, uri, updater_fn) do
-      with {:ok, _, new_store} <- get_and_update(store, uri, updater_fn) do
-        {:ok, new_store}
+    def update(uri, updater_fn) do
+      with {:ok, _} <- get_and_update(uri, updater_fn) do
+        :ok
```
scottming, line 93: So this means after this pr, we won't update the store genserver state when call `get_and_update` or `update`, right?

scottming (approved):

apps/common/lib/lexical/document/store.ex
scohen: We weren't updating the genserver state before.

PR #161, opened 2023-05-13
Fix CreateWorkDoneProgress for VScode and Emacs

### LINKED ISSUE #159: WorkDoneProgress.Create is a request, not a notification
I'm getting the following error under lsp-mode 

> Warning (lsp-mode): Unknown notification: window/workDoneProgress/create

This is true, `window/workDoneProgress/create` is a request, not a notification, but it's a request that comes from the server (there are a few).

It gets even more complicated:
[The spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#serverInitiatedProgress) says that we need to check the client's capabilities before we can even initiate this request.

scottming:
Fixes #159 and https://github.com/lexical-lsp/vscode-lexical/issues/14

### VScode

https://github.com/lexical-lsp/lexical/assets/12830256/1fac0de2-55c2-446e-a9a4-b2ff13e10697

One difference between VScode and neovim is that VScode does not display short-term progress.

### Emacs `lsp-mode` + `doom-modeline`

https://github.com/lexical-lsp/lexical/assets/12830256/6f88540f-729f-464e-bfc6-41dcc57f7cdc


apps/protocol/lib/lexical/protocol/requests.ex
```diff
@@ -33,6 +33,12 @@ defmodule Lexical.Protocol.Requests do
     defrequest "textDocument/definition", Types.Definition.Params
   end
 
+  defmodule CreateWorkDoneProgress do
```
scottming, line 36: As `CreateWorkDoneProgress` is a request, I think keeping the name consistent with the title is more appropriate.

scottming: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_workDoneProgress_create

apps/protocol/lib/lexical/protocol/json_rpc.ex
```diff
@@ -30,11 +31,22 @@ defmodule Lexical.Protocol.JsonRpc do
...
     Requests.decode(method, request)
   end
 
   defp do_decode(%{"method" => method} = notification) do
     Notifications.decode(method, notification)
   end
+
+  defp do_decode(%{"error" => %{"code" => -32_601}}) do
+    # NOTE: vscode can't handle CreateWorkDoneProgress until `Requests.Initialize` is done,
+    # That's why the magic error code is here.
```
scohen, line 45: This is the wrong solution. No one should handle *any* or responses until the initialize response has been sent. Doing work prior to that is a lexical bug.

apps/protocol/lib/lexical/protocol/requests.ex
scohen: ok.

apps/server/lib/lexical/server/configuration/support.ex
```diff
@@ -39,20 +40,37 @@ defmodule Lexical.Server.Configuration.Support do
...
   end
 
+  def get do
+    :persistent_term.get({__MODULE__, :config_support}, %__MODULE__{})
+  end
+
   defp bool(b) when b in [true, false], do: b
   defp bool(_), do: false
+
+  defp set(config) do
```
scohen, line 73: shouldn't this be handled at the level of `Configuration` and not `Configuration.Support` ?

apps/server/lib/lexical/server/project/progress/state.ex
```diff
@@ -42,13 +44,21 @@ defmodule Lexical.Server.Project.Progress.State do
...
   defp write(%{token: token} = progress) when not is_nil(token) do
-    progress |> Value.to_protocol() |> Transport.write()
+    if client_support_work_done?() do
+      progress |> Value.to_protocol() |> Transport.write()
+    end
   end
 
   defp write(_), do: :ok
+
+  defp client_support_work_done? do
```
scohen, line 61: nit: `client_supports_work_done?` but I feel that this should be done inside the configuration so we can have:
```elixir
Configuration.supports?(:work_done_progress)
```
and the implementation would be:
```elixir
@supports_keys ~w(:work_done_progress)a

def supports?(key) when key in @supports_keys do 
  get_in(Configuration.get(), [:support, key], false)
end
```

scohen (approved):
revert the changes to project and you're good.

PR #162, opened 2023-05-14
Fix diagnostic for missing fields due to @enforce_keys

scottming:
image


scohen (commented):
Thanks for writing the test.

scohen (approved):

PR #166, opened 2023-05-16
Fixed infinite loop in document updates

scohen:
After I switched to lsp-mode, my editor was hanging whenever I would update a file using git, and never when I made changes to it in the editor itself. This is because, unlike eglot, when lsp-mode detects a file system change, it sends the entire document over as a replace update, and not an update with many text edits. This would cause the document to go into an infinite loop, and the request would never end, and lexical would become unresponsive.

The reason for this is that the ContentChangeEvent is a type alias with multiple types, one with just a "text" key and one with "text" and a "range" key. The version with a "text" and "range"" key were properly converted to a native Edit by the `Any` convertible implementation, but the version with just the text wasn't properly converted, but the conversion succeeded because the `Any` code thought it had converted it properly. The solution is to give a `Convertible` implementation to the specific `ContentChangeEvent` and have that just build a simple edit.

I also think that the "deadlocks" I was experiencing prior were due to this bug, and I'm going to back out those changes, as they'll yield a more correct codebase.

This is what I get for fixing a problem without actually understanding it first.

### DIFF
--- a/apps/common/lib/lexical/document/store.ex
+++ b/apps/common/lib/lexical/document/store.ex
@@ -2,8 +2,8 @@ defmodule Lexical.Document.Store do
   @moduledoc """
   A backing store for source file documents
 
-  This implementation stores documents in a public ETS table. Since documents are versioned, it's OK if we get writes
-  out of order, as they'll be rejected if the versions don't match up.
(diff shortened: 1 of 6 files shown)


scottming:
> This is what I get for fixing a problem without actually understanding it first.
NOTED

scottming (approved):

PR #169, opened 2023-05-18
Don't crash if env contains no words

doughsay:
VSCode will send  no words if you just hit `ctrl+space` on an empty line, which crashes the completions.  This fixes it.  I'm not entirely sure how to write a test for this, as when I tried `assert complete(project, "|") ...` it returned `{:error, :out_of_bounds}`...


scohen:
I was literally working on removing this function

doughsay:
fixed in https://github.com/lexical-lsp/lexical/pull/172

PR #170, opened 2023-05-18
Added fragment capabilities to document

scohen:
While talking with Jose Valim about how he could help lexical, we discussed giving a document fragment to a function in Code.Fragment. It seemed to me at the time that one of the functions already there might fit the bill, so I coded up the ability to request a fragment of a document using positions.

Unfortunately, the function in question won't work for us, but I felt that document fragments might be helpful in the future, so here's a PR.

### DIFF
--- a/apps/server/test/document_test.exs
+++ b/apps/server/test/document_test.exs
@@ -638,4 +638,90 @@ defmodule Lexical.DocumentTest do
                ])
     end
   end
+
+  def document(text) do
+    new(file_uri(), String.trim(text), 0)
+  end
+
+  describe "fragment/2" do
+    test "works with out of bounds lines" do
+      doc =
+        """
+        one
+        two
+        three
+        """
+        |> document()
+        |> Document.fragment(new_position(3, 0))
+
+      assert "one\ntwo\nthree" == doc
+    end
+
+    test "can omit entire lines" do
+      doc =
+        """
+        one
+        two
+        three
+        """
+        |> document()
+        |> Document.fragment(new_position(2, 0))
+
+      assert "one\ntwo\n" == doc
+    end
+
+    test "can truncate the end of a line" do
+      doc =
+        """
+        one
+        two
+        three
+        """
+        |> document()
+        |> Document.fragment(new_position(2, 2))
+
+      assert "one\ntwo\nth" == doc
+    end
+  end
+
+  describe "fragment/3" do
+    test "can omit an entire line" do
+      doc =
+        """
+        one
+        two
+        three
+        """
+        |> document()
+        |> Document.fragment(new_position(1, 0), new_position(3, 0))
+
+      assert "two\nthree" == doc
+    end
+
+    test "can truncate the beginning of a line" do
+      doc =
+        """
+        one
+        two
+        three
+        """
+        |> document()
+        |> Document.fragment(new_position(0, 2), new_position(3, 0))
+
+      assert "e\ntwo\nthree" == doc
+    end
+
+    test "can truncate both the beginning and end of a line" do
+      doc =
+        """
+        one
+        two
+        three
+        """
+        |> document()
+        |> Document.fragment(new_position(0, 2), new_position(1, 1))
+
+      assert "e\nt" == doc
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


apps/server/test/document_test.exs
```diff
@@ -638,4 +638,90 @@ defmodule Lexical.DocumentTest do
...
+
+    test "can truncate the beginning of a line" do
+      doc =
+        """
+        one
+        two
+        three
+        """
+        |> document()
+        |> Document.fragment(new_position(0, 2), new_position(3, 0))
```
scottming, line 709: To be honest, I was confused when I first saw this test. My inherent impression of `position` is 1-based. After checking the product code, I found that the type of `position` here is actually `Lexical.Protocol.Types.Position(Convertible.t())`.

scottming: Moreover, shouldn't we add some simple tests for the type of Lexical.Position?

scohen: check out the imports, these are LSP positions, not Document positions

Which tests should we add for Lexical.Position? There's no logic in that module at all.

scottming: Sorry, I mean `Lexical.Document.Position`, this aliased at product code: https://github.com/lexical-lsp/lexical/blob/3f2759d9108d15fd680f007c67a3ec3e5cc0dfed/apps/common/lib/lexical/document.ex#L13

scottming (approved):

apps/server/test/document_test.exs
scottming: Just expressing my confusion, feel free to handle it as you see fit

scohen: I’m still confused as to what tests we should add

scottming: I mean one test similar to this:

```elixir
iex(11)> text =
...(11)>  """
...(11)>  one
...(11)>  two
...(11)>  three
...(11)>  """ |> String.trim()
"one\ntwo\nthree"
iex(12)>  doc = Document.new("file", text, 1)
#Lexical.Document<
  path: "/Users/scottming/Code/lexical/file",
  version: 1,
  dirty?: false,
  lines: %Lines<"one..."(3 lines)>,
  ...
>
iex(13)>  doc |> Document.fragment(Document.Position.new(1, 3), Document.Position.new(4, 1))
"e\ntwo\nthree"
iex(14)> alias Lexical.Document
Lexical.Document
iex(15)> text =
...(15)>  """
...(15)>  one
...(15)>  two
...(15)>  three
...(15)>  """ |> String.trim()
"one\ntwo\nthree"
iex(16)>  doc = Document.new("file", text, 1)
#Lexical.Document<
  path: "/Users/scottming/Code/lexical/file",
  version: 1,
  dirty?: false,
  lines: %Lines<"one..."(3 lines)>,
  ...
>
iex(17)>  doc |> Document.fragment(Document.Position.new(1, 3), Document.Position.new(4, 1))
"e\ntwo\nthree"
```

Although the result of your current test is the same, you have indicated in this [line](https://github.com/lexical-lsp/lexical/blob/3f2759d9108d15fd680f007c67a3ec3e5cc0dfed/apps/common/lib/lexical/document.ex#L106) that `Document.fragment` also supports `Document.Position.t()`. I believe that supporting 1-based positions is part of this functionality, so I think adding one test for it makes sense.

scohen: I can add those tests, but the way that the document works is that is uses the convertible protocol to change `Lexical.Protocol.Types.Position`s into `Lexical.Document.Position`s and works on those. In other words, the tests using the protocol types is already testing that functionality, just with a conversion step.

scottming: Ok, thank you for your explanation. I found out that maybe I didn't read the code carefully enough. You actually converted the position [here](https://github.com/lexical-lsp/lexical/pull/170/files#diff-7358ad5bcbada13c3b6c2b5985ee2fb572796abece6b39e2bde98c4da896f355R113) at the beginning, so I think adding more tests seems to be meaningless.

you're good.

PR #172, opened 2023-05-19
Removed string-based completion env operations

scohen:
I was using naive word-based heuristics to decide several things when completions were requested. This PR removes that approach in favor of a syntax-based approach that should be more robust.

FYI @doughsay

### DIFF
--- a/apps/common/lib/lexical/completion/environment.ex
+++ b/apps/common/lib/lexical/completion/environment.ex
@@ -11,7 +11,6 @@ defmodule Lexical.Completion.Environment do
   @callback in_context?(t, context_type) :: boolean
 
   @callback empty?(maybe_binary) :: boolean
-  @callback last_word(t) :: String.t()
   @callback prefix_tokens(t) :: [lexer_token]
   @callback prefix_tokens(t, token_count) :: [lexer_token]
 end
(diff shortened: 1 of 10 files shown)


doughsay:
This does fix the same issue I tried fixing in https://github.com/lexical-lsp/lexical/pull/169 🎉 

...but it still fails on an edge-case that my fix also still did. I can't comment directly on this line because it's out of scope of this PR: https://github.com/lexical-lsp/lexical/pull/172/files#diff-03833d69f557e938be9ebf0f8f547f6d0ea28c928f479b10fb8feba7bd99fa9cR34

```elixir
{:ok, env} = Env.new(project, document, position)
```

It's possible (in VSCode at least) to cause this to return `{:error, :out_of_bounds}` if you open a completely blank new document and request completions (i.e. `ctrl+space` in vscode) without any newlines or spaces in the document.

scottming:
I tested this branch in Neovim and it works very well.

But now, do we also provide auto-completion for just one character?

image

scottming:
So the last commit is to enable autocompletion for single characters, right? I have no objections to that because, as @doughsay  pointed out, it's fast and sometimes helpful.

scottming (approved):

PR #173, opened 2023-05-20
Credo integration

scottming:
https://github.com/lexical-lsp/lexical/assets/12830256/3a79e40e-ee63-4539-814e-44c60454a21b

(ps. Just a reminder that my diagnostic is set to be hidden by default, represented only by an underscore. This is intentionally done on my part.)

I love this feature. It's very fast, taking only 7ms for real-time credo on a single file from the second time onwards, while lexical for the entire project takes only 1s.

The functionality is now very stable and it can recognize `.credo.exs`. Currently, all you need to do is add the dependency [scottming/lexical-credo](https://github.com/scottming/lexical-credo),  in your project and you'll have enhanced capabilities for lexical.

However, there are some issues that I haven't figured out yet. For example, whether "behavior" is appropriate or not, and the Credo module in the protocol app seems inappropriate. Additionally, some parts of the code may be a bit rough. I'm quite busy these days and may need until next week to optimize it further.


scohen:
This needs a lot of work. 
I think the way plugins work is backwards; we want other projects to define plugins and we detect them. Lexical needs to publish a `lexical-plugin` package that other projects use that allows them to define different kinds of plugins, which will be detected when we compile that project. You can see the issues with the current plugin system by credo being defined in the `protocol` app.

The goal for plugins is that lexical shouldn't have to use credo as a dependency, nor should plugins need a genserver to be maintained. I also don't understand why plugins need to be handled differently than our code when it comes to diagnostics, etc. That makes the code paths much more complicated and I don't see the benefit.  A plugin should be able to do whatever we do presently, the way we do it.

scottming:
I fully agree with this point, and I also strongly agree with the goal mentioned in your comments.

But let me first answer this question:

> I also don't understand why plugins need to be handled differently than our code when it comes to diagnostics, etc

The main purpose of that place is to prevent plugins from clearing Mix's diagnostics. Diagnostics will be overwritten, so we need to ensure that each plugin only clears its own diagnostics.

scohen:
Going to close this, as I think plugins will make this much easier.

PR #175, opened 2023-05-22
Add Nix Flake

hauleth:
This adds Nix Flake which (almost) allows usage of the Lexical via `nix run github:lexical-lsp/lexical -- version` (~~the "almost" part is that one need to define `RELEASE_COOKIE=/dev/null` beforehand as `releases/COOKIE` file is removed by Nix for now~~ that is resolved). This allows user of Nix to use Lexical in their environment without writing their own derivation (Nix name for package).

### DIFF
--- a/nix/lexical.nix
+++ b/nix/lexical.nix
@@ -1,0 +1,34 @@
+{ mixRelease
+, fetchMixDeps
+, erlang
+}: mixRelease rec {
+  pname = "lexical";
+  version = "development";
+
+  src = ./..;
+
+  mixFodDeps = fetchMixDeps {
+    inherit pname;
+    inherit version;
+
+    src = ./..;
+
+    sha256 = "sha256-a715z1ti1J0kKzLcIS1UuFpPAMEx2VlFpkmZrRTSWh4=";
+  };
+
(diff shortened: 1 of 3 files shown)


scohen:
@hauleth is this ready to go?

hauleth:
I think so. But there probably would need to be additional CI check as well that would handle updating the dependencies hash. But that can be added later.

nix/lexical.nix
```diff
@@ -0,0 +1,34 @@
+{ mixRelease
+, fetchMixDeps
+, erlang
+}: mixRelease rec {
+  pname = "lexical";
+  version = "development";
+
+  src = ./..;
```
viniciusmuller, line 8: What about using [`pkgs.lib.cleanSource`](https://github.com/NixOS/nixpkgs/blob/master/lib/sources.nix#L55) on these `src` references? It cleans unnecessary stuff from the resulting derivation, such as git repositories.

There's also a way to ignore items on the gitignore on the resulting derivation using [`pkgs.gitignore`](https://nixos.org/manual/nixpkgs/stable/#sec-pkgs-nix-gitignore-usage)

PR #176, opened 2023-05-22
Do not create intermediate binaries

hauleth:
This reduces memory footprint as we can compute the code units count by using simple integer comparisons. This should result in faster code, but I haven't benchmarked it yet.


scohen (approved):
Looks good; I was a little concerned about fencepost problems to implement it this way, and thought "well, this will only be called on lines that had unicode on them", which given elixir source code, is fairly few and far between.

PR #179, opened 2023-05-23
Fixed code actions / improved code mod api

scohen:
Code actions were kind of broken, they were unable to process all elements of an elixir document. They would fail if one of the clauses in a case statement had an unused variable, for example. This was due to the old code processing one line at a time.

The new method uses Sourceror Zippers, which can focus on a line of code and process it. Even better, we can use Sourceror patches to generate a list of patches, which can then be turned directly into edits with no string processing at all. This new way of dealing with code is a lot easier than it was previously, and i'm much happier with the API.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control.ex
+++ b/apps/remote_control/lib/lexical/remote_control.ex
@@ -9,7 +9,7 @@ defmodule Lexical.RemoteControl do
   alias Lexical.RemoteControl.ProjectNode
   require Logger
 
-  @allowed_apps ~w(common path_glob remote_control elixir_sense)a
+  @allowed_apps ~w(common path_glob remote_control elixir_sense sourceror)a
 
   @app_globs Enum.map(@allowed_apps, fn app_name -> "/**/#{app_name}*/ebin" end)
 
--- a/apps/remote_control/lib/lexical/remote_control/code_mod/ast.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_mod/ast.ex
@@ -1,5 +1,9 @@
 defmodule Lexical.RemoteControl.CodeMod.Ast do
   alias Lexical.Document
+  alias Lexical.Document.Edit
+  alias Sourceror.Zipper
+
+  require Sourceror
 
   @type t :: any()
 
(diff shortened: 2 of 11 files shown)


apps/remote_control/lib/lexical/remote_control/code_mod/replace_with_underscore.ex
```diff
@@ -1,27 +1,19 @@
...
   @spec edits(Document.t(), non_neg_integer(), String.t() | atom) ::
           {:ok, Changes.t()} | :error
   def edits(%Document{} = document, line_number, variable_name) do
     variable_name = ensure_atom(variable_name)
 
-    with {:ok, line_text} <- Document.fetch_text_at(document, line_number),
-         {:ok, line_ast} <- Ast.from(line_text),
-         {:ok, transformed_text} <- apply_transform(line_text, line_ast, variable_name) do
-      edits = to_edits(line_number, line_text, transformed_text)
+    case apply_transform(document, line_number, variable_name) do
```
scottming, line 11: It looks great here, much simpler.

apps/remote_control/lib/lexical/remote_control/code_mod/replace_with_underscore.ex
```diff
@@ -33,55 +25,21 @@ defmodule Lexical.RemoteControl.CodeMod.ReplaceWithUnderscore do
...
-  defp leading_indent(line_text) do
-    case Regex.scan(@indent_regex, line_text) do
-      [indent] -> indent
-      _ -> ""
-    end
-  end
+    result =
+      Ast.traverse_line(document, line_number, [], fn
+        {{^unused_variable_name, _meta, nil} = node, _} = zipper, patches ->
+          [patch] = Sourceror.Patch.rename_identifier(node, underscored_variable_name)
```
scottming, line 34: It looks great here, much simpler.

apps/remote_control/lib/lexical/remote_control.ex
```diff
@@ -9,7 +9,7 @@ defmodule Lexical.RemoteControl do
   alias Lexical.RemoteControl.ProjectNode
   require Logger
 
-  @allowed_apps ~w(common path_glob remote_control elixir_sense)a
+  @allowed_apps ~w(common path_glob remote_control elixir_sense sourceror)a
```
scottming, line 12: We need to consider one issue: if we support plug-ins in the future, there may be many projects that depend on `sourceror`.

scottming: We also need to solve dependency conflict issues.

apps/remote_control/lib/lexical/remote_control/code_mod/ast.ex
```diff
@@ -14,6 +18,171 @@ defmodule Lexical.RemoteControl.CodeMod.Ast do
...
+    case Sourceror.string_to_quoted(s, token_metadata: true, columns: true) do
+      {:ok, code, _comments} ->
+        {:ok, code}
+
+      error ->
+        error
+    end
+  end
+
+  def zipper_at(%Document{} = document, %Document.Position{} = position) do
```
scottming, line 30: Shouldn't this function be private?

apps/remote_control/lib/lexical/remote_control.ex
scohen: I'm not following, the other projects will need to figure out dependency issues with respect to sourceror, which shouldn't be too bad.

apps/remote_control/lib/lexical/remote_control/code_mod/ast.ex
scohen: No, I think this could be really useful for code mods.

apps/remote_control/lib/lexical/remote_control/code_mod/replace_with_underscore.ex
scohen: Yes, this was a big improvement in both code readability and ease of use.

PR #180, opened 2023-05-24
Remove patch from progress/state_test

scottming:


scohen (approved):
Please use commit messages to discuss the context around the change.

PR #184, opened 2023-05-25
New project structure / beginning of plugins

scohen:
Making plugins necessitated a new project structure, as we're going to need to publish a separate library for plugin authors so they can use them. We'll also need a bunch of the modules they'll use, so we need the least stressful way of depending on the libraries we publish.

Creating a separate repo would make our lives very difficult, so I added a projects directory at the top level, and the published libraries go in there. We utilize path dependencies for our internal libraries.

I also added the beginnings of a plugin architecture, but I wanted to get this change out before it gets too big. Plugins are quite complicated, and we're going to need to do a bit of refactoring when we add one. That refactoring, plus all of this moving would make the PR very difficult to review.

This PR also includes a new genserver dispatcher so that the compile tracer isn't slowed down by extraneous work.


apps/remote_control/lib/lexical/remote_control/mix.ex
```diff
@@ -47,4 +46,14 @@ defmodule Lexical.RemoteControl.Mix do
...
     end)
   end
+
+  if Mix.env() == :test do
+    defp unique_project_name(%Project{} = project) do
+      Project.atom_name(project)
+    end
+  else
+    defp unique_project_name(%Project{} = project) do
+      :"#{Project.name(project)}.#{System.unique_integer()}"
```
scottming, line 56: This is for solving the multiple editors opening the same project problem, right?

apps/server/lib/lexical/server/state.ex
```diff
@@ -101,6 +101,7 @@ defmodule Lexical.Server.State do
            &Document.apply_content_changes(&1, version, event.content_changes)
          ) do
       {:ok, updated_source} ->
+        Api.diagnose(state.configuration.project, updated_source)
```
scottming, line 104: Do we no longer need the throttler?

Or where should we place the throttler?

apps/remote_control/lib/lexical/remote_control/plugin.ex
```diff
@@ -0,0 +1,95 @@
...
+
+  require Logger
+  import Messages
+
+  def on_module_updated(plugin_module) do
+    if Keyword.has_key?(plugin_module.__info__(:attributes), :lexical_plugin) do
+      current_plugiuns = get()
+
+      unless exists?(current_plugiuns, plugin_module) do
+        plugin_module.init()
```
scottming, line 15: We must ensure that `init/0` is only called after the project has been compiled. Can the design here achieve this goal?

scottming:
In addition to the mentioned issues, we also need to consider the problem of Diagnostics clearing, which doesn't seem to be reflected in this PR.

projects/lexical_plugin/lib/lexical/plugin/diagnostic.ex
```diff
@@ -0,0 +1,28 @@
+defmodule Lexical.Plugin.Diagnostic do
+  alias Lexical.Plugint.Diagnostic
+
+  @callback diagnose(Document.t()) :: [Diagnostic.Result.t()]
```
scottming, line 4: Where is the `Document.t()` from? I didn't see you alias `Lexical.Document`

scohen:
This pr doesn’t build diagnostics yet, that’s going to be another large change

projects/lexical_plugin/lib/lexical/plugin/diagnostic.ex
```diff
@@ -0,0 +1,28 @@
+defmodule Lexical.Plugin.Diagnostic do
+  alias Lexical.Plugint.Diagnostic
```
scottming, line 2:
```suggestion
  alias Lexical.Plugin.Diagnostic
```

apps/remote_control/lib/lexical/remote_control/plugin.ex
```diff
@@ -0,0 +1,95 @@
...
+        add(current_plugiuns, plugin_module)
```
scottming, line 16:
```suggestion
      current_plugins = get()

      unless exists?(current_plugins, plugin_module) do
        plugin_module.init()
        add(current_plugins, plugin_module)
```

apps/remote_control/lib/lexical/remote_control/mix.ex
scottming: These changes will cause some problems. Project node can't automatically be closed.

scottming: Please try to address one goal with each PR as much as possible

apps/remote_control/lib/lexical/remote_control/compilation/dispatch.ex
```diff
@@ -0,0 +1,52 @@
...
+      |> RemoteControl.notify_listener()
+    end
+  end
+
+  defp progress_message(file) do
+    relative_path = Path.basename(file)
+    message = "compiling: " <> relative_path
+
+    label =
+      if String.starts_with?(relative_path, "deps") do
```
scottming, line 44: `relative_path` become a basename, it will never start with `deps`, so the logic is not correct.

apps/remote_control/lib/lexical/remote_control/compilation/dispatch.ex
```diff
@@ -0,0 +1,52 @@
...
+    if Path.extname(file) == ".ex" do
+      file
+      |> progress_message()
+      |> RemoteControl.notify_listener()
+    end
+  end
+
+  defp progress_message(file) do
+    relative_path = Path.basename(file)
+    message = "compiling: " <> relative_path
```
scottming, line 41: Two colons make the UI of VScode look ugly. Besides, I prefer `dirname/basename` here because it can provide more accurate information and is also shorter.

image

```elixir
    relative_path = Path.relative_to_cwd(file)
    basename = Path.basename(file)
    message = file |> Path.dirname() |> Path.basename() |> Path.join(basename)
```
image

apps/remote_control/lib/lexical/remote_control/mix.ex
scohen: Nope, I was getting a build error, and having a unique name fixed it, but it broke formatting.

scohen: I figured out the actual problem; it was a naming conflict with the `lexical` project in the `projects` directory and the main app. I renamed the one in the `projects` directory to `lexical_shared`

apps/server/lib/lexical/server/state.ex
scohen: We will likely need it, but I didn't want to reimplement. Let's see how it goes.

apps/remote_control/lib/lexical/remote_control/compilation/dispatch.ex
scohen: That's interesting; my UI is completely different; I don't see the `mix compile` part at all

These messages go by so fast that they're not really viewable by mere mortals. The whole path was way too long, and my UI barely has space for the filename.

scohen: The more I think about this, the more I think the message should just say "xxx files compiled"

scohen:
> In addition to the mentioned issues, we also need to consider the problem of Diagnostics clearing, which doesn't seem to be reflected in this PR.

This is a good point, let me share my strategy with you now. 

I want to normalize all diagnostic messages to utilize the plugin replies --even builds. Each plugin will then own a specific source (and we'll treat ours with the source of `lexical`). Then, clearing diagnostics will be a matter of:

1. For a project build, clear all diagnostics when the project is compiled.
2. For a file, clear diagnostics when the file is changed and let the plugins re-add them.

apps/remote_control/lib/lexical/remote_control/compilation/dispatch.ex
scottming: I'm ok with this: `deps/.../something/other.ex`

Regarding this prefix `building project compiling`, I think it's best not to add it. This can be easily configured on the client side, whether it's Emacs or neovim.

However, in my understanding, VSCode cannot be configured in this way. Adding such a prefix may cause interference to users since these characters are also quite long: `building project compiling`.

apps/remote_control/lib/lexical/remote_control/compilation/dispatch.ex
scottming: Right, Calculating percentages can bring a better UI, but it's difficult.

scohen: It’s not really percentages, just a count of compiled files.

apps/remote_control/lib/lexical/remote_control/compilation/dispatch.ex
scohen: > This can be easily configured on the client side, whether it's Emacs or neovim

how? I looked at the config and couldn't find anything in lsp mode that allows this configuration

scohen: It's also a little strange to point out the length when this pr dramatically reduces it. 

How about <project name> compiling <file>

scohen: Also, I think you're missing what I'm suggesting. The label would be "building" and the message would be "compiling file" this is fewer characters than what was there before, and if you omit the message as emacs does, it will still make sense

scottming:
Apply your last suggestions: https://github.com/lexical-lsp/lexical/pull/184#discussion_r1208611614

Then you are good.😊

scottming (approved):

PR #188, opened 2023-05-28
Fix the parameter issue of Remote Callable in pipeline.

### LINKED ISSUE #185: When completing with pipe connection, remove the first parameter.
image

it should be `Enum.map(fun)~`

Manually deleting the first parameter is really annoying, even in vim mode, I still need to press many keys.

scottming:
This PR also includes another fix, which is that functions or macros with arity 0 should not appear in the autocomplete of pipelines.

Fixes #185 

image
image


scottming:
but the label should be `Enum.map(enumerable, fun)` or `Enum.map(fun)`?

any thoughts? @scohen

apps/server/lib/lexical/server/code_intelligence/completion/env.ex
```diff
@@ -490,4 +493,8 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
         {:ok, tokens, ''}
     end
   end
+
+  defp maybe_trim_dot(character, line) do
+    if String.ends_with?(line, "."), do: character - 1, else: character
```
scohen, line 498: This feels very specific to one case. Was it the only case that was missed? If it was, where's the unit test that shows the prior behavior and the fixed behavior?

apps/server/test/lexical/server/code_intelligence/completion/translations/function_test.exs
```diff
@@ -12,6 +12,22 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.FunctionTest d
...
+      assert completion.insert_text == "loaded_applications()"
+    end
+
+    test "do not suggest arity 0 functions if in a pipeline", %{project: project} do
+      {:error, :not_found} =
+        project
+        |> complete("|> Application.loaded_app|")
+        |> fetch_completion(kind: :function)
+    end
+
```
scohen, line 30: something else is going on here; I don't consistently get the first parameters removed from pipes, yet the test on line 31 passes.

apps/server/test/lexical/server/code_intelligence/completion/translations/function_test.exs
```diff
@@ -12,6 +12,22 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.FunctionTest d
...
+      {:ok, completion} =
+        project
+        |> complete("Application.loaded_app|")
+        |> fetch_completion(kind: :function)
+
+      assert completion.insert_text == "loaded_applications()"
+    end
+
+    test "do not suggest arity 0 functions if in a pipeline", %{project: project} do
+      {:error, :not_found} =
```
scohen, line 25:
```suggestion
      assert {:error, :not_found} =
```

Otherwise, this unit tests has no assertions.

scohen:
> but the label should be Enum.map(enumerable, fun) or Enum.map(fun)? I prefer to Enum.map(fun)

The latter.

apps/server/lib/lexical/server/code_intelligence/completion/env.ex
```diff
@@ -343,7 +343,10 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
   defp surround_context(%__MODULE__{} = env) do
     with {:ok, line} <- Document.fetch_text_at(env.document, env.position.line),
          %{context: _} = context <-
-           Code.Fragment.surround_context(line, {1, env.zero_based_character}) do
+           Code.Fragment.surround_context(
+             line,
+             {1, maybe_trim_dot(env.zero_based_character, line)}
+           ) do
```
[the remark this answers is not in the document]
scottming: You're right, Yesterday I didn't realize the benefits of token stream, but now I feel that this abstraction is really convenient.

apps/server/lib/lexical/server/code_intelligence/completion/env.ex
scottming: Ok, I deleted this function

apps/server/test/lexical/server/code_intelligence/completion/translations/function_test.exs
scottming: I don't quite understand this question, do you mean this test: 

`    test "arity 1 omits arguments if in a pipeline", %{project: project} do`?

scohen:
...or even `|> Enum.map(fun)`

apps/server/test/lexical/server/code_intelligence/completion/translations/function_test.exs
scottming: after using the token stream, I think this issue you mentioned already fixed

apps/server/lib/lexical/server/code_intelligence/completion/env.ex
```diff
@@ -113,13 +113,24 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
...
   def in_context?(%__MODULE__{} = env, :pipe) do
-    with {:ok, line, context} <- surround_context(env),
-         {:ok, {:operator, '|>'}} <- previous_surround_context(line, context) do
-      true
-    else
-      _ ->
-        false
-    end
+    env
+    |> prefix_tokens(:all)
```
scohen, line 117:
```suggestion
    |> token_stream()
```

apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
```diff
@@ -4,13 +4,25 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Callable do
 
   @callables [Result.Function, Result.Macro]
 
+  def completion(%callable_module{argument_names: []} = callable, %Env{} = env)
+      when callable_module in @callables do
+    if not Env.in_context?(env, :pipe) do
+      do_completion(callable, env)
+    end
```
scohen, line 11:
```suggestion
    unless Env.in_context?(env, :pipe) do
      do_completion(callable, env)
    end
```

apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
```diff
@@ -82,4 +87,12 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Callable do
   defp name_and_arity(%_{name: name, arity: arity}) do
     "#{name}/#{arity}"
   end
+
+  defp reduce_args_when_pipe(%{argument_names: argument_names} = callable, %Env{} = env) do
```
scohen, line 91: I don't like modifying the original callable and sending it through. I think it's a lot clearer to have a function that operates on the argument names;  I don't think this is an improvement in clarity over the original.

apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
```diff
@@ -4,13 +4,25 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Callable do
 
   @callables [Result.Function, Result.Macro]
 
+  def completion(%callable_module{argument_names: []} = callable, %Env{} = env)
```
scohen, line 7: why are you looking at the argument names, when you have an `arity` key in callables? That's a lot clearer, imo:
```elixir
def completion(%callable_module{arity: 0}, %Env{} = env) do 
...
```

apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
scohen: the original had a bug though, if it was a zero arg function, it would fail.

apps/server/test/lexical/server/code_intelligence/completion/translations/function_test.exs
scottming: I created a function named `normalize_argument_names`, and "callable_snippet" and "label" both require context awareness, so passing "env" to them makes sense.

apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
scottming: I created a function named normalize_argument_names, and `callable_snippet/2` and `label/2` both require context awareness, so passing "env" to them makes sense.

apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
```diff
@@ -44,23 +55,29 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Callable do
...
 
-  defp callable_snippet(%_{} = callable, %Env{} = env) do
-    argument_names =
-      if Env.in_context?(env, :pipe) do
-        tl(callable.argument_names)
-      else
-        callable.argument_names
-      end
+  defp argument_names(%_{arity: 0} = callable, _env) do
+    callable.argument_names
```
scohen, line 66: I'd just put [] 🤣

scohen (approved):

apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
scohen: that's much licer.

Issue #182, opened 2023-05-25
Make the completion of struct more consistent


scottming:
This #181 PR has dramatically improved the struct completions, but we can do even better by making the struct completion more consistent. Currently, we can complete sub-module structs in some places, while in others, we cannot.

![image](https://github.com/lexical-lsp/lexical/assets/12830256/ffee7ba0-b858-4b3e-96c3-957fcd81eabb)

![image](https://github.com/lexical-lsp/lexical/assets/12830256/e2261773-1b8b-4337-ad87-eb21b53088a5)

scohen:
So two things: 
The first thing to focus on is to take control of emitting completions for structs away from elixir_sense, and instead build identical functionality inside of lexical using what we already have. That shouldn't be too hard in my estimation, the tools already exist, it's just putting them to use. The toughest thing will be fuzzy matching, but I think that's not so bad. What I'd do is build an index like this:

1. Split the modules on ".". Then group them by their prefixes. So, the layout will look like:
```
[] => [every struct module name]
["Lexical"] => [every struct module name that lexical defines]
["Lexical", "Document"] => [every struct module name under "Lexical.Document"
```

2. Then , when you're completing, use `Code.Fragment.cursor_context` there are, I believe two cases you can get back:
  a. `{:struct, {:dot, {:alias, 'Foo.Bar'}}}`. Take "Foo.Bar", split it on "." and return all the struct names with that prefix. This is an exact match, as the prefix is complete, and there's no suffix. 
 b. `{:struct, 'Foo.Bar.Ba'}` Split on ".", then search exact on the prefix "Foo.Bar" and do a fuzzy match on "Ba" vs. the module names using `String.jaro_distance`. Pick some value that gets a pretty good match, and return any module names that are over that value.


I also had a thought about the UX of struct completions this morning that I'd like to try. 
The first rule is we only complete structs that are direct descendants of the typed module, so if I type `%Lexical.` I'd get completions for `%Lexical.Document{}` and `%Lexical.Project{}`.  However, preceding the struct suggestions, we'd get additional matches that will allow us to explore further. In the case above, I'd see `%Lexical.Document... (6 more)` and `Lexical.Server... (2 more)`. Selecting those would complete `%Lexical.Document.|`and `%Lexical.Server.` at which point, I'd see more of the subtree. To make what I'm saying more clear, imagine the following table is the menu.

| Label | Insert text |
|--------|--------|
[...]

PR #190, opened 2023-05-29
Improve UX when completing struct

scottming:
Fixes the first part of #182 

This PR has two improvements:

The first one is related to 182 issues. Additionally, I don't think we need fuzzy search. For example, if the target is `%Document.Lines{}`, without fuzzy search, we can simply type `%Doc` and then press enter or tab for auto-completion. More options will appear automatically and we can filter by typing `l`. With fuzzy search, maybe we could complete it by typing %Docl, but you need to remember that this module exists in advance. I don't think this experience is good. The current way is already very convenient for me.

~~The second improvement adds structure documentation, including **fields**, **types** and **default values**. I really like this UI but I'm unsure if you guys like the code implementation too. I'd like to hear your opinions first before adding more tests.~~

https://github.com/lexical-lsp/lexical/assets/12830256/c3fb4355-a51f-4fdd-bee0-f5e55ed7b392


apps/remote_control/lib/lexical/remote_control/api.ex
```diff
@@ -47,6 +48,10 @@ defmodule Lexical.RemoteControl.Api do
     ])
   end
 
+  def struct_doc(%Project{} = project, struct_module) do
```
scottming, line 51: I think the API here could be designed to be more versatile, not just limited to struct doc.

apps/remote_control/lib/lexical/remote_control/completion/struct.ex
```diff
@@ -0,0 +1,80 @@
...
+
+      _ ->
+        "#{inspect(struct.__struct__(), pretty: true, width: 40)}"
+        |> trim_parent(full_name)
+    end
+  end
+
+  defp fetch_struct_type(struct) do
+    with {:ok, types} <- Code.Typespec.fetch_types(struct),
+         {:type, t} <- Enum.find(types, &match?({:type, {:t, _, _}}, &1)) do
```
scohen, line 25: this relies on the typespec being called `t`. You should look for a type that is being defined in terms of a map with a`__struct__` key.

apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
```diff
@@ -24,15 +24,28 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
 
     defines_struct? = Intelligence.defines_struct?(env.project, module.full_name)
 
+    immediate_descentent_structs =
+      immediate_descendent_struct_modules(env.project, module.full_name)
+
+    defines_struct_in_descentents? =
```
scohen, line 30: `immediate_struct_in_descendents` and `immediate_descendent_structs`

apps/remote_control/lib/lexical/remote_control/api.ex
```diff
@@ -47,6 +48,10 @@ defmodule Lexical.RemoteControl.Api do
...
+    RemoteControl.call(project, Struct, :doc, [struct_module])
```
scohen, line 52: this should go through `Completion`, and it should be called `documentation` not `doc`

apps/remote_control/lib/lexical/remote_control/completion/struct.ex
```diff
@@ -0,0 +1,80 @@
+defmodule Lexical.RemoteControl.Completion.Struct do
```
scohen, line 1: The documentation is complicating the rest of this PR. We don't presently add documentation to other things, and I think we should table this

scohen:
>  I don't think we need fuzzy search

I gave you an example of where fuzzy search comes in handy. Providing a different example where you don't think it's handy doesn't make your case. I think it's important, and we already have it for other types of completions.

I've looked at this, and it's pretty complicated for what you get. Let's not focus on documentation right now, as that's the lion's share of the complexity here.

apps/remote_control/lib/lexical/remote_control/completion/struct.ex
scottming: Ok, we can do the documentation part in another pr

apps/server/lib/lexical/server/code_intelligence/completion/translations/struct.ex
```diff
@@ -20,6 +19,19 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Struct do
...
+      kind: :module,
+      label: "#{module_name}...(#{more} more structs)",
+      detail: "#{full_name}."
+    ]
+
+    insert_text = "#{module_name}."
+    range = edit_range(env)
+
+    builder.text_edit_snippet(env, insert_text, range, builder_opts)
+  end
```
scohen, line 33: I really don't like this pattern, translations are about translating an elixir sense completion into LSP. What we want here is to collapse and aggregate module completions that are returned from remote_control. Do we need another place to do this work?

Issue #192, opened 2023-05-30
Remove the matching things when a function's arguments of completion starts with a function matching


scottming:
<img width="485" alt="image" src="https://github.com/lexical-lsp/lexical/assets/12830256/70119124-f104-4419-bb69-0ad4b843592e">

like the example above, `cursor_context(env)` is enough.

scottming:
Fixed by #193

PR #193, opened 2023-05-30
Pulled out the name of arguments in pattern match args

scohen:
I was still getting pattern matches when completing arguments (like `%Project{} = project`). This attempts to parse out the argument name, which is assumed to be a variable on either side of the match, and falls back to `arg_#{index}`.


scohen:
Fixes #192

scottming:
I tried this branch, LGTM

image

scottming (approved):

Issue #141, opened 2023-05-05
When the `.lexical` does not exist, it will be compiled twice.


scottming:
https://user-images.githubusercontent.com/12830256/236463707-567e61d9-add3-4850-b237-512caba39d33.mp4

You can quickly reproduce this by the work-done-progress branch. ~What's worse, the main branch should now be compiled **three** times, which is why sometimes it takes a long time for the lexical to get ready.~

~Furthermore, if .lexical does not exist and I have deleted the deps folder, we will encounter this error(only for lexical).~

```elixir
2023-05-05T21:01:45.007682+08:00 error: Error in process <0.9016.0> on node 'lexical@127.0.0.1' with exit value:, {noproc,[{erlang,link,[<0.142.0>],[{error_info,#{module => erl_erts_errors}}]},{'Elixir.Module.ParallelChecker','-spawn/3-fun-0-',5,[{file,"lib/module/parallel_checker.ex"},{line,56}]}]}
```

~For other projects, there is a situation where this progress cannot be completed.~

<img width="266" alt="image" src="https://user-images.githubusercontent.com/12830256/236466079-e237d360-b97e-420e-9b54-aecfc26fbb10.png">

scottming:
Part of it has been fixed by #135 .

But I doubt whether it needs to be compiled twice.

PR #194, opened 2023-05-30
Removed initial compile

scohen:
It appears that creating the directory is enough. No need to do the compile. I also just called `loadpaths` rather than building the paths by hand. Hopefully, this makes things easier.


scohen:
Fixes #141

scottming (approved):

PR #195, opened 2023-05-31
Boost callable completions that are not double underscore/default

### LINKED ISSUE #183: Lower the priority of some default functions in the module's sorting order
image

```elixir
module_info/0
__info__(atom)
```

viniciusmuller:
Lowers the priority of double underscore or default functions when completing functions and captures.

Behavior with this patch: 
image

Closes #183

### DIFF
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
@@ -71,7 +71,13 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Callable do
 
   defp sort_text(%_{name: name, arity: arity}) do
     normalized = String.replace(name, "__", "")
-    "#{normalized}/#{arity}"
+    fun = "#{normalized}/#{arity}"
+
+    unless String.starts_with?(name, "__") or name in ["module_info"] do
+      Env.boost(fun)
+    else
+      fun
+    end
   end
 
   defp label(%_{name: name, argument_names: argument_names}) do
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
@@ -73,10 +73,10 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Callable do
     normalized = String.replace(name, "__", "")
     fun = "#{normalized}/#{arity}"
 
-    unless String.starts_with?(name, "__") or name in ["module_info"] do
-      Env.boost(fun)
-    else
+    if String.starts_with?(name, "__") or name in ["module_info"] do
       fun
+    else
+      Env.boost(fun)
     end
   end
 
--- a/apps/server/test/lexical/server/code_intelligence/completion/translations/function_test.exs
+++ b/apps/server/test/lexical/server/code_intelligence/completion/translations/function_test.exs
@@ -117,7 +117,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.FunctionTest d
       [capture, args_capture] =
         project
         |> complete(source)
-        |> Enum.filter(&(&1.sort_text == "&is_map/1"))
+        |> Enum.filter(&(String.ends_with?(&1.sort_text, "is_map/1")))
 
       assert capture.detail == "(Capture)"
       assert capture.insert_text == "is_map/1"
--- a/apps/server/test/lexical/server/code_intelligence/completion/translations/function_test.exs
+++ b/apps/server/test/lexical/server/code_intelligence/completion/translations/function_test.exs
@@ -201,26 +201,32 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.FunctionTest d
       assert completion.sort_text == "info/1"
     end
 
-    test "dunder functions have lower completion priority", %{project: project} do
-      completions =
-        complete(project, "GenServer.|")
-        |> Enum.sort_by(& &1.sort_text)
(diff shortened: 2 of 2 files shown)


apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
```diff
@@ -71,7 +71,13 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Callable do
 
   defp sort_text(%_{name: name, arity: arity}) do
     normalized = String.replace(name, "__", "")
-    "#{normalized}/#{arity}"
+    fun = "#{normalized}/#{arity}"
+
+    unless String.starts_with?(name, "__") or name in ["module_info"] do
```
scohen, line 76: nit: This will fail credo, don't use unless with an else. Favor if.

viniciusmuller:
Hey @scohen, I've updated the code to use if instead of unless and also updated a test that was failing since now the completions will be boosted by default (maybe we also could boost to a lower level than 5)

Also, do you think it makes sense to add any other functions to the `["module_info"]` list? I can also try to write a test for this "unprioritized behavior"

apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
```diff
@@ -71,7 +71,13 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Callable do
...
   defp sort_text(%_{name: name, arity: arity}) do
     normalized = String.replace(name, "__", "")
-    "#{normalized}/#{arity}"
+    fun = "#{normalized}/#{arity}"
+
+    if String.starts_with?(name, "__") or name in ["module_info"] do
+      fun
+    else
+      Env.boost(fun)
+    end
```
scottming, line 80: The logic in this section seems convoluted, mainly because the `normalization` and `conditional statements` are far apart and not connected.

Therefore, I prefer some code like this:

```elixir
  defp sort_text(%_{name: name, arity: arity}) do
    if String.starts_with?(name, "__") or name in ["module_info"] do
      "#{normalize(name)}/arity"
    else
      boosted = name |> normalized() |> Enum.boost
      "#{boosted}/arity"
    end
  end

  defp normalize(name) do
    String.replace(name, "__", "")
  end
```

scohen: I think what was there is fine. If you follow the suggestion, you have two places where normalizations happen and the logic to build the sort text is duplicated

scottming: Ok, ignore my suggestion.

scohen:
@viniciusmuller please write some tests, and I can't think of any functions that I'd  de-boost right now other than the dunder ones.

apps/server/test/lexical/server/code_intelligence/completion/translations/function_test.exs
```diff
@@ -200,5 +200,27 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.FunctionTest d
...
+
+    test "dunder functions have lower completion priority", %{project: project} do
+      completions =
+        complete(project, "GenServer.|")
+        |> Enum.sort_by(& &1.sort_text)
+
+      total_completions = length(completions)
+      info_idx = Enum.find_index(completions, &(&1.sort_text == "info/1"))
+
+      assert info_idx > total_completions * 0.80
```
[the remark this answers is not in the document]
viniciusmuller: I did some work on this:

```ex
    test "dunder functions have lower completion priority", %{project: project} do
      completions =
        project
        |> complete("GenServer.|")
        |> Enum.sort_by(& &1.sort_text)
        |> Enum.with_index()

      is_dunder = fn {fun, _} -> String.starts_with?(fun.label, "__") end
      dunder_functions = Enum.filter(completions, is_dunder)
      non_dunder = Enum.reject(completions, is_dunder)

      for {_f, idx} <- dunder_functions do
        assert Enum.all?(non_dunder, &smaller_index(&1, idx))
      end
    end

    test "default functions have lower completion priority", %{project: project} do
      completions =
        project
        |> complete("GenServer.|")
        |> Enum.sort_by(& &1.sort_text)
        |> Enum.with_index()

      is_default = fn {fun, _} -> fun.sort_text in ["module_info/0"] end
      default_funs = Enum.filter(completions, is_default)
      other_funs = Enum.reject(completions, is_default)

      for {_f, idx} <- default_funs do
        assert Enum.all?(other_funs, &smaller_index(&1, idx))
      end
    end
  end

  defp smaller_index({_, i}, idx), do: i < idx
```

But I don't like it very much, using this logic, the dunder case fails because the `module_info` function is not a dunder function and gets after it, and I don't think it would be ideal to filter it out in the dunder test case.

Maybe a simpler solution would be: get all the dunder and default functions and assert they do not start with a booster character, what do you think @scohen?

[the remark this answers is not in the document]
scohen: > Maybe a simpler solution would be: get all the dunder and default functions and assert they do not start with a booster character, what do you think

it might even be easier than that. Make a function called `boosted?` that checks for a boost character. Then it's 
```elixir
completions = 
project
|> complete("GenServer.|")
|> Enum.sort_by(& &1.sort_text)

{dunder_completions, non_dunder_completions} = Enum.split_with(completions, &String.starts_with?(&1.label, "__"))

for completion <- dunder_completions do 
  refute boosted?(completion)
end

for completion <- non_dunder_functions do 
  assert boosted?(completion)
end
```

scohen (approved):

PR #196, opened 2023-05-31
Support Struct fields completion when in struct arguments context

### LINKED ISSUE #189: Only complete the field keys when our cursor is already in the struct.
image

For the `%Document.Lines`, it only has two fields: lines and strat.

image
 

Therefore, when we are inside the parentheses and manually trigger completion with `ctrl+space`, completing these two fields is sufficient.

Maybe, auto-completion of the fields can be triggered whether the cursor is `%struct{|}` or after the comma `%struct{field1: value, |}`.

scottming:
There are four commits, Please review them one by one.

### Show case

With this completion, it means that we no longer need to remember the specific names of fields, and only need to manually trigger the completion.

image

image

Fixes #189

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/api.ex
+++ b/apps/remote_control/lib/lexical/remote_control/api.ex
@@ -47,6 +47,13 @@ defmodule Lexical.RemoteControl.Api do
     ])
   end
 
+  def complete_struct_fields(%Project{} = project, %Document{} = document, %Position{} = position) do
+    RemoteControl.call(project, RemoteControl.Completion, :struct_fields, [
+      document,
+      position
+    ])
+  end
+
   def definition(%Project{} = project, %Document{} = document, %Position{} = position) do
     RemoteControl.call(project, CodeIntelligence.Definition, :definition, [
       document,
--- a/apps/remote_control/lib/lexical/remote_control/code_intelligence/ast.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_intelligence/ast.ex
@@ -1,0 +1,77 @@
+defmodule Lexical.RemoteControl.CodeIntelligence.Ast do
+  alias Lexical.Document
+  alias Lexical.Document.Position
+  alias Lexical.RemoteControl.CodeMod.Ast.Aliases
+
+  require Logger
+
+  @doc """
+  Expands the aliases in the given `document`, `postion` and `module_aliases`.
+
+  When we refer to a module, it's usually a short name,
+  so it's probably aliased or in a nested module,
+  so we need to find the real full name of the module at the cursor position.
+
+  For example, if we have:
+
+    ```elixir
+    defmodule Project do
+      defmodule Issue do
+        defstruct [:message]
+      end
+
+      def message(%Issue{|} = issue) do # cursor marked as `|`
+      end
+    end
+    ```
+
+  Then the the expanded module is `Project.Issue`.
+
+  Another example:
+
+    ```elixir
+    defmodule Project do
+      defmodule Issue do
+        defstruct [:message]
+      end
+    end
+
+    defmodule MyModule do
+      alias Project, as: MyProject
+
+      def message(%MyProject.Issue{|} = issue) do
+      end
+    end
+    ```
+
+  Then the the expanded module is still `Project.Issue`.
+
+  And sometimes we can't find the full name by the `Aliases.at/2` function,
+  then we just return the `Module.concat(module_aliases)` as it is.
+  """
+  @type short_alias :: atom()
+  @type module_aliases :: [short_alias]
+
+  @spec expand_aliases(
+          document :: Document.t(),
+          position :: Position.t(),
+          module_aliases :: module_aliases()
+        ) :: {:ok, module()} | :error
+  def expand_aliases(%Document{} = document, %Position{} = position, module_aliases)
+      when is_list(module_aliases) do
+    [first | rest] = module_aliases
+
+    with {:ok, aliases_mapping} <- Aliases.at(document, position),
+         {:ok, from} <- Map.fetch(aliases_mapping, first) do
+      {:ok, Module.concat([from | rest])}
+    else
+      _ ->
+        {:ok, Module.concat(module_aliases)}
+    end
+  end
+
+  def expand_aliases(_, _, nil) do
+    Logger.warning("Aliases are nil, can't expand them")
+    :error
+  end
+end
--- a/apps/remote_control/lib/lexical/remote_control/code_mod/ast/aliases.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_mod/ast/aliases.ex
@@ -272,16 +272,7 @@ defmodule Lexical.RemoteControl.CodeMod.Ast.Aliases do
   end
 
   def at(%Document{} = doc, %Position{} = position) do
-    %{line: line} = position
-    # https://github.com/elixir-lang/elixir/issues/12673#issuecomment-1592845875
-    position = Position.new(line + 1, 1)
-
-    document_fragment = Document.fragment(doc, position)
-
-    case Code.Fragment.container_cursor_to_quoted(document_fragment,
-           columns: true,
-           token_metadata: true
-         ) do
+    case safe_container_cursor_to_quoted(doc, position) do
       {:ok, quoted} ->
         reducer = Reducer.new()
 
@@ -297,5 +288,33 @@ defmodule Lexical.RemoteControl.CodeMod.Ast.Aliases do
 
   defp collect(elem, %Reducer{} = reducer) do
     {elem, Reducer.update(reducer, elem)}
+  end
+
+  defp safe_container_cursor_to_quoted(doc, position) do
+    # https://github.com/elixir-lang/elixir/issues/12673#issuecomment-1592845875
+    # Note: because of the above issue: Using `cursor_context` + `container_cursor_to_quoted`
+    # can't deal with some cases like: `alias Foo.Bar, as: AnotherBar`,
+    # so we need to add a new line to make sure we can get the parrent node of the cursor
+    %{line: line} = position
+    added_new_line_position = Position.new(line + 1, 1)
+    document_fragment = Document.fragment(doc, added_new_line_position)
+
+    case container_cursor_to_quoted(document_fragment) do
+      {:ok, quoted} ->
+        {:ok, quoted}
+
+      _error ->
+        # NOTE: Adding new line doesn't always work,
+        # so we need to try again without adding new line
+        document_fragment = Document.fragment(doc, position)
+        container_cursor_to_quoted(document_fragment)
+    end
+  end
+
+  defp container_cursor_to_quoted(document_fragment) do
+    Code.Fragment.container_cursor_to_quoted(document_fragment,
+      columns: true,
+      token_metadata: true
+    )
   end
 end
--- a/apps/remote_control/lib/lexical/remote_control/completion.ex
+++ b/apps/remote_control/lib/lexical/remote_control/completion.ex
@@ -1,5 +1,7 @@
 defmodule Lexical.RemoteControl.Completion do
+  alias Lexical.Document
   alias Lexical.Document.Position
+  alias Lexical.RemoteControl.CodeIntelligence.Ast
   alias Lexical.RemoteControl.Completion.Candidate
 
   def elixir_sense_expand(doc_string, %Position{} = position) do
@@ -19,5 +21,35 @@ defmodule Lexical.RemoteControl.Completion do
         candidate
       end
     end
+  end
+
+  def struct_fields(%Document{} = document, %Position{} = position) do
+    container_struct_module =
+      document
+      |> Lexical.Ast.cursor_path(position)
+      |> container_struct_module()
+
+    with {:ok, struct_module} <- Ast.expand_aliases(document, position, container_struct_module),
+         true <- function_exported?(struct_module, :__struct__, 0) do
+      struct_module
+      |> struct()
+      |> Map.from_struct()
+      |> Map.keys()
+      |> Enum.map(&Candidate.StructField.new(&1, struct_module))
+    else
+      _ -> []
+    end
+  end
+
+  defp container_struct_module(cursor_path) do
+    Enum.find_value(cursor_path, fn
+      # current module struct: `%__MODULE__{|}`
+      {:%, _, [{:__MODULE__, _, _} | _]} -> [:__MODULE__]
+      # struct leading by current module: `%__MODULE__.Struct{|}`
+      {:%, _, [{:__aliases__, _, [{:__MODULE__, _, _} | tail]} | _]} -> [:__MODULE__ | tail]
+      # Struct leading by alias or just a aliased Struct: `%Struct{|}`, `%Project.Struct{|}`
+      {:%, _, [{:__aliases__, _, aliases} | _]} -> aliases
+      _ -> nil
+    end)
   end
 end
--- a/apps/remote_control/test/lexical/remote_control/completion_test.exs
+++ b/apps/remote_control/test/lexical/remote_control/completion_test.exs
@@ -1,0 +1,111 @@
+defmodule Lexical.RemoteControl.CompletionTest do
+  alias Lexical.Document
+  alias Lexical.Document.Position
+  alias Lexical.RemoteControl.Completion
+
+  import Lexical.Test.CursorSupport
+  import Lexical.Test.CodeSigil
+
+  use ExUnit.Case, async: true
+
+  describe "struct_fields/2" do
+    test "returns the field completion for current module" do
+      source = ~q<
+        defmodule Project.Issue do
+          defstruct [:message]
+          @type t :: %__MODULE__{|}
+        end
+      >
+
+      [message] = struct_fields(source)
+      assert message.name == :message
+    end
+
+    test "returns the field completions for aliased module" do
+      source = ~q<
+        defmodule Project.Issue do
+          alias Project.Issue
+          defstruct [:message]
+          @type t :: %Issue{|}
+        end
+      >
+
+      [message] = struct_fields(source)
+      assert message.name == :message
+    end
+
+    test "returns [] when cursor is not in a struct" do
+      source = ~q<
+        defmodule Project.Issue do
+          alias Project.Issue
+          defstruct [:message]
+          @type t :: %Issue{}|
+        end
+      >
+
+      assert struct_fields(source) == []
+    end
+
+    test "returns the field completions when cursor is in the current module child's arguments" do
+      source = ~q<
+        defmodule Project do
+          defmodule Issue do
+            @type t :: %Issue{}
+            defstruct [:message]
+          end
+
+          def message(%__MODULE__.Issue{|} = issue) do
+            issue.message
+          end
+        end
+      >
+
+      [message] = struct_fields(source)
+      assert message.name == :message
+    end
+
+    test "returns the field completion when cursor is in an alias child's arguments" do
+      source = ~q<
+        defmodule Project do
+          defmodule Issue do
+            defstruct [:message]
+          end
+        end
+
+        defmodule MyModule do
+          alias Project
+
+          def message(%Project.Issue{|} = issue) do
+            issue.message
+          end
+        end
+      >
+
+      [message] = struct_fields(source)
+      assert message.name == :message
+    end
+  end
+
+  defp document(source) do
+    text = strip_cursor(source)
+    Document.new(file_uri(), text, 1)
+  end
+
+  defp file_uri do
+    "file:///elixir.ex"
+  end
+
+  defp position(source) do
+    {line, column} = cursor_position(source)
+    Position.new(line, column)
+  end
+
+  defp struct_fields(source) do
+    document = document(source)
+    position = position(source)
+    text = Document.to_string(document)
+    Code.compile_string(text)
+
+    Completion.struct_fields(document, position)
+  end
+end
--- a/apps/server/lib/lexical/server/code_intelligence/completion.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion.ex
@@ -62,12 +62,28 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
       Enum.empty?(prefix_tokens) or not context_will_give_meaningful_completions?(env) ->
         Completion.List.new(items: [], is_incomplete: true)
 
+      Env.in_context?(env, :struct_arguments) and not Env.in_context?(env, :struct_field_value) and
+          not prefix_is_trigger?(env) ->
+        project
+        |> RemoteControl.Api.complete_struct_fields(env.document, env.position)
+        |> Enum.map(&Translatable.translate(&1, Env, env))
+
       true ->
         {document, position} = Env.strip_struct_reference(env)
 
         project
         |> RemoteControl.Api.complete(document, position)
         |> to_completion_items(project, env, context)
+    end
+  end
+
+  defp prefix_is_trigger?(env) do
+    case Env.prefix_tokens(env, 1) do
+      [{_, token, _}] ->
+        to_string(token) in trigger_characters()
+
+      _ ->
+        false
     end
   end
 
--- a/apps/server/lib/lexical/server/code_intelligence/completion.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion.ex
@@ -52,7 +52,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
       prefix_tokens == [] ->
         empty_completion_list()
 
-      match?([{:operator, :do}], prefix_tokens) and Env.empty?(env.suffix) ->
+      match?([{:operator, :do, _}], prefix_tokens) and Env.empty?(env.suffix) ->
         do_end_snippet = "do\n$0\nend"
 
         env
--- a/apps/server/lib/lexical/server/code_intelligence/completion/env.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/env.ex
@@ -77,16 +77,16 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
     env
     |> prefix_token_stream()
     |> Enum.reduce_while(false, fn
-      {:paren, :")"}, _ ->
+      {:paren, :")", _}, _ ->
         {:halt, false}
 
-      {:operator, :&}, _ ->
+      {:operator, :&, _}, _ ->
         {:halt, true}
 
-      {:int, _} = maybe_arity, _ ->
+      {:int, _, _} = maybe_arity, _ ->
         {:cont, maybe_arity}
 
-      {:operator, :/}, {:int, _} ->
+      {:operator, :/, _}, {:int, _, _} ->
         # if we encounter a trailing /<arity> in the prefix, the
         # function capture is complete, and we're not inside it
         {:halt, false}
@@ -147,16 +147,16 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
     env
     |> prefix_token_stream()
     |> Enum.reduce_while(false, fn
-      {:identifier, _}, _ ->
+      {:identifier, _, _}, _ ->
         {:cont, false}
 
-      {:operator, :.}, _ ->
+      {:operator, :., _}, _ ->
         {:cont, false}
 
-      {:alias, _}, _ ->
+      {:alias, _, _}, _ ->
         {:cont, false}
 
-      {:arrow_op, nil}, _ ->
+      {:arrow_op, nil, _}, _ ->
         {:halt, true}
 
       _x, _acc ->
@@ -171,8 +171,8 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
     |> Enum.reduce_while(
       false,
       fn
-        {:operator, :">>"}, _ -> {:halt, false}
-        {:operator, :"<<"}, _ -> {:halt, true}
+        {:operator, :">>", _}, _ -> {:halt, false}
+        {:operator, :"<<", _}, _ -> {:halt, true}
         _, _ -> {:cont, false}
       end
     )
@@ -197,27 +197,27 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
     |> prefix_token_stream()
     |> Stream.with_index()
     |> Enum.reduce_while(false, fn
-      {{:curly, :"{"}, _index}, :eol ->
+      {{:curly, :"{", _}, _index}, :eol ->
         {:cont, false}
 
-      {{:comma, _}, _index}, :eol ->
+      {{:comma, _, _}, _index}, :eol ->
         {:cont, false}
 
-      {{:eol, _}, _index}, _acc ->
+      {{:eol, _, _}, _index}, _acc ->
         {:cont, :eol}
 
-      {_, _}, :eol ->
+      {{_, _, _}, _}, :eol ->
         {:halt, false}
 
-      {{:curly, :"}"}, _index}, _ ->
+      {{:curly, :"}", _}, _index}, _ ->
         {:halt, false}
 
-      {{:identifier, 'alias'}, 0}, _ ->
+      {{:identifier, ~c"alias", _}, 0}, _ ->
         # there is nothing after the alias directive, so we're not
         # inside the context *yet*
         {:halt, false}
 
-      {{:identifier, 'alias'}, _index}, _ ->
+      {{:identifier, ~c"alias", _}, _index}, _ ->
         {:halt, true}
 
       _, _ ->
@@ -227,17 +227,17 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
 
   @impl Environment
   def in_context?(%__MODULE__{} = env, :import) do
-    in_directive?(env, 'import')
+    in_directive?(env, ~c"import")
   end
 
   @impl Environment
   def in_context?(%__MODULE__{} = env, :use) do
-    in_directive?(env, 'use')
+    in_directive?(env, ~c"use")
   end
 
   @impl Environment
   def in_context?(%__MODULE__{} = env, :require) do
-    in_directive?(env, 'require')
+    in_directive?(env, ~c"require")
   end
 
   @impl Environment
@@ -368,10 +368,10 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
     env
     |> prefix_token_stream()
     |> Enum.reduce_while(false, fn
-      {:identifier, ^context_name}, _ ->
+      {:identifier, ^context_name, _}, _ ->
         {:halt, true}
 
-      {:eol, _}, _ ->
+      {:eol, _, _}, _ ->
         {:halt, false}
 
       _, _ ->
@@ -390,47 +390,52 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
   defp normalize_token(token) do
     case token do
       :eol ->
-        {:eol, '\n'}
+        {:eol, ~c"\n", []}
 
-      {:bin_string, _, [string_value]} ->
-        {:string, string_value}
+      {:bin_string, context, [string_value]} ->
+        {:string, string_value, to_position(context)}
 
-      {:bin_string, _, interpolated} ->
-        {:interpolated_string, interpolated}
+      {:bin_string, context, interpolated} ->
+        {:interpolated_string, interpolated, to_position(context)}
 
-      {:capture_op, _context, value} ->
-        {:operator, value}
+      {:capture_op, context, value} ->
+        {:operator, value, to_position(context)}
 
-      {:dual_op, _context, value} ->
-        {:operator, value}
+      {:dual_op, context, value} ->
+        {:operator, value, to_position(context)}
 
-      {:type_op, _context, _value} ->
-        {:operator, :"::"}
+      {:type_op, context, _value} ->
+        {:operator, :"::", to_position(context)}
 
-      {:mult_op, _, operator} ->
-        {:operator, operator}
+      {:mult_op, context, operator} ->
+        {:operator, operator, to_position(context)}
 
-      {:in_op, _, _} ->
-        {:operator, :in}
+      {:in_op, context, _} ->
+        {:operator, :in, to_position(context)}
 
-      {:operator, _, value} ->
-        {:operator, value}
+      {:operator, context, value} ->
+        {:operator, value, to_position(context)}
 
-      {:sigil, _, sigil_char, _sigil_context, _, _opts, delim} ->
-        {:sigil, [sigil_char], delim}
+      {:sigil, {line, column, _}, sigil_char, _sigil_context, _, _opts, delim} ->
+        # NOTE: should we need to return context too?
+        {:sigil, [sigil_char], {line, column}, delim}
 
-      {type, {_, _, nil}, value} when is_list(value) ->
-        {normalize_type(type), value}
+      {type, {line, column, nil}, value} when is_list(value) ->
+        {normalize_type(type), value, {line, column}}
 
-      {type, {_, _, token_value}, _} ->
-        {normalize_type(type), token_value}
+      {type, {line, column, token_value}, _} ->
+        {normalize_type(type), token_value, {line, column}}
 
-      {type, _context, value} when is_atom(value) ->
-        {normalize_type(type), value}
+      {type, context, value} when is_atom(value) ->
+        {normalize_type(type), value, to_position(context)}
 
-      {operator, _} ->
-        {map_operator(operator), operator}
+      {operator, context} ->
+        {map_operator(operator), operator, to_position(context)}
     end
+  end
+
+  defp to_position({line, column, _}) do
+    {line, column}
   end
 
   defp map_operator(:"("), do: :paren
@@ -448,7 +453,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
 
   defp prefix_token_stream(%__MODULE__{} = env) do
     init_function = fn ->
-      {env, '', env.position.line}
+      {env, ~c"", env.position.line}
     end
 
     next_function = fn
@@ -501,13 +506,13 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
 
     case :future_elixir_tokenizer.tokenize(current_context, line_number, 1, []) do
       {:ok, _, _, _, tokens} ->
-        {:ok, Enum.reverse(tokens), ''}
+        {:ok, Enum.reverse(tokens), ~c""}
 
-      {:error, {_, _, 'unexpected token: ', _}, _, _, _} ->
-        {:ok, [], '\n' ++ current_context}
+      {:error, {_, _, ~c"unexpected token: ", _}, _, _, _} ->
+        {:ok, [], ~c"\n" ++ current_context}
 
       {:error, _, _, _, tokens} ->
-        {:ok, tokens, ''}
+        {:ok, tokens, ~c""}
     end
   end
 end
--- a/apps/server/lib/lexical/server/code_intelligence/completion/env.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/env.ex
@@ -1,5 +1,6 @@
 defmodule Lexical.Server.CodeIntelligence.Completion.Env do
   alias Future.Code, as: Code
+  alias Lexical.Ast
   alias Lexical.Completion.Builder
   alias Lexical.Completion.Environment
   alias Lexical.Document
@@ -109,6 +110,35 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
 
       _ ->
         false
+    end
+  end
+
+  @impl Environment
+  def in_context?(%__MODULE__{} = env, :struct_arguments) do
+    cursor_path = Ast.cursor_path(env.document, env.position)
+
+    Enum.any?(cursor_path, fn
+      # struct leading by current module: `%__MODULE__.Struct{|}`
+      # or leading by a module alias: `%Alias.Struct{|}`
+      # or just a struct: `%Struct{|}`
+      {:%, _, [{:__aliases__, _, _aliases} | _]} -> true
+      # current module struct: `%__MODULE__{|}`
+      {:%, _, [{:__MODULE__, _, _} | _]} -> true
+      _ -> false
+    end)
+  end
+
+  @impl Environment
+  def in_context?(%__MODULE__{} = env, :struct_field_value) do
+    if in_context?(env, :struct_arguments) do
+      env
+      |> prefix_tokens(2)
+      |> Enum.any?(fn
+        {:kw_identifier, _, _} -> true
+        _ -> false
+      end)
+    else
+      false
     end
   end
 
--- a/apps/server/lib/lexical/server/code_intelligence/completion/env.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/env.ex
@@ -103,10 +103,19 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
       {:ok, _line, {:struct, _}} ->
         true
 
-      {:ok, line, {:local_or_var, [?_, ?_ | rest]}} ->
+      {:ok, _line, {:local_or_var, [?_ | _rest]}} ->
         # a reference to `%__MODULE`, often in a function head, as in
         # def foo(%__)
-        String.starts_with?("MODULE", List.to_string(rest)) and String.contains?(line, "%__")
+
+        starts_with_percent? =
+          env
+          |> prefix_tokens(2)
+          |> Enum.any?(fn
+            {:percent, :%, _} -> true
+            _ -> false
+          end)
+
+        starts_with_percent? and (ancestor_is_def?(env) or ancestor_is_type?(env))
 
       _ ->
         false
@@ -514,5 +523,29 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
       {:error, _, _, _, tokens} ->
         {:ok, tokens, ~c""}
     end
+  end
+
+  defp ancestor_is_def?(env) do
+    env.document
+    |> Ast.cursor_path(env.position)
+    |> Enum.any?(fn
+      {:def, _, _} ->
+        true
+
+      {:defp, _, _} ->
+        true
+
+      _ ->
+        false
+    end)
+  end
+
+  defp ancestor_is_type?(env) do
+    env.document
+    |> Ast.cursor_path(env.position)
+    |> Enum.any?(fn
+      {:type, _, _} -> true
+      _ -> false
+    end)
   end
 end
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/struct_field.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/struct_field.ex
@@ -1,4 +1,5 @@
 defmodule Lexical.Server.CodeIntelligence.Completion.Translations.StructField do
+  alias Future.Code, as: Code
   alias Lexical.RemoteControl.Completion.Candidate
   alias Lexical.Server.CodeIntelligence.Completion.Env
   alias Lexical.Server.CodeIntelligence.Completion.Translatable
@@ -9,11 +10,39 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.StructField do
     :skip
   end
 
+  def translate(%Candidate.StructField{call?: false} = struct_field, builder, %Env{} = env) do
+    name = struct_field.name
+    value = to_string(name)
+
+    builder_opts = [
+      kind: :field,
+      label: "#{name}: #{value}"
+    ]
+
+    insert_text = "#{name}: ${1:#{value}}"
+    range = edit_range(env)
+
+    builder.text_edit_snippet(env, insert_text, range, builder_opts)
+  end
+
   def translate(%Candidate.StructField{} = struct_field, builder, %Env{} = env) do
     builder.plain_text(env, struct_field.name,
       detail: struct_field.name,
       label: struct_field.name,
       kind: :field
     )
+  end
+
+  def edit_range(env) do
+    prefix_end = env.position.character
+
+    case Code.Fragment.cursor_context(env.prefix) do
+      {:local_or_var, field_char} ->
+        edit_begin = env.position.character - length(field_char)
+        {edit_begin, prefix_end}
+
+      _ ->
+        {prefix_end, prefix_end}
+    end
   end
 end
--- a/apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
+++ b/apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
@@ -228,6 +228,131 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
     end
   end
 
+  describe "in_context?(env, :struct_arguments)" do
+    def wrap_with_module(text) do
+      """
+      defmodule MyModule do
+        #{text}
+      end
+      """
+    end
+
+    def wrap_with_function(text) do
+      """
+      def func do
+        #{text}
+      end
+      """
+    end
+
+    def wrap_with_function_arguments(text) do
+      """
+      def func(#{text}) do
+      end
+      """
+    end
+
+    test "is true if the cursor is directly after the opening curly" do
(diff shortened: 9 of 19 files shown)


apps/server/lib/lexical/server/code_intelligence/completion/env.ex
```diff
@@ -475,6 +570,28 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
...
+    prefix_tokens = Enum.reverse(tokens)
+
+    column =
+      Enum.find_value(prefix_tokens, fn
+        {:identifier, {_, column, _}, :__MODULE__} -> column
+        {:alias, {_, column, _}, _} -> column
+        _ -> nil
+      end)
+
+    # else
```
scottming, line 586: Should we need to make this function more dynamic? not only search the alias token column?

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -227,12 +227,49 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
     end
   end
 
+  describe "in_context?(env, :struct_arguments)" do
+    test "is true if the cursor after curly" do
+      env = new_env("%User{|}")
+      assert in_context?(env, :struct_arguments)
+    end
+
```
scohen, line 235:
```elixir
test "is true even if the value of a struct key is a tuple" do 
  env = new_env("%User{favorite_numbers: {3}|")
  assert in_context?(env, :struct_arguments)
end
```

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast.ex
```diff
@@ -0,0 +1,101 @@
+defmodule Lexical.RemoteControl.CodeIntelligence.Ast do
+  alias Lexical.Document
+  alias Lexical.Document.Position
+
+  def original_module(aliases_mapping, module) do
```
scottming, line 5: I'm not sure this function should be here or not.

if you think this is the right place, I can add some tests like:

```elixir
  describe "original_module/2" do
    test "it should return the original module" do
      aliases_mapping = %{
        User => Project.Structs.User,
        :__MODULE__ => MyModule
      }

      assert Ast.original_module(aliases_mapping, User) == Project.Structs.User
    end

    test "it should return the original when the module is a nested" do
      aliases_mapping = %{
        User => Project.Structs.User,
        :__MODULE__ => MyModule
      }

      assert Ast.original_module(aliases_mapping, User.Profile) == Project.Structs.User.Profile
    end

    test "it should return the current module" do
      aliases_mapping = %{
        User => Project.Structs.User,
        :__MODULE__ => MyModule
      }

      assert Ast.original_module(aliases_mapping, :__MODULE__) == MyModule
    end

    test "it should return the module when the module is not in aliases mapping" do
      aliases_mapping = %{
        User => Project.Structs.User,
        :__MODULE__ => MyModule
      }

      assert Ast.original_module(aliases_mapping, Project.XX) == Project.XX
    end
  end

```

apps/server/lib/lexical/server/code_intelligence/completion.ex
```diff
@@ -60,6 +61,10 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
       Enum.empty?(prefix_tokens) or not context_will_give_meaningful_completions?(env) ->
         Completion.List.new(items: [], is_incomplete: true)
 
+      Env.in_context?(env, :struct_arguments) ->
+        struct_module = Env.nearest_struct(env)
```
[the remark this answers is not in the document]
scottming: Yeah, I think prefix_alias would be better.

apps/server/lib/lexical/server/code_intelligence/completion.ex
```diff
@@ -60,6 +61,10 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
...
+        StructField.complete(env, Env, struct_module)
```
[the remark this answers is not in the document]
scottming: can I use it just to translate a list of fields?

apps/server/lib/lexical/server/code_intelligence/completion/env.ex
```diff
@@ -111,6 +111,26 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
     end
   end
 
+  @impl Environment
+  def in_context?(%__MODULE__{} = env, :struct_arguments) do
+    env
```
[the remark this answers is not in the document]
scottming: I didn't find an easy way to do this:

I think if encounter `}`, I need to convert the prefix to quoted and get the path to the `}` to see is a parent is `%{}` or not.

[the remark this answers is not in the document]
scottming:
```elixir
  {:__cursor__, [line: 4], []},
  {:favorite_numbers, {:__cursor__, [line: 4], []}},
  {:%{}, [line: 4], [favorite_numbers: {:__cursor__, [line: 4], []}]},
  {:%, [line: 4],
   [
     {:__aliases__, [line: 4], [:User]},
     {:%{}, [line: 4], [favorite_numbers: {:__cursor__, [line: 4], []}]}
   ]},

```

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast.ex
```diff
@@ -0,0 +1,101 @@
+defmodule Lexical.RemoteControl.CodeIntelligence.Ast do
```
scottming, line 1: I think in the near future, `find references` and `go to definition` will use the functions in this module too, so I put them here.

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast.ex
```diff
@@ -0,0 +1,197 @@
...
+        quoted
+        |> alias_mapping()
+        |> Map.merge(current_module_mapping(quoted))
+
+      _ ->
+        %{}
+    end
+  end
+
+  defp alias_mapping(quoted) do
```
scohen, line 43: does this function need to be written in terms of a reduce over the current document? 
I'm imagining you collect aliases in a context as you encounter them, and then store ranges where the context is valid. I think that'd improve the readability of a lot of this, and would prevent you from needing all these helpers that need to go back earlier in the document and rebuild context. 

If nothing else, you could still use this method to build up aliases until a certain point, you just stop at that position.

apps/server/lib/lexical/server/code_intelligence/completion.ex
```diff
@@ -69,6 +79,18 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
     end
   end
 
+  defp prefix_not_a_trigger?(env) do
```
scohen, line 82: `prefix_is_trigger?` and reverse the logic in the function. Then `not prefix_is_trigger?` in the cond.

Always name functions and variables after the positive case.

apps/server/lib/lexical/server/code_intelligence/completion.ex
```diff
@@ -69,6 +79,18 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
...
+    [last_prefix] = Env.prefix_tokens(env, 1)
+
+    case last_prefix do
+      {_, token, _} ->
+        to_string(token) not in trigger_characters()
+
+      _ ->
+        true
+    end
+  end
```
scohen, line 92:
```suggestion
    case Env.prefix_tokens(env, 1) do
      [{_, token, _}] ->
        to_string(token) not in trigger_characters()

      _ ->
        true
    end
  end
```

apps/server/lib/lexical/server/code_intelligence/completion/env.ex
```diff
@@ -475,4 +512,16 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
         {:ok, tokens, ''}
     end
   end
+
+  defp path_to_cursor(env) do
+    fragment = Document.fragment(env.document, env.position)
+
+    case Code.Fragment.container_cursor_to_quoted(fragment, columns: true) do
+      {:ok, quoted} ->
+        quoted |> Macro.path(&match?({:__cursor__, _, _}, &1)) |> List.wrap()
```
scohen, line 521:
```suggestion
        quoted 
          |> Macro.path(&match?({:__cursor__, _, _}, &1)) 
          |> List.wrap()
```

apps/server/lib/lexical/server/code_intelligence/completion/env.ex
```diff
@@ -475,4 +512,16 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
         {:ok, tokens, ''}
     end
   end
+
+  defp path_to_cursor(env) do
```
scohen, line 516: this function is called `path_to_cursor` but it takes an `env` and returns a path`

apps/remote_control/lib/lexical/remote_control/api.ex
```diff
@@ -47,6 +47,16 @@ defmodule Lexical.RemoteControl.Api do
     ])
   end
 
+  @types ~w(struct_fields)a
+
+  def complete(%Project{} = project, %Document{} = document, %Position{} = position, type)
```
scohen, line 52: why not just call this `struct_fields` or `complete_struct_fields` ?

apps/remote_control/test/lexical/remote_control/code_intelligence/ast_test.exs
```diff
@@ -0,0 +1,166 @@
+defmodule Lexical.RemoteControl.CodeIntelligence.AstTest do
```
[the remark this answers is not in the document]
scottming: It won't work until the 1.15 Elixir version. https://github.com/elixir-lang/elixir/issues/12566#issuecomment-1555261924

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast.ex
```diff
@@ -0,0 +1,197 @@
...
+  end
+
+  def aliases_at_position(%Document{} = document, %Position{} = position) do
+    document
+    |> Document.fragment(position)
+    |> collect_aliases()
+  end
+
+  defp collect_aliases(fragment) do
+    case Code.Fragment.container_cursor_to_quoted(fragment) do
```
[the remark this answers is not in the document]
scottming: what is container_context_to_quoted?

apps/remote_control/lib/lexical/remote_control/completion.ex
```diff
@@ -18,4 +20,50 @@ defmodule Lexical.RemoteControl.Completion do
...
+  def struct_fields(%Document{} = document, %Position{} = position) do
+    context =
+      Code.Fragment.surround_context(
+        Document.to_string(document),
+        {position.line, position.character}
+      )
+
+    case matched_alias(context) do
+      {:ok, struct_alias, struct_alias_end_column} ->
+        # NOTE: `@type t :: %|Struct{}` will make the `cursor_to_quoted` not return `aliases` ast
```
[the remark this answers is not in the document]
scottming: Is there a better way? 

In my understanding, `string_to_quoted` may have more issues as it can encounter syntax errors.

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast.ex
```diff
@@ -0,0 +1,197 @@
...
+      end)
+
+    if path do
+      [defs | _] = path
+      {:defmodule, _, [{:__aliases__, _, aliases} | _]} = defs
+      aliases
+    end
+  end
+
+  defp alias_source(quoted, original) do
```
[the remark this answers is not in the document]
scottming: This function is for something like `alias __MODULE__.Child` or `alias MyModule.Child`, so the source is `MyModule.Child`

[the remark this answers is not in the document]
scottming: maybe `original_module/2` is better?

apps/server/lib/lexical/server/code_intelligence/completion/env.ex
scottming: I think `cursor_path/1` would be better.

apps/remote_control/test/lexical/remote_control/code_intelligence/ast_test.exs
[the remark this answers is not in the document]
scohen: We can bring the changed code into our project if need be. I was more pointing out that there's no unit test that exercises this.

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast.ex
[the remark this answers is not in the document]
scohen: I'd call that a complete, full, or unalised module.

apps/server/lib/lexical/server/code_intelligence/completion/env.ex
```diff
@@ -111,6 +111,26 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
...
+  @impl Environment
+  def in_context?(%__MODULE__{} = env, :struct_arguments) do
+    env
+    |> prefix_tokens(:all)
+    |> Enum.chunk_every(2, 1, :discard)
+    |> Enum.reduce_while(false, fn
+      [{:curly, :"}"}, _], _ ->
+        {:halt, false}
+
+      [curly: :"{", alias: _], _ ->
```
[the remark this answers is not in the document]
scottming: We do not need this anymore after having Future.Code

apps/server/lib/lexical/server/code_intelligence/completion.ex
scottming: Good habits.

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast/aliases.ex
```diff
@@ -0,0 +1,208 @@
...
+    end
+  end
+
+  defp unexpanded_alias_mapping(parent_modules, children_ast) do
+    for {:__aliases__, _, [child]} <- children_ast, into: %{} do
+      {Module.concat([child]), Module.concat(parent_modules ++ [child])}
+    end
+  end
+
+  defp fetch_aliases_of_current_module(quoted, alias_line) do
```
scottming, line 177: This is the user-typed current module: `__MODULE__`, maybe I should use `cursor_current_module` in another place.

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast.ex
```diff
@@ -0,0 +1,197 @@
...
+    modules_length = length(modules)
+
+    modules
+    |> Enum.with_index(fn module, index ->
+      {module, modules |> Enum.slice(index, modules_length) |> Enum.reverse() |> Module.concat()}
+    end)
+    |> Enum.into(%{})
+  end
+
+  defp incomplete_modules(mapping) do
```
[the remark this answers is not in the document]
scottming: ok, changed to depentent_modules

apps/remote_control/lib/lexical/remote_control/completion.ex
```diff
@@ -18,4 +20,50 @@ defmodule Lexical.RemoteControl.Completion do
...
   end
+
+  def struct_fields(%Document{}, %Position{character: nil}) do
+    []
+  end
+
+  def struct_fields(%Document{} = document, %Position{} = position) do
+    context =
+      Code.Fragment.surround_context(
+        Document.to_string(document),
```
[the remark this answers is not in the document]
scottming: I don't understand this comment.

I was trying to fetch the struct module at the cursor.

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast.ex
```diff
@@ -0,0 +1,197 @@
...
+    end
+  end
+
+  defp alias_source(quoted, original) do
+    [first_alised_module | tail] = original
+
+    if match?({:__MODULE__, _, _}, first_alised_module) do
+      {:__MODULE__, position, _} = first_alised_module
+      source = source_aliases_of_current_module(quoted, position[:line])
+      source && Module.concat(source ++ tail)
```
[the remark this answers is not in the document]
scottming: change to `{:ok, aliases} | :error` pattern

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast/aliases.ex
```diff
@@ -0,0 +1,214 @@
...
+      end)
+
+    if path do
+      [defs | _] = path
+      {:defmodule, _, [{:__aliases__, _, aliases} | _]} = defs
+      {:ok, aliases}
+    else
+      Logger.info("cannot find the module definition of __MODULE__ in #{alias_line}")
+      :error
+    end
```
scohen, line 196: this might read easier as a case
```elixir

```suggestion
    case path do
      [{:defmodule, _, [{:__aliases__, _, aliases} | _]} | _] -> 
        {:ok, aliases}
    _other ->
      Logger.info("cannot find the module definition of __MODULE__ in #{alias_line}")
      :error
    end
```

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast/aliases.ex
```diff
@@ -0,0 +1,214 @@
...
+  end
+
+  defp unexpanded_alias_mapping(parent_modules, children_ast) do
+    for {:__aliases__, _, [child]} <- children_ast, into: %{} do
+      {Module.concat([child]), Module.concat(parent_modules ++ [child])}
+    end
+  end
+
+  defp fetch_aliases_of_current_module(quoted, alias_line) do
```
scohen, line 182: `fetch_current_module_aliases`

apps/remote_control/test/lexical/remote_control/code_intelligence/ast/aliases_test.exs
```diff
@@ -0,0 +1,149 @@
...
+
+      assert aliases[Project.MyModule] == Project.MyModule
+      assert aliases[Child] == Project.MyModule.Child
+      assert aliases[Other] == Project.MyModule.Other
+      assert aliases[:__MODULE__] == Project.MyModule
+    end
+  end
+
+  defp at_position(source) do
+    Ast.Aliases.at_position(document(source), position(source))
```
scohen, line 133: pipeline?
```suggestion
   source
   |> document()
   |>  Ast.Aliases.at_position(position(source))
```

apps/remote_control/test/lexical/remote_control/code_intelligence/ast/aliases_test.exs
```diff
@@ -0,0 +1,149 @@
...
+  alias Lexical.Document
+  alias Lexical.Document.Position
+  alias Lexical.RemoteControl.CodeIntelligence.Ast
+
+  import Lexical.Test.CursorSupport
+  import Lexical.Test.CodeSigil
+
+  use ExUnit.Case, async: true
+
+  describe "at_position/2" do
```
scohen, line 11: this test feels light to me given how much code is in aliases. Are you sure all code paths and all the things you've accounted for are tested here?
What about things that weren't included?

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast/aliases.ex
```diff
@@ -0,0 +1,214 @@
...
+    with true <- current_module_as_atom?(head),
+         {:__MODULE__, position, _} = head,
+         {:ok, aliases} <- fetch_aliases_of_current_module(quoted, position[:line]) do
+      Module.concat(aliases ++ tail)
+    else
+      _ -> Module.concat(original)
+    end
+  end
+
+  defp current_module_as_atom?(module) do
```
scohen, line 211: this function is misnamed. It is checking for a three tuple where the first element is `:__MODULE__`, yet uses the word `as` which implies some conversion is taking place. 
Suggest: `current_module_alias?` Though I'm not sure why you're matching like that at all, given line 203. 
why not (in the `full` function)
```elixir
with {:__MODULE__, metadata, _} <- head, 
     {:ok, line} <- Keyword.fetch(metadata, :line),
    {:ok, aliases} <- fetch_current_module_aliases(quoted, line) do 
  Module.concat(aliases ++ tail)
else _ ->
  Module.concat(original)
end
```

apps/remote_control/lib/lexical/remote_control/completion.ex
```diff
@@ -18,4 +20,42 @@ defmodule Lexical.RemoteControl.Completion do
...
+      )
+
+    case matched_alias(context) do
+      {:ok, struct_alias} ->
+        struct_module = Ast.expand(document, position, struct_alias)
+
+        fields =
+          if function_exported?(struct_module, :__struct__, 0) do
+            for {field_name, _v} <- Map.from_struct(struct_module.__struct__()), do: field_name
+          end
```
scohen, line 42:
```suggestion
          if function_exported?(struct_module, :__struct__, 0) do
            struct_module 
              |> struct()
              |> Map.from_struct()
              |> Map.keys()
          else 
            []
          end
```
Also remove the call to List.wrap()

apps/remote_control/lib/lexical/remote_control/completion.ex
```diff
@@ -18,4 +20,42 @@ defmodule Lexical.RemoteControl.Completion do
...
+
+        List.wrap(fields)
```
scohen, line 44: if it's not a struct, this will return `[nil]` that doesn't seem to be desired, right?

apps/remote_control/lib/lexical/remote_control/completion.ex
```diff
@@ -18,4 +20,42 @@ defmodule Lexical.RemoteControl.Completion do
...
+      _ ->
+        []
+    end
+  end
+
+  defp matched_alias(%{context: {:local_or_var, '__MODULE__'}}) do
+    {:ok, :__MODULE__}
+  end
+
+  defp matched_alias(%{context: {:struct, struct}}) do
```
scohen, line 55: can you rename `struct` to `struct_module_charlist` ?

apps/remote_control/test/lexical/remote_control/completion_test.exs
```diff
@@ -0,0 +1,47 @@
...
+  alias Lexical.Document.Position
+  alias Lexical.RemoteControl.Completion
+
+  import Lexical.Test.CursorSupport
+  import Lexical.Test.CodeSigil
+
+  use ExUnit.Case, async: true
+
+  describe "struct_fields/2" do
+    test "it should move the cursor into struct arguments and return the correct fields" do
```
scohen, line 12: I disagree with this; we should most definitely _not_ "move the cursor" anywhere; from where the cursor is positioned here, we should be completing struct modules, not fields.

apps/server/lib/lexical/server/code_intelligence/completion.ex
```diff
@@ -61,6 +62,15 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
...
         Completion.List.new(items: [], is_incomplete: true)
 
+      Env.in_context?(env, :struct_arguments) and not Env.in_context?(env, :value) and
+          not prefix_is_trigger?(env) ->
+        {line, column} = Env.prefix_alias_position(env)
+        poisiton = Position.new(line, column)
+
+        project
+        |> RemoteControl.Api.complete_struct_fields(env.document, poisiton)
+        |> StructField.translate(Env, env)
```
scohen, line 72: this is calling a protocol directly, which isn't how protocols are supposed to be used. I suspect it's necessary because you're calling a different function for struct completions rather than "normal" completions, which feels like it's incorrect to me. Why was that necessary? 

The feeling I had is that you should ask `remote_control` for the completions and that `remote_control` should deliver struct field completions back. You shouldn't have to do a check prior to calling it and call a different function.

scohen (changes requested):

apps/remote_control/lib/lexical/remote_control/completion.ex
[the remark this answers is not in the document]
scottming: we can't do that, `surround_context` need the whole document, otherwise, it will return a `none`

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast/aliases.ex
```diff
@@ -0,0 +1,214 @@
+defmodule Lexical.RemoteControl.CodeIntelligence.Ast.Aliases do
+  alias Future.Code, as: Code
+  alias Lexical.Document
+  alias Lexical.Document.Position
+  require Logger
+
+  def at_position(%Document{} = document, %Position{} = position) do
+    %{line: line} = position
+    # https://github.com/elixir-lang/elixir/issues/12673#issuecomment-1592845875
+    position = Position.new(line + 1, 1)
```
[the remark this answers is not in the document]
scottming: Jose mentioned this at the linked URL, for example, without adding a new line, we can't deal with this: https://github.com/elixir-lang/elixir/issues/12673#issuecomment-1592852227

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast/aliases.ex
scottming: we can't just use `{:__MODULE__, metadata, _} <- head` to do the match, that will trigger a matching exception when the head is like "HeadModule"

apps/remote_control/test/lexical/remote_control/completion_test.exs
scottming: Moving the cursor is the previous method, I forgot to update the description.

apps/server/lib/lexical/server/code_intelligence/completion.ex
scottming: Ok, I changed this.

apps/remote_control/test/lexical/remote_control/code_intelligence/ast/aliases_test.exs
scottming: I have added all the scenarios I can think of. If you have more edge cases, please feel free to let me know.

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast.ex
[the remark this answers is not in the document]
scohen: sorry, `container_cursor_to_quoted`

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast/aliases.ex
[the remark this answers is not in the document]
scohen: Can you add a comment describing why you're doing this in addition to the link? Links rot, and don't provide immediate answers to the question "why is this here?". Especially if you're on a plane.

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast/aliases.ex
scohen: > to do the match, that will trigger a matching exception
That's not how with statements work. If it fails to match, it'll take the else, not raise a `MatchError` what _you're_ doing with the `=` on the line below will raise an exception though.

scohen: example:
```elixir
with {:__MODULE__, metadata, _} <- :hi do
  IO.puts "here"
else _ ->
   IO.puts "there"
end
```

apps/remote_control/test/lexical/remote_control/code_intelligence/ast/aliases_test.exs
scohen: What about aliases that overwrite one another?
Aliases inside of a function definition? 
Aliases inside of an anonymous function and then right after it?
```elixir
alias Something.Else
def my_fn do 
  fn anony = fn -> 
    alias Another.Else
   |
  end
end
```
```elixir
alias Something.Else
def my_fn do 
  fn anony = fn -> 
    alias Another.Else
  end
  |
end
```

apps/remote_control/lib/lexical/remote_control/completion.ex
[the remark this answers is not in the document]
scottming: Ok, I pass the prefix aliases to this function, then we do not need this, and the code is much much simpler.

apps/server/lib/lexical/server/code_intelligence/completion.ex
```diff
@@ -70,6 +78,16 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
     end
   end
 
+  defp prefix_is_trigger?(env) do
```
scohen, line 81: The `Completion.Context.trigger_kind` will be `:trigger_character` if completion was triggered by a trigger character.

apps/server/lib/lexical/server/code_intelligence/completion/env.ex
```diff
@@ -291,6 +312,19 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
     end
   end
 
+  @spec prefix_aliases(t()) :: [atom() | String.t()]
+  def prefix_aliases(env) do
```
scohen, line 316: what does this do? Functions that deal with AST arcana need to be documented. I would also like to see documentation around the different clauses in the `find_value` function, telling future us what's being found.

apps/server/lib/lexical/server/code_intelligence/completion/translations/struct_field.ex
```diff
@@ -9,6 +9,20 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.StructField do
     :skip
   end
 
+  def translate(%Candidate.StructField{call?: false} = struct_field, _builder, %Env{} = env) do
+    %{name: name} = struct_field
```
scohen, line 13:
```suggestion
    name = struct_field.name
```

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -228,12 +233,132 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
     end
   end
 
+  describe "in_context?(env, :struct_arguments)" do
+    def my_module(text) do
```
scohen, line 237: naming: `wrap_with_module` "my" means nothing.

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -228,12 +233,132 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+      """
+      defmodule MyModule do
+        #{text}
+      end
+      """
+    end
+
+    def func_variable(text) do
```
scohen, line 245: `wrap_with_function` this does not declare a variable

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -228,12 +233,132 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+      """
+      def func do
+        #{text}
+      end
+      """
+    end
+
+    def func_args(text) do
```
scohen, line 253: `wrap_with_function_arguments`

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -228,12 +233,132 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+      assert in_context?(env, :struct_arguments)
+    end
+
+    test "is true when the struct is in the function arguments" do
+      env = "%User{|}" |> func_args() |> my_module() |> new_env()
+      assert in_context?(env, :struct_arguments)
+    end
+
+    test "is false when the cursor is not in a container" do
+      env = new_env("%User{|}")
```
scohen, line 276: this looks like it's in a struct reference to me.

(This is one reason I find `container_cursor_to_quoted` annoying, it seems to make decisions for you that I don't understand)

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -228,12 +233,132 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+      refute in_context?(env, :struct_arguments)
+    end
+
+    test "is true if the cursor after the field name" do
+      env = "%User{name: |}" |> my_module() |> new_env()
+      assert in_context?(env, :struct_arguments)
+    end
+
+    test "is true if the cursor after the field value" do
```
scohen, line 285:
```suggestion
    test "is true if the cursor is after the field value" do
```

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -228,12 +233,132 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+      env = "%User{|}" |> func_args() |> my_module() |> new_env()
+      assert in_context?(env, :struct_arguments)
+    end
+
+    test "is false when the cursor is not in a container" do
+      env = new_env("%User{|}")
+      refute in_context?(env, :struct_arguments)
+    end
+
+    test "is true if the cursor after the field name" do
```
scohen, line 280:
```suggestion
    test "is true if the cursor is after the field name" do
```

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -228,12 +233,132 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+      env = "%User{name: \"John\"|}" |> my_module() |> new_env()
+      assert in_context?(env, :struct_arguments)
+    end
+
+    test "is true if the cursor starts in the middle of the struct" do
+      env = "%User{name: \"John\", |}" |> my_module() |> new_env()
+      assert in_context?(env, :struct_arguments)
+    end
+
+    test "is false if the cursor after the closing curly" do
```
scohen, line 295:
```suggestion
    test "is false if the cursor is after the closing curly" do
```

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -228,12 +233,132 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+      env = "%User{}|" |> my_module() |> new_env()
+      refute in_context?(env, :struct_arguments)
+    end
+
+    test "is true if the cursor in current module arguments" do
```
scohen, line 300:
```suggestion
    test "is true if the cursor is in current module arguments" do
```

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -228,12 +233,132 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+      env = "%__MODULE__{|}" |> func_variable() |> my_module() |> new_env()
+      assert in_context?(env, :struct_arguments)
+    end
+
+    test "is true even the struct alias is at the above line" do
```
scohen, line 305:
```suggestion
    test "is true if the struct alias spans multiple lines" do
```

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -228,12 +233,132 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+        %User{
+          name: "John",
+          |
+        }
+      ]
+      env = new_env(source)
+      assert in_context?(env, :struct_arguments)
+    end
+
+    test "is true even if the value of a struct key is a tuple" do
```
scohen, line 316: 👍

apps/server/lib/lexical/server/code_intelligence/completion/env.ex
```diff
@@ -112,21 +112,42 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
...
+    |> cursor_path()
+    |> Enum.any?(fn
+      {:%, _, [{:__aliases__, _, _aliases} | _]} -> true
+      {:%, _, [{:__MODULE__, _, _} | _]} -> true
+      _ -> false
+    end)
+  end
+
+  @impl Environment
+  def in_context?(%__MODULE__{} = env, :value) do
```
scohen, line 127:
```suggestion
  def in_context?(%__MODULE__{} = env, :struct_field_value) do
```

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -228,12 +233,132 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+    end
+
+    def func_args(text) do
+      """
+      def func(#{text}) do
+      end
+      """
+    end
+
+    test "is true if the cursor after curly" do
```
scohen, line 260:
```suggestion
    test "is true if the cursor is directly after the opening curly" do
```

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -228,12 +233,132 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+    end
+  end
+
+  describe "in_context?(env, :value)" do
+    test "is true if the cursor is after a value character" do
+      env = new_env("%{foo: 1|}")
+      assert in_context?(env, :value)
+    end
+
+    test "is ture if the cursor is after a colon" do
```
scohen, line 333:
```suggestion
    test "is true if the cursor is after a colon" do
```

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -228,12 +233,132 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+    end
+  end
+
   describe "in_context?(env, :struct_reference)" do
     test "is true if the reference starts on the beginning of the line" do
       env = new_env("%User|")
       assert in_context?(env, :struct_reference)
     end
 
+    test "is false if the cursor in curly" do
```
scohen, line 357: so, here's my thinking... I actually think this should be `true`. We're in a struct reference for sure, but we're _also_ in the arguments. Thoughts?

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -739,4 +864,62 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+      source = ~q<
+        defmodule User do
+          defstruct [:name, :age]
+
+          def name(%__MODULE__{|}) do
+          end
+        end
+      >
+
+      assert [:__MODULE__] == source |> new_env() |> prefix_aliases()
```
scohen, line 879: ah, I think these should be called `aliases_at_position` or `position_aliases` 
The env's prefix is really just a single line.

apps/server/lib/lexical/server/code_intelligence/completion.ex
```diff
@@ -61,6 +61,14 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
       Enum.empty?(prefix_tokens) or not context_will_give_meaningful_completions?(env) ->
         Completion.List.new(items: [], is_incomplete: true)
 
+      Env.in_context?(env, :struct_arguments) and not Env.in_context?(env, :value) and
+          not prefix_is_trigger?(env) ->
+        aliases = Env.prefix_aliases(env)
+
+        project
+        |> RemoteControl.Api.complete_struct_fields(env.document, env.position, aliases)
```
scohen, line 69: any reason why you pass the aliases into this function rather than just calculating them where they're used? You'll safe an argument

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -739,4 +864,62 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+    end
+
+    test "when in a current module child" do
```
scohen, line 882:
```suggestion
    test "when a nested module" do
```

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -739,4 +864,62 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+          defmodule Profile do
+            defstruct [:info]
+          end
+
+          def name(%User.Profile{|}) do
+          end
+        end
+      >
+
+      assert [:User, :Profile] == source |> new_env() |> prefix_aliases()
```
scohen, line 922: So does prefix aliases just get the referenced modules in the `env`'s prefix? I think there is a better name than that, like `module_references` . 

Though I'm still struggling with `prefix_aliases`, I don't know what it does really.
What would it return for 
```elixir
foo = %Foo.User{name: "Steve", account: %Account{}}|
```
`[[:Foo, :User], :Account]`, or [:Foo, :User, :Account]`?

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -228,12 +233,132 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
     test "is true if the reference starts on the beginning of the line" do
       env = new_env("%User|")
       assert in_context?(env, :struct_reference)
     end
 
+    test "is false if the cursor in curly" do
+      env = new_env("%User{|}")
+      refute in_context?(env, :struct_reference)
+    end
+
```
scohen, line 361: Solid job of testing this.

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -739,4 +864,62 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+          def name(%User{|}) do
+          end
+        end
+      >
+
+      assert [:User] == source |> new_env() |> prefix_aliases()
+    end
+
+    test "when in a struct child" do
```
scohen, line 910: the test descriptions are pretty vague here; they sound more like something you see in a "describe". 
What are you testing? Put that in the test name.

apps/server/test/lexical/server/code_intelligence/completion/translations/struct_field_test.exs
```diff
@@ -87,4 +87,78 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.StructFieldTes
     assert completion.detail == "first_name"
     assert completion.label == "first_name"
   end
+
+  describe "in struct arguments" do
+    def my_module(text) do
```
scohen, line 92: `wrap_with_module`

apps/server/test/lexical/server/code_intelligence/completion/translations/struct_field_test.exs
```diff
@@ -87,4 +87,78 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.StructFieldTes
...
+        end
+      ]
+
+      {:ok, [completion, _]} =
+        project
+        |> complete(source)
+        |> fetch_completion(kind: :field)
+
+      assert completion.insert_text == "last_login_at: ${1:last_login_at}"
+      assert completion.insert_text_format == :snippet
```
scohen, line 139: maybe use `apply_completion` rather than asserting the snippet.

scohen (commented):
It's looking pretty good now. I especially liked all the testing about `struct_arguments`; I had some concerns with naming and I think `prefix_aliases` needs description / fleshing out /more tests.

apps/server/lib/lexical/server/code_intelligence/completion/env.ex
scohen, line 316:
```suggestion
  def prefix_aliases(%__MODULE__{} = env) do
```

scohen:
I tried this out, and got a stack trace:
completion context: 
```elixir
  def new(%Project{} = project, %Document{} = document, %Position{} = cursor_position) do
    %Document{path: path, ur|}
    zero_based_character = cursor_position.character - 1
  ...
```



Stack trace: 
```
19:48:14.962 [error] ** (ErlangError) Erlang error: {:exception, :function_clause, [{Access, :get, [{:error, {[line: 37, column: 25], "unexpected expression after keyword list. Keyword lists must always come last in lists and maps. Therefore, this is not allowed:\n\n    [some: :value, :another]\n    %{some: :value, another => value}\n\nInstead, reorder it to be the last entry:\n\n    [:another, some: :value]\n    %{another => value, some: :value}\n\nSyntax error after: ", "','"}}, :Document, nil], [file: 'lib/access.ex', line: 283]},
 {LXRelease.RemoteControl.CodeIntelligence.Ast, :expand, 3, [file: 'lib/lexical/remote_control/code_intelligence/ast.ex', line: 9]}, {LXRelease.RemoteControl.Completion, :struct_fields, 3, [file: 'lib/lexical/remote_control/completion.ex', line: 29]}]}
    (kernel 8.3.2.3) erpc.erl:505: :erpc.call/5
    (server 0.1.0) lib/lexical/server/code_intelligence/completion.ex:69: LXRelease.Server.CodeIntelligence.Completion.completions/3
    (server 0.1.0) lib/lexical/server/code_intelligence/completion.ex:37: LXRelease.Server.CodeIntelligence.Completion.complete/4
    (server 0.1.0) lib/lexical/server/provider/handlers/completion.ex:12: LXRelease.Server.Provider.Handlers.Completion.handle/2
    (server 0.1.0) lib/lexical/server/provider/queue.ex:99: anonymous fn/2 in LXRelease.Server.Provider.Queue.State.as_task/2
    (elixir 1.13.4) lib/task/supervised.ex:89: Task.Supervised.invoke_mfa/2
    (elixir 1.13.4) lib/task/supervised.ex:34: Task.Supervised.reply/4
    (stdlib 3.17.2.3) proc_lib.erl:226: :proc_lib.init_p_do_apply/3
```
FYI, this is under elixir 1.13 / erlang 24, as defined in our `.tools-versions`

The problem is that `Aliases` fails to parse the context, it's due to `container_cursor_to_quoted` returning an error. 
I actually messed up a little bit `Aliases.at` should return an ok tuple, as it can return an error tuple. I'll fix.

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
scottming: it should return []

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
scottming: Semantically it is true, but from a usage point of view, struct reference is currently only used for struct completions, and it requires changes to the implementation of **struct references**.

If we need to change it, we must do it in another PR. From a practical point of view, I don't see a very strong reason at the moment

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
scottming: right, I remove them and add more tests in the `remote_control` app.

apps/server/lib/lexical/server/code_intelligence/completion.ex
scottming: I don't quite understand this.

apps/server/test/lexical/server/code_intelligence/completion/translations/struct_field_test.exs
scottming: `apply_completion` requires a `TextEdit{}`. I think a normal snippet would suffice here, as it is not a structural complement like ` %Strr|`, which modifies the position.

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
scottming: I have no better idea.

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast.ex
```diff
@@ -52,13 +52,26 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Ast do
   @spec expand_aliases(
           document :: Document.t(),
           position :: Position.t(),
-          module_aliases :: list()
+          module_aliases :: list(),
+          already_in_container? :: boolean()
```
scottming, line 56: I think this function is not just used by the struct fields, so I also put the `already_in_container` arguments here too.

apps/server/lib/lexical/server/code_intelligence/completion.ex
scohen: If you want to figure out if the completion was caused by a trigger character (which is the name of this function), check the `Completion.Context.trigger_kind`. If the value is `trigger_character` then completion was caused by a trigger character.

apps/server/lib/lexical/server/code_intelligence/completion/env.ex
```diff
@@ -112,21 +113,42 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
     end
   end
 
+  @impl Environment
+  def in_context?(%__MODULE__{} = env, :struct_arguments) do
+    cursor_path = Ast.cursor_path(env.document, env.position)
+
+    Enum.any?(cursor_path, fn
+      {:%, _, [{:__aliases__, _, _aliases} | _]} -> true
+      {:%, _, [{:__MODULE__, _, _} | _]} -> true
```
scohen, line 122: All AST needs comments describing what it is.

apps/remote_control/lib/lexical/remote_control/code_mod/ast/aliases.ex
```diff
@@ -298,4 +289,32 @@ defmodule Lexical.RemoteControl.CodeMod.Ast.Aliases do
...
+
+      _error ->
+        # NOTE: Adding new line not always works,
+        # so we need to try again without adding new line
+        document_fragment = Document.fragment(doc, position)
+        container_cursor_to_quoted(document_fragment)
+    end
+  end
+
+  def container_cursor_to_quoted(document_fragment) do
```
scohen, line 314: why's this public? Shouldn't we always use `safe_container_cursor_to_quoted`? If so, just call the other one `container_cursor_to_quoted`

apps/server/lib/lexical/server/code_intelligence/completion.ex
scohen: @scottming this is still unaddressed.

apps/server/lib/lexical/server/code_intelligence/completion/env.ex
```diff
@@ -112,21 +113,46 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
...
+      # or just a struct: `%Struct{|}`
+      {:%, _, [{:__aliases__, _, _aliases} | _]} -> true
+      # current module struct: `%__MODULE__{|}`
+      {:%, _, [{:__MODULE__, _, _} | _]} -> true
+      _ -> false
+    end)
+  end
+
+  @impl Environment
+  def in_context?(%__MODULE__{} = env, :struct_field_value) do
```
scohen, line 132: won't this return true for arguments and static keyword lists? 
I think you'll need to ensure you're in struct arguments too.

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -228,12 +233,132 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+    test "is true if the cursor starts in the middle of the struct" do
+      env = "%User{name: \"John\", |}" |> wrap_with_module() |> new_env()
+      assert in_context?(env, :struct_arguments)
+    end
+
+    test "is false if the cursor is after the closing curly" do
+      env = "%User{}|" |> wrap_with_module() |> new_env()
+      refute in_context?(env, :struct_arguments)
+    end
+
```
scohen, line 299:
```elixir
test "is false in static keywords" do 
  env = "[foo: |]" |> wrap_with_module() |> new_env() 
  refute in_context?(env, :struct_arguments)
end
```

Also add the same test with function arguments and another for map keys.
Though now that I think of this, what you're really checking for isn't struct arguments, it's map update syntax.

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
scohen: aside from things being correctly named, and doing one thing and one thing well? To me, there are two things you're checking for, and they've been muddled together, which is why things aren't working well. They are:

1. We're inside a struct reference. 
2. We're inside map / keyword update syntax. 
When both are true, we're doing what you want. We can improve ergonomics on `in_context?` if you like, by doing something like;
```elixir
in_context?(env, [:struct_reference, :map_update])
```
Which will return true if you're in both a struct reference and a map update reference.

scohen: > and it requires changes to the implementation of struct references.

I don't understand, why?

apps/server/lib/lexical/server/code_intelligence/completion.ex
scottming: > If you want to figure out if the completion was caused by a trigger character (which is the name of this function)

I think `trigger_character` is not the same as this function,  like if we print the `trigger_kind`, all of the tests of `  describe "in struct arguments" do` in `struct-fields` are `trigger_character`.

but what I want is: if the prefix is one of `[".", "@", "&", "%", "^", ":", "!", "-", "~"]`, then don't trigger the struct fields completions.

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
scottming: I think I can remove this test because this test is for `struct_references` not `struct_arguments`, but my pr didn't change any product code of `in_context(env, :struct_references)`

I really don't want to keep this pr inflated, of course we can move this discussion to a new issue, and use another PR to solve this.

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast.ex
```diff
@@ -0,0 +1,77 @@
...
+      defmodule Issue do
+        defstruct [:message]
+      end
+
+      def message(%Issue{|} = issue) do # cursor marked as `|`
+      end
+    end
+    ```
+
+  Then the the expaned module is `Project.Issue`.
```
scohen, line 28:
```suggestion
  Then the the expanded module is `Project.Issue`.
```

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast.ex
```diff
@@ -0,0 +1,77 @@
...
+
+    defmodule MyModule do
+      alias Project, as: MyProject
+
+      def message(%MyProject.Issue{|} = issue) do
+      end
+    end
+    ```
+
+  Then the the expaned module is still `Project.Issue`.
```
scohen, line 47:
```suggestion
  Then the the expanded module is still `Project.Issue`.
```

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast.ex
```diff
@@ -0,0 +1,77 @@
...
+
+  And sometimes we can't find the full name by the `Aliases.at/2` function,
```
scohen, line 49: why?

apps/remote_control/lib/lexical/remote_control/code_mod/ast/aliases.ex
```diff
@@ -298,4 +289,32 @@ defmodule Lexical.RemoteControl.CodeMod.Ast.Aliases do
...
+    %{line: line} = position
+    added_new_line_position = Position.new(line + 1, 1)
+    document_fragment = Document.fragment(doc, added_new_line_position)
+
+    case container_cursor_to_quoted(document_fragment) do
+      {:ok, quoted} ->
+        {:ok, quoted}
+
+      _error ->
+        # NOTE: Adding new line not always works,
```
scohen, line 307:
```suggestion
        # NOTE: Adding new line doesn't always work,
```

scohen (approved):
Some small comments / changes, mostly in documentation. 
This looks ready.

apps/remote_control/lib/lexical/remote_control/code_intelligence/ast.ex
scottming: For example, without alias, like `%Foo.Structs.Struct{|}`

PR #197, opened 2023-05-31
We now use Lexical.Plugin.Diagnostics

scohen:
Switched from using Mix.Compiler.Diagnostic to our plugin diagnostics. This is so we can handle diagnostics from everywhere the same way.

While I was converting error.ex, I found that we went through great lengths to extract the file path from the error messages, but we actually had it right next to us in the build, so I passed in the document, which eliminated some complexity there.


projects/lexical_plugin/lib/lexical/plugin/diagnostic/result.ex
```diff
@@ -1,22 +1,40 @@
...
+  @type severity :: :message | :warning | :error
+  @type mix_position ::
+          non_neg_integer()
+          | {pos_integer(), non_neg_integer()}
+          | {pos_integer(), non_neg_integer(), pos_integer(), non_neg_integer()}
+
+  @type position :: mix_position() | Document.Range.t() | Document.Position.t()
   @type t :: %__MODULE__{
-          location: Lexical.Document.Range.t() | Lexical.Document.Position.t(),
+          position: position,
```
scottming, line 14: Is there any consideration for using "position" instead of "location" here?

scottming (approved):

apps/server/lib/lexical/server/project/diagnostics.ex
```diff
@@ -41,39 +41,15 @@ defmodule Lexical.Server.Project.Diagnostics do
       %__MODULE__{state | diagnostics_by_uri: cleared}
     end
 
-    def add(%__MODULE__{} = state, %Compiler.Diagnostic{} = diagnostic) do
-      source_uri = Document.Path.to_uri(diagnostic.file)
-
+    def add(%__MODULE__{} = state, %Diagnostic.Result{} = diagnostic) do
```
scottming, line 44: So, will the logic for cleaning up Diagnostic plugins be in another PR?

scohen: What am I missing here? I'm thinking about these three cases. 

1. The project is compiled
  a. The file is not synced: We keep the existing diagnostics, because they're likely more accurate due to as-you-type
  b. The file is synced: We delete all diagnostics when we get word the project is going to be compiled. Plugins will then receive a request to re-diagnose the entire project
2. A file is compiled: We detect the file has changed and delete the file's diagnostics completely. We will then ask plugins to diagnose the file.

projects/lexical_plugin/lib/lexical/plugin/diagnostic/result.ex
scohen: Yeah, I felt that despite elixir calling them locations, I went with the lsp terminology. They're very close in meaning.

apps/server/lib/lexical/server/project/diagnostics.ex
scottming: These cases are logically correct, but you missed that plugin cannot clear each other's diagnostics. 

For example, if my project uses both the `recode` and `credo` plugins when receiving diagnostics from `recode,` we cannot clear the diagnostics generated by `credo` because they have different rules, whether at the project level or file level.

scohen: plugins don't clear one another's diagnostics, the act of compiling the project or the file does. I don't see what the issue you're citing is. Can you be more spercific.

say we have recode and credo, and you change a file. Lexical kills the deps in the file and then runs both recode and credo, which will produce more diagnostics (or not).

scohen: the sequence of events is this:
1. The file is changed. 
2. The diagnostics server is notified of the change
3. The diagnostics server clears the file's diagnostics
4. The file is recompiled, and compilation diagnostics are sent to the diagnostic server
5. All diagnostics plugins are asked to diagnose the file. Any emitted diagnostics go the diagnostic server.

At every receipt of diagnostics, the diagnostic server publishes the diagnostics it has, so the client will get ever more diagnostics.

Issue #178, opened 2023-05-23
Enhance completion on test file context


scottming:
<img width="573" alt="image" src="https://github.com/lexical-lsp/lexical/assets/12830256/7ebe3f89-d174-427f-ac41-7b454d5e6b15">

These three completions won't help, same as the completions of `descri|` and `asser|`, because no one writes code like this:

```elixir
  test(message, var \\ quote do
    _
  end, contents)
```

It should be 

<img width="572" alt="image" src="https://github.com/lexical-lsp/lexical/assets/12830256/c0bece41-2c12-4497-b0a7-ba4aa4207101">

This is provided by luasnip.

scohen:
All that needs to happen here is to detect the `test` macro in `server/code_intelligence/completion/translations/macro.ex` and provide snippets.

Sleepful:
I cannot repro this, my suggestions come without the `quote` part
<img width="609" alt="Screen Shot 2023-06-03 at 12 01 02" src="https://github.com/lexical-lsp/lexical/assets/7144046/c4f8e43e-c297-4ec2-9ebf-f4736bc0b3b7">

scohen:
The quote thing was due to the way elixir sense reports argument names. That should be fixed now

scottming:
@Sleepful Steve fixed the quote thing, but we still need to remove the parentheses

scohen:
You don't need to remove anything, all you need to do is add another clause in the macro translation I linked to above.

scohen:
To be more clear: The reason you're seeing this behavior is that the macro translation is taking this code path:
https://github.com/lexical-lsp/lexical/blob/6c4545ffcafaac37bed1012d12e63a8aaa8b434f/apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex#L448

All that needs to be done is to add another function head before that path that checks for test functions and that returns a snippet (maybe several --one for no arguments, one with a context map, etc). 

It's maybe a five-line change, excepting tests.

Sleepful:
Thank you, it is very clear what I need to do 😁 
The code for adding a snippet is very clear! I also see why it would be nice for library authors to add plugins so they can add their own snippets

PR #208, opened 2023-06-07
[issue-178] Snippet translations for  macro

Sleepful:
Open PR, should be ready for test macros

### DIFF
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
@@ -6,7 +6,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
 
   use Translatable.Impl, for: Candidate.Macro
 
-  @snippet_macros ~w(def defp defmacro defmacrop defimpl defmodule defprotocol defguard defguardp defexception)
+  @snippet_macros ~w(def defp defmacro defmacrop defimpl defmodule defprotocol defguard defguardp defexception test)
 
   def translate(%Candidate.Macro{name: name}, _builder, _env)
       when name in ["__before_compile__", "__using__", "__after_compile__"] do
@@ -436,6 +436,62 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
       label: label
     )
     |> builder.boost()
+  end
+
+  @stub_label ~S(test "message"           )
+  @plain_label ~S(test "message" do...     )
+  @context_label ~S(test "message", %{} do...)
+
+  def translate(%Candidate.Macro{name: "test", arity: 1}, builder, env) do
+    stub_label = @stub_label
+
+    stub_snippet = """
+    test "${0:message}"
+    """
+
+    env
+    |> builder.snippet(stub_snippet,
+      detail: "A stub test",
+      kind: :class,
+      label: stub_label
+    )
+    |> builder.boost(1)
+  end
+
+  def translate(%Candidate.Macro{name: "test", arity: 2}, builder, env) do
+    plain_label = @plain_label
+
+    plain_snippet = """
+    test "${1:message}" do
+      $0
+    end
+    """
+
+    env
+    |> builder.snippet(plain_snippet,
+      detail: "A test",
+      kind: :class,
+      label: plain_label
+    )
+    |> builder.boost(2)
+  end
+
+  def translate(%Candidate.Macro{name: "test", arity: 3}, builder, env) do
+    context_label = @context_label
+
+    context_snippet = """
+    test "${1:message}", %{${2:context}} do
+      $0
+    end
+    """
+
+    env
+    |> builder.snippet(context_snippet,
+      detail: "A test that receives context",
+      kind: :class,
+      label: context_label
+    )
+    |> builder.boost(3)
   end
 
   def translate(%Candidate.Macro{name: "__MODULE__"} = macro, builder, env) do
--- a/apps/server/test/lexical/server/code_intelligence/completion/translations/macro_test.exs
+++ b/apps/server/test/lexical/server/code_intelligence/completion/translations/macro_test.exs
@@ -612,4 +612,38 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.MacroTest do
       refute boosted?(completion)
     end
   end
+
+  test "test completion snippets", %{project: project} do
+    assert {:ok, [stub, with_body, with_context | _ignored]} =
+             project
+             |> complete(inside_exunit_context("test|"))
+             |> fetch_completion("test ")
+
+    assert ~S(test "message"           ) = stub.label
+    assert "A stub test" = stub.detail
+    assert :snippet = stub.insert_text_format
+    assert "test \"${0:message}\"\n" = stub.insert_text
+
+    assert ~S(test "message" do...     ) = with_body.label
+    assert "A test" = with_body.detail
+    assert :snippet = with_body.insert_text_format
+
+    assert "test \"${1:message}\" do\n  ${0:body}\nend\n" = with_body.insert_text
+
+    assert ~S(test "message", %{} do...) = with_context.label
+    assert "A test that receives context" = with_context.detail
+    assert :snippet = with_context.insert_text_format
+
+    assert "test \"${1:message}\", %{${2:context}} do\n  ${0:body}\nend\n" =
+             with_context.insert_text
+  end
+
+  defp inside_exunit_context(text) do
+    ~q[
+      defmodule Test do
+        use ExUnit.Case
+
+        #{text}
+    ]
+  end
 end


apps/server/test/lexical/server/code_intelligence/completion/translations/macro_test.exs
```diff
@@ -612,4 +612,29 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.MacroTest do
...
     end
   end
+
+  test "test has three completions", %{project: project} do
+    # Error:
+    # right: {:error, :not_found}
+    assert {:ok, completion} =
+             project
+             |> complete("test|")
+             |> fetch_completion("test ")
```
Sleepful, line 622: so I was expecting this to work fine, but I suppose I need to have ExUnit in the context of `project`. I need to explore the code for `project` more to figure out how to do this.

If you already know how to do this easily, please tell me :D

apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
```diff
@@ -6,13 +6,62 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
...
+    test "${1:name}" do
+      $0
+    end
+    """
+
+    builder.snippet(env, snippet,
+      detail: macro.spec,
+      kind: :class,
+      label: label,
+      sort_text: builder.boost(label, 10)
```
Sleepful, line 44: I copy/pasted these from another test but these properties are not clear to me:
`kind` and `sort_text`

scohen: They're part of the LSP spec. `kind` is the type of completion you're completing, for example: `:function` or `:macro`. Sort text is what it sounds like, text that's used when ordering results by the editor.

apps/server/test/lexical/server/code_intelligence/completion/translations/macro_test.exs
scohen: When you use the `complete` function, you're passing in source code, and all the context that it has. The entirety of your source code here is `test|`, and as `test` isn't something imported from `Kernel`, you have no completions. Instead, give the completer more context, a you would in an actual test

```elixir
code = ~q[
defmodule FakeTest do 
  use ExUnit.case 
  te|
]

assert {:ok, completion} = 
  project
  |> complete(code)
  |> fetch_completion("test ")
```

apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
```diff
@@ -6,13 +6,62 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
 
   use Translatable.Impl, for: Result.Macro
 
-  @snippet_macros ~w(def defp defmacro defmacrop defimpl defmodule defprotocol defguard defguardp defexception)
+  @snippet_macros ~w(def defp defmacro defmacrop defimpl defmodule defprotocol defguard defguardp defexception test)
```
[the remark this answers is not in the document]
scottming: In addition, they may have some differences compared to regular macros, such as the need to generate two snippets.

Sleepful:
Todo:

- [ ] Merge `main` and fix up boost: “Boosting now takes an item rather than the sort text, and has a different signature.”
- [ ] Improve `assert` and `asset_*`
- [ ] Look through ExUnit for anything else

apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
```diff
@@ -6,13 +6,62 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
...
+    label = "#{macro.name} (Define test/2)"
+
+    snippet = """
+    test "${1:name}" do
+      $0
+    end
+    """
+
+    builder.snippet(env, snippet,
+      detail: macro.spec,
```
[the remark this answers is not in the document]
Sleepful: I do not know. I found that macro struct is defined here https://github.com/lexical-lsp/lexical/blob/f45d0ccb0d72a87c532158525af4959a5b2ae3a6/apps/remote_control/lib/lexical/remote_control/completion/candidate.ex#L40 

I also found that the `macro` var in this function seems to come from this `local_completions` here https://github.com/lexical-lsp/lexical/blob/f45d0ccb0d72a87c532158525af4959a5b2ae3a6/apps/server/lib/lexical/server/code_intelligence/completion.ex#L73

But... I still do not know if `macro` has a `spec` in this scope, how can I find out?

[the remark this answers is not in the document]
Sleepful: does the spec come from elixir_sense? 
I chased `local_completions` and landed here:
https://github.com/lexical-lsp/lexical/blob/f45d0ccb0d72a87c532158525af4959a5b2ae3a6/apps/remote_control/lib/lexical/remote_control/api.ex#L44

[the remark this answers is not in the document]
Sleepful: So I ran the code and I see this:

```
02:21:52.675 [info] Local completions are [

%LXRelease.RemoteControl.Completion.Candidate.Function{argument_names: ["tuple"], arity: 1, name: "tuple_size", origin: "Kernel", type: :function, visibility: :public, spec: "@spec tuple_size(tuple) :: non_neg_integer", metadata: %{guard: true}}, 

%LXRelease.RemoteControl.Completion.Candidate.Macro{argument_names: ["value", "fun"], arity: 2, name: "then", origin: "Kernel", type: :macro, visibility: :public, spec: "", metadata: %{since: "1.12.0"}}, 

%LXRelease.RemoteControl.Completion.Candidate.Macro{argument_names: ["message"], arity: 1, name: "test", origin: "ExUnit.Case", type: :macro, visibility: :public, spec: "", metadata: %{}}, 

%LXRelease.RemoteControl.Completion.Candidate.Macro{argument_names: ["message", "contents"], arity: 2, name: "test", origin: "ExUnit.Case", type: :macro, visibility: :public, spec: "", metadata: %{defaults: 1}}, 

%LXRelease.RemoteControl.Completion.Candidate.Macro{argument_names: ["message", "var", "contents"], arity: 3, name: "test", origin: "ExUnit.Case", type: :macro, visibility: :public, spec: "", metadata: %{defaults: 1}}]
```

so you can see some structs in there that have `name: "test"` and `spec: ""`.

So it is indeed empty!

So I guess I just delete the `detail: macro.spec` field, correct?

apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
```diff
@@ -438,6 +438,58 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
     |> builder.boost()
   end
 
+  def translate(%Candidate.Macro{name: "test", arity: 1} = macro, builder, env) do
```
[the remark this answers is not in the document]
scohen: Overall here, I think the code is mostly fine, but the UX needs help. Some comments

For whatever reason, I don't think about test with regard to arity, so the `/arity` label isn't helping.  Similarly, the argument names are just `message`, `var` and `contents` for the `/3` function, so that's not helpful. 

What I found worked best was giving them labels like:
```elixir
  stub_label = ~S(test "message"           )
  plain_label = ~S(test "message" do...     )
  context_label = ~S(test "message", %{} do...)
```
One thing to note is that the labels are the same length, so they stack nicely. I also gave them details like "A stub test", "A test" and "A test with a context variable". Which helped to disambiguate.

apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
```diff
@@ -438,6 +438,58 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
...
+    test "${0:name}"
+    """
+
+    env
+    |> builder.snippet(snippet,
+      detail: "",
+      kind: :class,
+      label: label
+    )
+    |> builder.boost(3)
```
scohen, line 454: the higher the boost, the more important the thing; I think this is backwards, as this is just the stub.

apps/server/test/lexical/server/code_intelligence/completion/translations/macro_test.exs
```diff
@@ -612,4 +612,50 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.MacroTest do
...
+             project
+             |> complete(inside_exunit_context("test|"))
+             |> fetch_completion("test ")
+
+    assert %Completion.Item{
+             label: "test (Define an empty test/1)",
+             detail: "",
+             insert_text_format: :snippet,
+             insert_text: "test \"${0:name}\"\n"
+           } = get_in(completions, [Access.at(0)])
```
scohen, line 636: I really don't like assertions like this; If something goes wrong, you have to parse a diff. I much prefer assertions like:
```elixir
assert item.label == "test ..."
assert item.insert_text_format == "snippet"
```
Additionally, doing assertions on the snippets is prescriptive; we have an `apply_completion` function, so you can ensure the outcome is what you expect

apps/server/test/lexical/server/code_intelligence/completion/translations/macro_test.exs
```diff
@@ -612,4 +612,50 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.MacroTest do
...
+    assert [%Completion.Item{}, %Completion.Item{}, %Completion.Item{}] = completions
+  end
+
+  test "test completion snippets", %{project: project} do
+    assert {:ok, completions} =
+             project
+             |> complete(inside_exunit_context("test|"))
+             |> fetch_completion("test ")
+
+    assert %Completion.Item{
```
scohen, line 631: Why not just match out the completions above rather than use `get_in` and `Access.at(..)`?

```elixir
assert {:ok, [stub, with_body, with_context]} = ...
```

apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
Sleepful: I was thinking in terms of chronological order, stub coming first. But I think your suggestion is better.

apps/server/test/lexical/server/code_intelligence/completion/translations/macro_test.exs
Sleepful: I just got a little used to tossing around the lensing functions (e.g. `get_in`). But I like your idea better.

apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
```diff
@@ -6,13 +6,62 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
...
+    end
+    """
+
+    builder.snippet(env, snippet,
+      detail: macro.spec,
+      kind: :class,
+      label: label,
+      sort_text: builder.boost(label, 10)
+    )
+  end
```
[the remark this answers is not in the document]
Sleepful: returning list not working, I attempted to do it that way here: https://github.com/lexical-lsp/lexical/pull/237
it multiplies each snippet by 3 because `translate` gets called 3 times it seems. I assume.

apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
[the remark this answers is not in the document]
Sleepful: Updated, I thought `A test that receives context` sounds a bit more specific

apps/server/test/lexical/server/code_intelligence/completion/translations/macro_test.exs
Sleepful: > If something goes wrong, you have to parse a diff

Oh such a good point

Sleepful:
@scohen ready :) 
consolidated into a single commit

apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
```diff
@@ -438,6 +438,62 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
...
+    )
+    |> builder.boost(1)
+  end
+
+  def translate(%Candidate.Macro{name: "test", arity: 2}, builder, env) do
+    plain_label = @plain_label
+
+    plain_snippet = """
+    test "${1:message}" do
+      ${0:body}
```
scohen, line 466:
```suggestion
      $0
```
I think adding body is a little annoying, it's not syntactically valid, and if you miss a tab stop, it causes an error.

scohen (approved):
Fix the one comment, then you're good.

scohen:
Fixes #178

PR #209, opened 2023-06-07
Delete `deps.compile` from README

scottming:
More than one person has reported to me that they failed at `deps.compile`. I suspect it's caused by the relative path, and besides, our release doesn't need to compile because it will automatically compile. Therefore, I deleted this line to reduce misunderstandings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,6 @@ To install lexical fetch the source code from git, then do the following:
 
  ```
  mix deps.get
- mix deps.compile
  mix release lexical
  ```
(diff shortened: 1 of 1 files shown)


scohen:
I'd be more interested in finding out why `deps.compile` failed, or why they think it failed, because `compile` calls `deps.compile`.

scottming:
telegram-cloud-photo-size-5-6336765581851736149-y
but Unfortunately, I can't replicate this error.

Issue #60, opened 2023-03-23
Diagnostics persistence issues


scohen:
I've noticed the following issues with diagnostics:

1. Presently, they disappear when you save your file, and then reappear when you type. This is due to the fix that swapped lines 39 and 42 in `diagnostics.ex`. Swapping these lines back fixes this issue.
2. as-you-type diagnostics aren't reliably cleared when you save. I get a lot of `function Logger.in doesn't exist` at a specific line, even though i've saved the file with the correct code of `Logger.info`.

PR #211, opened 2023-06-09
Plugin Architecture

### LINKED ISSUE #207: plugin empty diagnostics will clear mix diagnostics
I noticed that there is a delay in Diagnostics when editing now. It only appears when I stop editing, but not while actively making changes. This creates a significant difference in user experience compared to before and doesn't feel as real-time.

I think this will cause **a lot of disturbance** to users, because the Diagnostic will keep flashing during the editing process and it is not accurate. You can see the difference by watching these two 10-second videos below.

## now

https://github.com/lexical-lsp/lexical/assets/12830256/f10705ef-1c00-4bb1-94ea-483794495770

## before

https://github.com/lexical-lsp/lexical/assets/12830256/8d49070b-4a53-4915-99b3-225750548f13

This is because we pushed empty Diagnostics during the editing process. When the plugin diagnostics is empty, we should not push Diagnostics because this will overwrite our original Diagnostics and cause flickering.

you can comment on this line: https://github.com/lexical-lsp/lexical/blob/30b96d9/apps/server/lib/lexical/server/state.ex#L104 to let this problem go, but we need to fix this.

scohen:
Three commits here, review each separately.

Fixes #207 and #60


scohen:
One incidental change is that builds now have build numbers, which are monotonically incrementing integers. This has the result of drastically simplifying the logic behind clearing diagnostics. So far, I have not seen any lingering diagnostics I mentioned in #60 or the "flashing" behavior that Scott mentioned in  #207.

projects/lexical_plugin/lib/lexical/plugin/diagnostic/result.ex
```diff
@@ -1,15 +1,77 @@
 defmodule Lexical.Plugin.Diagnostic.Result do
+  @moduledoc """
+  The result of a diagnsotic run
```
zolrath, line 3:
```suggestion
  The result of a diagnostic run
```

scohen: Thank you!

projects/lexical_credo/lib/lexical_credo.ex
```diff
@@ -0,0 +1,102 @@
...
+      :ok
+    end
+  end
+
+  def document do
+    %Document{}
+  end
+
+  def handle(%Document{} = doc) do
+    Application.ensure_all_started(:credo)
```
zachallaun, line 20: Is this needed here as well as in `init`?

(Context: reviewing this PR for the sake of an experimental Mneme plugin 🙂)

projects/lexical_plugin/README.md
```diff
@@ -1,6 +1,94 @@
...
+## Overview
+Plugins are used to extend lexical's functionality and provide opt-in diagnostics, completion, and code intelligence features without having to build inside of Lexical itself.
+
+## Plugin Types 
+The goal of the plugin project is to have different types of plugins that affect lexical in different ways. Presently, only diagnostic plugins are supported. A diagnostic examines a mix project or a document and emits `Lexical.Plugin.Diagnostic.Result` structs to direct the user's attention to any issues it finds. 
+Diagnostic plugins can be used to integrate code linters like `credo` or to enforce project-specific rules and coding practices.
+
+### Creating a diagnostic plugin
+
+Create a new mix project with `mix new my_plugin`, and edit the `mix.exs` file to include the `:lexical_plgin` app.
```
zachallaun, line 14: `lexical_plgin` typo

projects/lexical_credo/lib/lexical_credo.ex
scohen: It is not

apps/remote_control/lib/lexical/remote_control/plugin/discovery.ex
```diff
@@ -0,0 +1,64 @@
...
+  will be looking for structs like `Lexical.Document', and be passed in structs like `LXRelease.Document`,
+  and the plugin will crash.
+  """
+  alias Lexical.Plugin
+  alias Mix.Tasks.Namespace
+
+  require Logger
+
+  def run do
+    for {app_name, _, _} <- :application.loaded_applications(),
```
scottming, line 18: There's a bug here; I found that the `run/0` only works when you first open a project(I mean, `.lexical` does not exist).

apps/server/lib/lexical/server/project/diagnostics.ex
```diff
@@ -114,13 +59,20 @@ defmodule Lexical.Server.Project.Diagnostics do
...
+        file_diagnostics(uri: uri, build_number: build_number, diagnostics: diagnostics),
+        %State{} = state
+      ) do
     state =
-      Enum.reduce(diagnostics, state, fn diagnostic, state ->
-        State.add(state, diagnostic)
-      end)
+      case diagnostics do
+        [] ->
+          State.clear(state, uri)
```
scottming, line 69: Yes, now the flashing situation seems to have disappeared in editing mode. However, I don't quite understand why it can be cleared directly when empty.

mix.exs
```diff
@@ -21,6 +21,7 @@ defmodule Lexical.LanguageServer.MixProject do
     [
       {:ex_doc, "~> 0.29.1", only: :dev, runtime: false},
       {:credo, "~> 1.7", only: [:dev, :test]},
+      {:lexical_credo, path: "projects/lexical_credo"},
```
scottming, line 24: don't you need to add `only: [:dev, :test]` here?

projects/lexical_credo/lib/lexical_credo.ex
```diff
@@ -0,0 +1,102 @@
...
+  alias Lexical.Document
+  alias Lexical.Plugin.Diagnostic
+  alias Lexical.Project
+
+  use Diagnostic, name: :lexical_credo
+  require Logger
+
+  def init do
+    with {:ok, _} <- Application.ensure_all_started(:credo) do
+      :ok
```
scottming, line 11: We don't need the CLI output, so I think we can call the `    GenServer.call(Credo.CLI.Output.Shell, {:suppress_output, true})` here

projects/lexical_credo/lib/lexical_credo.ex
```diff
@@ -0,0 +1,102 @@
...
+
+    execution_args = [
+      "--mute-exit-status",
+      "--read-from-stdin",
+      Document.Path.absolute_from_uri(doc.uri)
+    ]
+
+    execution = Credo.Execution.build(execution_args)
+
+    with_stdin(
```
scottming, line 32: I think we should try to keep the operation results consistent with `mix credo`, so we don't need to perform real-time credo checks for excluded files.

projects/lexical_credo/lib/lexical_credo.ex
scohen: @zachallaun What kind of plugins do you need in order to get Mneme working? I'm assuming diagnostics aren't enough, and you'll also need code actions.

apps/remote_control/lib/lexical/remote_control/plugin/discovery.ex
scohen: This can't be true; I've run this repeatedly with a project where `.lexical` exists.

apps/remote_control/lib/lexical/remote_control/build/state.ex
```diff
@@ -50,10 +55,15 @@ defmodule Lexical.RemoteControl.Build.State do
   end
 
   def compile_project(%__MODULE__{} = state, force?) do
+    state = increment_build_number(state)
     project = state.project
 
     Build.with_lock(fn ->
-      {elapsed_us, result} = :timer.tc(fn -> safe_compile_project(state.project, force?) end)
+      compile_requested_message =
+        project_compile_requested(project: project, build_number: state.build_number)
```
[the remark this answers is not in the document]
scohen: They won't though, again, there's no overwriting, only appending, unless the build number increments, and then they replace what's there. 
The reason that we don't need the build number in the project compiled message is that we've already received it in the `project_compile_requested` message. Despite this looking like a distributed system, it's really not one, so the messages are always in order, so you'll always get the `project_compile_requested` before the `project_compiled` message, and you can plan accordingly.

apps/server/lib/lexical/server/project/diagnostics.ex
scohen: Empty means we've run the plugins and normal diagnostics and they've returned nothing.

apps/remote_control/lib/lexical/remote_control/build/state.ex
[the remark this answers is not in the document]
scohen: One thing I had thought of doing is making plugin specific messages, that will make clearing things slightly simpler, but I didn't find any bugs with the current method.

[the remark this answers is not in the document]
scottming: You can replicate this bug by creating a function like the one below:

```elixir
defmodule WithCredoPlugin do
  @moduledoc false
  def foo() do
    f
  end
end
```

And the diagnostics should look like 

image

But when you save that file, it will become 

image

the diagnostics from `Mix` disappeared

apps/server/lib/lexical/server/provider/queue.ex
```diff
@@ -44,7 +44,7 @@ defmodule Lexical.Server.Provider.Queue do
     @spec cancel(t, pos_integer()) :: t
     def cancel(%__MODULE__{} = state, request_id) do
       with {:ok, %Task{} = task} <- Map.fetch(state.tasks_by_id, request_id),
-           true <- Process.exit(task.pid, :cancelled) do
+           true <- Process.exit(task.pid, :normal) do
```
scottming, line 47: Why change this?

this change will cause a completions bug in neovim.

image

just try to create a function by typing: `def foo|` can replicate this error.

apps/protocol/lib/lexical/protocol/json_rpc.ex
```diff
@@ -32,8 +32,7 @@ defmodule Lexical.Protocol.JsonRpc do
     {:ok, json_rpc}
   end
 
-  defp do_decode(%{"id" => id, "result" => nil}) do
-    Logger.info("Reply for id #{id} had no body")
+  defp do_decode(%{"id" => _id, "result" => nil}) do
```
scottming, line 35: I think you do not need the `require Logger` line right now.

apps/server/lib/lexical/server/provider/queue.ex
scottming:
```lua

[ERROR][2023-06-13 08:35:21] .../vim/lsp/rpc.lua:493	"No callback found for server response id 3"
[ERROR][2023-06-13 08:35:21] .../lua/vim/lsp.lua:1086	"LSP[lexical]"	"on_error"	{
  code = "NO_RESULT_CALLBACK_FOUND",
  err = {
    id = 3,
    jsonrpc = "2.0",
    result = { {
        detail = "",
        insertText = "def ${1:name}($2) do\n  $0\nend\n",
        insertTextFormat = 2,
        kind = 7,
        label = "def (Define a function)",
        sortText = " def (Define a function)"
      }, {
        detail = "",
        insertText = "defdelegate ${1:call}, to: ${2:module} $0\n",
        insertTextFormat = 2,
        kind = 7,
        label = "defdelegate (Define a delegate function)",
        sortText = "%defdelegate (Define a delegate function)"
      }, {
        detail = "",
        insertText = "defexception [${1:fields}] $0\n",
        insertTextFormat = 2,
        kind = 7,
        label = "defexception (Define an exception)",
        sortText = "%defexception (Define an excepti
        ```

scohen: not in an editor that _isn't_ neovim; What is going on here?

apps/remote_control/lib/lexical/remote_control/build/state.ex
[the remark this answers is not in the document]
scohen: Alright, figured this out. It should be fixed now. Unresolve this if you think it's broken.

apps/server/lib/lexical/server/provider/queue.ex
scohen: > Why change this?

Because it was logging an error because of a totally normal thing; the task was canceled by the client. No big deal.

> this change will cause a completions bug in neovim.

How? This should be completely internal to the VM, the task still gets canceled by the client.

scohen: Let me try cancelling this a different way, and see if it helps.

scohen: I've pushed a change, see if it works for you.

projects/lexical_credo/lib/lexical_credo.ex
scohen: The answer is that credo doesn't listen to the excludes or includes when using stdin as input.

scottming (approved):

PR #212, opened 2023-06-09
Module sorting / Refactor boost

### LINKED ISSUE #202: Place module completions ahead of function completions
I keep feeling that there are typically fewer modules suggested and I want to see them at the top. iex works like this, and I like it better than seeing functions first; modules can have a lot, and navigating to submodules is tough.

scohen:
This pr does two things:

 1. Reorders completions so that modules come before functions
 2. Refactors boost so that it operates on Completion.Item structs

I think the first thing is a welcome change; it's the default for most LS implementations to sort Modules before functions, and the makes boosting things easier. All items now have a default boost of 0, and items can be boosted with a number from 0-9 to boost items of the same kind, or numbers divisible by 10 to boost across item kinds.

Fixes #202

### DIFF
--- a/apps/server/test/lexical/server/code_intelligence/completion_test.exs
+++ b/apps/server/test/lexical/server/code_intelligence/completion_test.exs
@@ -92,23 +92,23 @@ defmodule Lexical.Server.CodeIntelligence.CompletionTest do
     end
   end
 
-  describe "sort_text" do
-    test "dunder functions have the dunder removed in their sort_text", %{project: project} do
+  describe "sorting" do
+    test "dunder functions aren't boosted", %{project: project} do
       assert {:ok, completion} =
                project
                |> complete("Enum.|")
                |> fetch_completion("__info__")
 
-      assert completion.sort_text == "info/1"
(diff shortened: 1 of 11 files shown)


apps/server/test/lexical/server/code_intelligence/completion_test.exs
```diff
@@ -152,5 +158,38 @@ defmodule Lexical.Server.CodeIntelligence.CompletionTest do
       assert {:ok, _} = fetch_completion(completions, label: "Foo-protocol")
       assert {:ok, _} = fetch_completion(completions, label: "Foo-struct")
     end
+
+    test "modules are sorted before functions", %{project: project} do
```
scottming, line 162: You mentioned in the document that local boost and global boost have different usage. Can we have more unit tests to demonstrate use cases?

scottming (approved):

apps/server/test/lexical/server/code_intelligence/completion_test.exs
scohen: Great idea

PR #216, opened 2023-06-13
Refactor: Completion.Results are now Completion.Candidates

scohen:
There are a lot of things called `Result` in this codebase. Upon using the completion results, it's pretty clear that they're not really "results", but they're candidates, ready for further refinement by the language server. Hopefully, this nomenclature is more clear.

### DIFF
--- a/apps/remote_control/test/fixtures/dependency/lib/dependency/structs.ex
+++ b/apps/remote_control/test/fixtures/dependency/lib/dependency/structs.ex
@@ -1,0 +1,3 @@
+defmodule Dependency.Structs.Service do
+  defstruct [:name, :host, :port]
+end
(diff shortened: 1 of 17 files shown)


apps/remote_control/test/fixtures/dependency/lib/dependency/structs.ex
```diff
@@ -0,0 +1,3 @@
+defmodule Dependency.Structs.Service do
+  defstruct [:name, :host, :port]
```
scottming, line 2: What is this for?

scottming: just for testing?

scottming (approved):

apps/remote_control/test/fixtures/dependency/lib/dependency/structs.ex
scohen: Yep, just for testing.

PR #220, opened 2023-06-14
Bump erlang to 25.3.2

bdanklin:
Per discussion [here](https://elixirforum.com/t/how-to-make-a-release-for-a-system-with-erlang-installed/35241/3) the erlang version should be identical in both the extension and lexical when building yourself.

This PR brings the erlang version in line with lexical-lsp/vscode-lexical repo.

```
=erl_crash_dump:0.5
Slogan: init terminating in do_boot ({load_failed,[lists,logger,logger_simple_h,logger_config,supervisor,logger_server,logger_filters,filename,heart,gen,error_logger,gen_server,code_server,file,ets,erl_parse,erl_eval,code,application_master,application_controller,application,error_handler,gen_event,file_io_server,file_server,erl_lint,proc_lib,logger_backend,kernel]})
```

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,4 +1,4 @@
-erlang 25.2.1
+erlang 25.3.2
 elixir 1.14.3-otp-25
 nodejs 12.16.3
 yarn 1.22.4


.tool-versions
```diff
@@ -1,4 +1,4 @@
-erlang 25.2.1
+erlang 25.3.2
```
scottming, line 1: Why not use the latest erlang version: `25.3.2.2`?

scottming (approved):

.tool-versions
bdanklin: a) I was considering a future contribution to the ci which includes parsing the .tool-versions file for elixir/otp versions and I previously had issues with using the extra erlang patch version when trying to use the tool versions in ci.
```
        uses: erlef/setup-beam@v1
        with:
          version-file: .tool-versions
          version-type: strict
``` 

b) I didn't want to make multiple PRs to bring the extension repo inline as well.

scottming: ok

scohen:
Can we just bundle erts instead?
@bdanklin can you try switching to `include_erts: true` in mix.exs?

Blond11516:
@bdanklin What exactly are you referencing with "brings the erlang version in line with lexical-lsp/vscode-lexical repo"?

vscode-lexical has nothing to do with the versions of either Elixir or Erlang. The version is dictated by the build that gets started (either one from the releases in the lexical repo or your own if configured so). Also it wouldn't matter with `include_erts: true` as scohen suggested.

bdanklin:
@Blond11516 I see my mistake. I used the output of asdf to see the erlang version in each repo. Therefore the erlang version reported in the vscode extension repo was actually my global version as it is not present in the tool versions - hence the mismatch.

@scohen I agree that `include_erts: true` should be set and I believe it should fix the issue properly, Ill be happy to test it and follow up with another contribution when I get some more free time. 

In that case the reason for this PR is solely to bump the erlang version. 

I'll leave it up to you guys if you want to merge it for what it is or close it and wait for something more concrete.

Cheers

scohen:
Yeah, not going to merge this; I'm going to progress on figuring out some way to build this behemoth so we can run on later erlang / elixir and older projects can use their versions.

PR #222, opened 2023-06-15
Plugin packaging

scohen:
Two commits here, one adds a lot of docs and tooling to support releasing things, the other changes the function names of diagnostics to `diagnose` rather than `handle`


projects/lexical_plugin/lib/lexical/plugin/supervisor.ex
```diff
@@ -9,6 +9,6 @@ defmodule Lexical.Plugin.Supervisor do
   end
 
   def async(plugin_module, subject) do
-    Task.Supervisor.async_nolink(__MODULE__, plugin_module, :handle, [subject])
+    Task.Supervisor.async_nolink(__MODULE__, plugin_module, :diagnose, [subject])
```
scottming, line 12: if the plugin type is not diagnostics in the future, will it use diagnose too?

scottming:
LGTM, but I do not have much experience with plugin packing, so maybe other teammates can add some comments.

scottming (approved):

projects/lexical_plugin/lib/lexical/plugin/supervisor.ex
scohen: No, different plugins will hav different function names.

projects/lexical_plugin/lib/lexical/plugin/v1/diagnostic.ex
```diff
@@ -0,0 +1,93 @@
...
+  the deadline is much longer than it is when `diagnose` is called with a `Lexical.Document`. This is because
+  analyzing a project should take a lot longer than analyzing a single document. Currently, Lexical sets
+  a deadline of around a second on project-level diagnostics and a tens-of-milliseconds deadline on document
+  plugins.
+
+
+  ## Errors and Timeouts
+  Plugins are very sensitive to errors and are disabled by lexical if they cause too many. At present,
+  a disabled plugin can only be re-enabled by restarting lexical, so ensure that your plugin doesn't crash,
+  and if you don't have errors, return `{:ok, []}` rather than some error response.
```
Blond11516, line 24:
```suggestion
  and if you don't have diagnostics, return `{:ok, []}` rather than some error response.
```

projects/lexical_shared/lib/lexical/struct_access.ex
```diff
@@ -1,14 +1,20 @@
 defmodule Lexical.StructAccess do
+  @moduledoc """
+  Allows structs to easily adopt the `Access` behavoiur.
```
Blond11516, line 3:
```suggestion
  Allows structs to easily adopt the `Access` behaviour.
```

projects/lexical_credo/README.md
```diff
@@ -1,21 +1,36 @@
...
+```
+
+Once you see that, you should start seeing Credo diagnostics in your editor.
+
+
+## Notes on the plugin
 
+The plugin runs Credo whenever you type, and most of your settings will be respected... for the most part.
+Because Credo is designed to work with files on the disk, and the code you're editing isn't the same as
+the code that's on the disk, we sending the file's contents to Credo via standard output. However, when
```
Blond11516, line 33:
```suggestion
the code that's on the disk, we are sending the file's contents to Credo via standard output. However, when
```

projects/lexical_credo/README.md
```diff
@@ -1,21 +1,36 @@
...
+we do this, Credo doesn't have a provision to provide the file name and and the filename is lost.
```
Blond11516, line 34:
```suggestion
we do this, Credo doesn't have a provision to provide the file name and the filename is lost.
```

Blond11516 (approved):

scohen:
> I do not have much experience with plugin packaging

Me either. I'll be honest, I'm not super happy with how the packaging stuff turned out, but I think we can improve it later.

PR #223, opened 2023-06-15
Added thoughts about code intelligence and indexing

scohen:


pages/intelligence.md
```diff
@@ -0,0 +1,137 @@
...
+
+Indexing the entire project, its dependencies, elixir and erlang when
+the language server starts is likely going to be too performance
+intensive to do (to wit, rewriting atoms during namespacing takes 15
+seconds on our moderately sized project), so we will need some sort of
+on-disk representation of the index.
+
+Such a representation will need to take the following into account:
+
+  * Branches can while the language server is running, invalidating large portions of the cache
```
Blond11516, line 59:
```suggestion
  * Branches can change while the language server is running, invalidating large portions of the cache
```

pages/intelligence.md
```diff
@@ -0,0 +1,137 @@
...
+
+```elixir
+
+@type source :: Lexical.uri()
+@entry_type :: :otp | :elixir | :dependency | :project
+
+@type entry %{
+  type: entry_type(),
+  source: source(),
+  entry_version: String.t(),
```
Blond11516, line 79: I see there's a different version format for each entry type. Should we parse those to structs and use a union type here?

pages/intelligence.md
```diff
@@ -0,0 +1,137 @@
...
+}
+```
+
+### Invalidation
+When lexical starts the first time, it will need to index the
+following entities in the following order:
+
+  1. The elixir release
+  2. All project dependencies
+  3. The project's source code and built artifacts
```
Blond11516, line 91: Are you referencing something like the compiled beam files of the project itself? Why de we need to index that on top of source files?

One advantage I could see is that macros are already epanded, but I've no idea if that's a valid advantage or how we would map findings for artifacts back to source code.

Blond11516 (approved):
Gave the doc a quick read and it looks pretty clear to me. I don't have a lot of context on much of this, but I think the idea level ideas are sound.

pages/intelligence.md
```diff
@@ -0,0 +1,137 @@
...
+search both beam files and source code to discover relevant parts and
+locations of elixir and erlang code, index them an make them
+searchable. This is a large task that will not be accomplished in one
+commit, and this document will help to lay out the scope of the work,
+as well as collect thoughts and drive the implementation.
+
+Code intelligence will be oragnized into the following discrete areas:
+
+* **Discovery** - Finding points of interest in the project's codebase, dependencies and elixir core modules
+* **Caching** - Writing the discovered items to disk and ensuring the index is up to date with the state of the project at *all times*. The goal here is to never have to delete the `.lexical` directory.
```
Sleepful, line 20: I wonder if we are going to need git hooks for situations where `git checkout <branch>`.
Probably better than "watching" _all_ the files for changes.
Or perhaps we "watch" only the `.git` directory for changes?

Sleepful: Hm another situation is that editing files in the editor is something that can be easily communicated to lexical, but file edits could come from other places like `rm <file>` or an editor without lsp configuration.

pages/intelligence.md
```diff
@@ -0,0 +1,137 @@
...
+* **Indexing** - Storing the discovered artifacts in a local in-memory index that enable fast searching.
+* **Searching** ideally using something like the [sublime text's fuzzy search](https://www.forrestthewoods.com/blog/reverse_engineering_sublime_texts_fuzzy_match/), but we definitely need some type of fuzzy search.
```
Sleepful, line 22: so IME the LS does not have to worry about the mechanism for search, LS instead worries about answering queries defined by the LSP. It is up to the client _how_ to search through the answers from the server, e.g. fuzzy search through the symbols in a document, or grep through them.

Are you thinking of some specific `search` mechanism _not_ defined in the LSP?

pages/intelligence.md
```diff
@@ -0,0 +1,137 @@
...
+* **Recognition** For several of the above examples (go to definition, find references), we need to be able to recognize what our cursor is hovering over. Is it a module? An atom? a variable? etc.
```
Sleepful, line 23: declaration or a reference for example :)

pages/intelligence.md
```diff
@@ -0,0 +1,137 @@
...
+ * Types
+ * Variables
+ * Atoms?
+ * Additional items, defined in plugins
+
+While investigating the above items, we will need to take into account
+the difference between usages and definitions. This will be pretty
+clear for most items (though aliasing and importing might muddy the
+waters somewhat), but atoms will be difficult, since there's no
+defintion to speak of, just a usage.
```
Sleepful, line 46: yes, atoms are like an integer or any other primitive, the integer value might be used _somewhere_ as an indication of state, but it is just an integer like any other.

pages/intelligence.md
```diff
@@ -0,0 +1,137 @@
...
+
+# Caching
+> There are only two hard problems in computer science: Naming things, cache invalidation and off by one errors.
```
Sleepful, line 49: 😆

pages/intelligence.md
```diff
@@ -0,0 +1,137 @@
...
+Such a representation will need to take the following into account:
+
+  * Branches can while the language server is running, invalidating large portions of the cache
+  * Project level search and replace can affect large numbers of files at once
+  * The versions of elixir and erlang can change via `asdf` whenever the user wants
+
+Basicallly, the project's assets can change at any time, and it's not
+guaranteed that the language server will be able to detect all of the
+changes, so the language server needs to treat this cache with some
+skepticism and constantly "check its work".
```
Sleepful, line 66: would it be too crazy to have a CLI command that the user can run that tells the server to re_validate stuff? 

At least for the time being, an alternative to `rm -rf .lexical`

Not sure if a good idea, but I would enjoy not using the `rm`

Sleepful: 2. would we need to do like a dependency tree so we can isolate invalidations to _some_ part of the code, and avoid recompiling the whole thing? 

thinking about performance

pages/intelligence.md
```diff
@@ -0,0 +1,137 @@
...
+
+Entries in the cache should be replaced and not updated, and we should
+have a mechanism to sweep old entries from the cache (for example, if
+a dependecy version is old and hasn't been updated in two weeks, we
+should remove it). This invalidation is going to be one of the tricky
+parts of the cache, and we should be aggressive with it, as
+re-indexing things is likely to be less bad than having an
+ever-growing cache on disk.  We need to think more about this and come
+up with something that strikes a balance between efficiency and
+on-disk storage.
```
Sleepful, line 111: This paragraph confuses me. 

I guess the entire section is a little bit confusing. 

1. What is an `entity`?
2. Is an `entity` the same as an `entry`?
3. Is the cache going to hold `entities`/`entries` for the entire codebase or only some parts of the codebase?

pages/intelligence.md
```diff
@@ -0,0 +1,137 @@
...
+
+# Indexing
+Indexing and searching go hand-in-hand, you need to make an index in a
+format that's useful, and as such, we'll approach this at a later
+date. In the interim, we can probably get away with a linear search;
+there aren't _that many_ things that we're indexing and elixir is kind
+of fast for this kind of thing.
```
Sleepful, line 118: the index is going to go on disk, right?

https://www.erlang.org/doc/man/erlang.html#term_to_binary-1 for now?

pages/intelligence.md
```diff
@@ -0,0 +1,137 @@
...
+When we do things like "go to definition" or "find references" we need
+to understand the context beneath the cursor to determine _what_ to
+look for. However, this can be somewhat challenging, because elixir
+has ambiguity in its syntax. For example:
+
+```elixir
+foo
+```
+
+can be a variable or a function call. Our recognition engine needs to
```
Sleepful, line 130: > can be a variable or a function call

thanks Ruby syntax? :p 

Funny given that now parenthesis are more or less enforced by `mix format` for function calls

pages/intelligence.md
```diff
@@ -0,0 +1,137 @@
+# Code intelligence
+
+We'd like to replace elixir_sense with our own implmentation for the following bits
```
Tuxified, line 3:
```suggestion
We'd like to replace elixir_sense with our own implementation for the following bits
```

pages/intelligence.md
```diff
@@ -0,0 +1,137 @@
...
+of code intelligence.
+
+ * Go to definition
+ * Find references
+ * Refactoring
+
+In order to have this level of intelligence of elixir code, we need to
+search both beam files and source code to discover relevant parts and
+locations of elixir and erlang code, index them an make them
```
Tuxified, line 12:
```suggestion
locations of elixir and erlang code, index them and make them
```

pages/intelligence.md
```diff
@@ -0,0 +1,137 @@
...
+searchable. This is a large task that will not be accomplished in one
+commit, and this document will help to lay out the scope of the work,
+as well as collect thoughts and drive the implementation.
+
+Code intelligence will be oragnized into the following discrete areas:
```
Tuxified, line 17:
```suggestion
Code intelligence will be organized into the following discrete areas:
```

pages/intelligence.md
```diff
@@ -0,0 +1,137 @@
...
+
+* **Discovery** - Finding points of interest in the project's codebase, dependencies and elixir core modules
+* **Caching** - Writing the discovered items to disk and ensuring the index is up to date with the state of the project at *all times*. The goal here is to never have to delete the `.lexical` directory.
+* **Indexing** - Storing the discovered artifacts in a local in-memory index that enable fast searching.
```
Tuxified, line 21:
```suggestion
* **Indexing** - Storing the discovered artifacts in a local in-memory index that enables fast searching.
```

pages/intelligence.md
Tuxified: I think it would also be useful to have other types of searches available. Given a line of code, to which function/module does it belong? Which modules use this macro etc.

pages/intelligence.md
Tuxified: Definitely useful, but I wonder if https://hexdocs.pm/elixir/Code.Fragment.html#surround_context/3 already covers enough of what we need?

pages/intelligence.md
Tuxified: I agree we want to index all 3 "types" of entities, but I'd like to suggest the reverse order so we are able to give suggestions etc sooner to the user (I'm guessing folks will look where a function from project code is defined rather than Elixirs function?). Perhaps we could also speed things up if we're able to ship the index for an Elixir release?

Tuxified (approved):
I agree with most (if not all) of this doc proposes. As mentioned, this is not a small task that can be done in 1 commit (or even 1 PR), so my question is: where do we start? What part can we build to get some value out of it?

pages/intelligence.md
scohen: Keep in mind that indexing the first two things will happen much less frequently than the third, so I don't think the order will be super relevant to how quickly we serve the index.

pages/intelligence.md
scohen: > I wonder if we are going to need git hooks

We won't need "git hooks", but we will need to pay attention to file changes (we already have support for them right now) and re-index files that change due to checkouts.

> file edits could come from other places like
Yes, that's one of the reasons why having reindexing jobs that pay attention to filesystem changes is so important.

pages/intelligence.md
scohen: Not really, for something like: `foo`, there is no "surround context" to go off of.

pages/intelligence.md
scohen: I'd rather lexical implement some kind of background check that would keep itself up to date.

pages/intelligence.md
scohen: An entity in the cache is the thing actually being cached
An entry in the cache is an entity + some metadata that helps with eviction. 
The cache will cache everything that we index, which is the entire codebase + elixir + your dependencies.

pages/intelligence.md
scohen: I'm wondering if we should just define the things as different structs

pages/intelligence.md
scohen: Yes, I was considering using something like `disk_log` or `dets` for this. Both use `term_to_binary` under the hood

pages/intelligence.md
scohen: well, not for DSLs, which we need to support. Things get really hairy there.

scohen:
> I agree with most (if not all) of this doc proposes. As mentioned, this is not a small task that can be done in 1 commit (or even 1 PR), so my question is: where do we start? What part can we build to get some value out of it?

Yeah, that's a super good question. I'm thinking we need to write down all the things we need to look for, then start with `find references`. 

We can start small; finding references to modules (or functions, or variables, or whatever). We start there because we don't have this feature already, and we can add to it incrementally. Adding incrementally allows us to test out a bunch of other things, like reindexing, and it also allows us to build scanners that can detect all of the things we need to detect.

pages/intelligence.md
scohen: >  It is up to the client how to search through the answers from the server, e.g. fuzzy search through the symbols in a document,

Well, sorta. If the language server doesn't offer any suggestions, then the client can't filter them further. Consider what would happen if we did a strict equality and the client searched for `ENum`, and we returned `[]` for completions.

pages/intelligence.md
scohen: (i really, really want to use built-in functionality by the way)

pages/intelligence.md
scohen: right, but the wrinkle is that modules are atoms too

pages/intelligence.md
scohen: > I think it would also be useful to have other types of searches available.

Yes, definitely, consider the above a starting point.

Sleepful: Cool! Perhaps it would be useful to start enumerating in a doc the queries that we support and mark them as whether they are queries defined by the LSP or something more custom

pages/intelligence.md
Sleepful: Right! So they need to be treated like... Enums?  
A dictionary of values that are indexed as they are "found throughout the code". Perhaps it gets a little bit tricky to remove indexed atoms once they are deleted from the source.

pages/intelligence.md
Sleepful: I guess it should not be too difficult since this is very standard for other tools, like compilers.
For nodejs I remember using this one to watch for files https://github.com/remy/nodemon
For unix there is `watch(1)`, but it sounds inefficient if you watch a whole directory: https://linux.die.net/man/1/watch
I guess it is only necessary to query the dates of file modification for existing files, and querying the directories for new files.

pages/intelligence.md
Sleepful: Got it.
So entities need to be indexed, all entities are indexed.
But when it comes caching, only some entities become and entry, based on... not sure.

Correct?

pages/intelligence.md
Sleepful: True. I would be fine with kicking that problem down the road until everything else works great, and only at that point start thinking about abstractions that work for the DSL problem.

pages/intelligence.md
scohen: The language server already has this: in [didChangeWatchedFiles](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles)

scottming:
> I agree with most (if not all) of this doc proposes. As mentioned, this is not a small task that can be done in 1 commit (or even 1 PR), so my question is: where do we start? What part can we build to get some value out of it?

This is also a big concern for me.

PR #224, opened 2023-06-18
Fix the issue of completing the current module in NeoVim and VScode.

### LINKED ISSUE #191: The completions for `%__MODULE__` in vscode and neovim is incorrect.
image

type `%__|` then ctrl+space

```elixir
18:03:26.005 [info] Local completions are [%Lexical.RemoteControl.Completion.Result.Macro{argument_names: [], arity: 0, name: "__CALLER__", origin: "Kernel.SpecialForms", type: :macro, visibility: :public, spec: "", metadata: %{}}, %Lexical.RemoteControl.Completion.Result.Macro{argument_names: [], arity: 0, name: "__DIR__", origin: "Kernel.SpecialForms", type: :macro, visibility: :public, spec: "", metadata: %{}}, %Lexical.RemoteControl.Completion.Result.Macro{argument_names: [], arity: 0, name: "__ENV__", origin: "Kernel.SpecialForms", type: :macro, visibility: :public, spec: "", metadata: %{}}, %Lexical.RemoteControl.Completion.Result.Macro{argument_names: [], arity: 0, name: "__MODULE__", origin: "Kernel.SpecialForms", type: :macro, visibility: :public, spec: "", metadata: %{}}, %Lexical.RemoteControl.Completion.Result.Macro{argument_names: [], arity: 0, name: "__STACKTRACE__", origin: "Kernel.SpecialForms", type: :macro, visibility: :public, spec: "", metadata: %{since: "1.7.0"}}, %Lexical.RemoteControl.Completion.Result.Macro{argument_names: ["term"], arity: 1, name: "__aliases__", origin: "Kernel.SpecialForms", type: :macro, visibility: :public, spec: "", metadata: %{}}, %Lexical.RemoteControl.Completion.Result.Macro{argument_names: ["term"], arity: 1, name: "__block__", origin: "Kernel.SpecialForms", type: :macro, visibility: :public, spec: "", metadata: %{}}]
18:03:26.006 [warn] Emitting completions: [#Protocol.Types.Completion.Item<[detail: "", insert_text: "__CALLER__", kind: :constant, label: "__CALLER__", sort_text: "%__CALLER__"]>, #Protocol.Types.Completion.Item<[detail: "", insert_text: "__DIR__", kind: :constant, label: "__DIR__", sort_text: "%__DIR__"]>, #Protocol.Types.Completion.Item<[detail: "", insert_text: "__ENV__", kind: :constant, label: "__ENV__", sort_text: "%__ENV__"]>, #Protocol.Types.Completion.Item<[detail: "", insert_text: "__MODULE__", kind: :constant, label: "__MODULE__", sort_text: "%__MODULE__"]>, #Protocol.Types.Completion.Item<[detail: "", insert_text: "__STACKTRACE__", kind: :constant, label: "__STACKTRACE__", sort_text: "%__STACKTRACE__"]>]
```

It seems like not using the `TextEdit`

and in neovim, there is no `%__MODULE__{|}` completions.

image

```lua
[DEBUG][2023-05-29 18:07:49] .../vim/lsp/rpc.lua:284	"rpc.send"	{
  id = 3,
  jsonrpc = "2.0",
  method = "textDocument/completion",
  params = {
    context = {
      triggerKind = 1
    },
    position = {
      character = 14,
      line = 9
    },
    textDocument = {
      uri = "file:///Users/scottming/Code/dummy/lib/dummy.ex"
    }
  }
}
[DEBUG][2023-05-29 18:07:51] .../vim/lsp/rpc.lua:387	"rpc.receive"	{
  id = 3,
  jsonrpc = "2.0",
  result = { {
      detail = "",
      insertText = "__CALLER__",
      kind = 21,
      label = "__CALLER__",
      sortText = "%__CALLER__"
    }, {
      detail = "",
      insertText = "__DIR__",
      kind = 21,
      label = "__DIR__",
      sortText = "%__DIR__"
    }, {
      detail = "",
      insertText = "__ENV__",
      kind = 21,
      label = "__ENV__",
      sortText = "%__ENV__"
    }, {
      detail = "",
      insertText = "__MODULE__",
      kind = 21,
      label = "__MODULE__",
      sortText = "%__MODULE__"
    }, {
      detail = "",
      insertText = "__STACKTRACE__",
      kind = 21,
      label = "__STACKTRACE__",
      sortText = "%__STACKTRACE__"
    } }
}
```

scottming:
I believe that only in the context of functions and types will we use the structure of the current module. Therefore, I used these two methods to determine it. In addition, we should keep the completion behavior of structure consistent, so I adopted TextEdit for `%__MODULE__` too.

Fixes #191

### DIFF
--- a/apps/server/lib/lexical/server/code_intelligence/completion/env.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/env.ex
@@ -102,10 +102,19 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
       {:ok, _line, {:struct, _}} ->
         true
 
-      {:ok, line, {:local_or_var, [?_, ?_ | rest]}} ->
+      {:ok, _line, {:local_or_var, [?_ | _rest]}} ->
         # a reference to `%__MODULE`, often in a function head, as in
         # def foo(%__)
-        String.starts_with?("MODULE", List.to_string(rest)) and String.contains?(line, "%__")
+
+        starts_with_percent? =
+          env
+          |> prefix_tokens(2)
+          |> Enum.any?(fn
+            {:percent, :%, _} -> true
+            _ -> false
+          end)
+
+        starts_with_percent? and (ancestor_is_def?(env) or ancestor_is_type?(env))
 
       _ ->
         false
@@ -516,6 +525,30 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
       {:error, _, _, _, tokens} ->
         {:ok, tokens, ''}
     end
+  end
+
+  defp ancestor_is_def?(env) do
+    env
+    |> cursor_path()
+    |> Enum.any?(fn
+      {:def, _, _} ->
+        true
+
+      {:defp, _, _} ->
+        true
+
+      _ ->
+        false
+    end)
+  end
+
+  defp ancestor_is_type?(env) do
+    env
+    |> cursor_path()
+    |> Enum.any?(fn
+      {:type, _, _} -> true
+      _ -> false
+    end)
   end
 
   defp cursor_path(env) do
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
@@ -3,6 +3,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
   alias Lexical.Server.CodeIntelligence.Completion.Env
   alias Lexical.Server.CodeIntelligence.Completion.Translatable
   alias Lexical.Server.CodeIntelligence.Completion.Translations.Callable
+  alias Lexical.Server.CodeIntelligence.Completion.Translations.Struct
 
   use Translatable.Impl, for: Candidate.Macro
 
@@ -440,12 +441,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
 
   def translate(%Candidate.Macro{name: "__MODULE__"} = macro, builder, env) do
     if Env.in_context?(env, :struct_reference) do
-      env
-      |> builder.snippet("%__MODULE__{$1}",
-        detail: "%__MODULE__{}",
-        label: "%__MODULE__{}",
-        kind: :struct
-      )
+      Struct.completion(env, builder, macro.name, macro.name)
     else
       env
       |> builder.plain_text("__MODULE__",
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/struct.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/struct.ex
@@ -69,19 +69,23 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Struct do
           prefix_end
 
         {:struct, typed_module_name} ->
-          edit_begin =
-            case left_offset_of(typed_module_name, ?.) do
-              {:ok, offset} ->
-                env.position.character - offset
+          beginning_of_edit(env, typed_module_name)
 
-              :error ->
-                env.position.character - length(typed_module_name)
-            end
-
-          edit_begin
+        {:local_or_var, [?_ | _rest] = typed} ->
+          beginning_of_edit(env, typed)
       end
 
     {edit_begin, env.position.character}
+  end
+
+  defp beginning_of_edit(env, typed_module_name) do
+    case left_offset_of(typed_module_name, ?.) do
+      {:ok, offset} ->
+        env.position.character - offset
+
+      :error ->
+        env.position.character - length(typed_module_name)
+    end
   end
 
   defp left_offset_of(string, character) do
--- a/apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
+++ b/apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
@@ -365,8 +365,18 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
     end
 
     test "is true if a module reference starts in function arguments" do
-      env = new_env("def my_function(%__|)")
+      env = new_env("def my_function(%_|)")
       assert in_context?(env, :struct_reference)
+    end
+
+    test "is ture if a module reference start in a t type spec" do
+      env = new_env("@type t :: %_|")
+      assert in_context?(env, :struct_reference)
+    end
+
+    test "is false if module reference not starts with %" do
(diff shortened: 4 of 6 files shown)


apps/server/lib/lexical/server/code_intelligence/completion/env.ex
```diff
@@ -102,10 +102,10 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
       {:ok, _line, {:struct, _}} ->
         true
 
-      {:ok, line, {:local_or_var, [?_, ?_ | rest]}} ->
+      {:ok, line, {:local_or_var, [?_ | _rest]}} ->
         # a reference to `%__MODULE`, often in a function head, as in
         # def foo(%__)
-        String.starts_with?("MODULE", List.to_string(rest)) and String.contains?(line, "%__")
+        String.contains?(line, "%_") and (ancestor_is_def?(env) or ancestor_is_type?(env))
```
scohen, line 108: does this work with 
```elixir
def something(%MyThing{}, __MODULE__)
```
?

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -365,7 +365,12 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
     end
 
     test "is true if a module reference starts in function arguments" do
-      env = new_env("def my_function(%__|)")
+      env = new_env("def my_function(%_|)")
+      assert in_context?(env, :struct_reference)
+    end
```
scohen, line 370: you need to exercise this more; Think about the ways in which the code you created can fail. For example, you're using a `String.contains?` on the entire line in `env.ex`. That seems really off to me, since you don't seem  to care if the percent comes before _or_ after the cursor.

apps/server/lib/lexical/server/code_intelligence/completion/env.ex
```diff
@@ -518,6 +518,30 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
     end
   end
 
+  defp ancestor_is_def?(env) do
+    env
+    |> cursor_path()
+    |> Enum.find_value(false, fn
```
scohen, line 524: why use `find_value` here and not `any?` I think `any?` is clearer

apps/server/lib/lexical/server/code_intelligence/completion/env.ex
```diff
@@ -518,6 +518,30 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Env do
...
+        false
+    end)
+  end
+
+  defp ancestor_is_type?(env) do
+    env
+    |> cursor_path()
+    |> Enum.find_value(false, fn
+      {:type, _, _} -> true
+      _ -> false
```
scohen, line 541: same here. `any?` is clearer
```suggestion
    |> Enum.any?(fn
      {:type, _, _} -> true
      _ -> false     
```

apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
```diff
@@ -440,12 +441,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
 
   def translate(%Candidate.Macro{name: "__MODULE__"} = macro, builder, env) do
     if Env.in_context?(env, :struct_reference) do
-      env
-      |> builder.snippet("%__MODULE__{$1}",
-        detail: "%__MODULE__{}",
-        label: "%__MODULE__{}",
-        kind: :struct
-      )
+      Struct.completion(env, builder, "__MODULE__", "__MODULE__")
```
scohen, line 444:
```suggestion
      Struct.completion(env, builder, macro.name, macro.name)
```

apps/server/lib/lexical/server/code_intelligence/completion/translations/struct.ex
```diff
@@ -69,21 +69,25 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Struct do
...
-
-          edit_begin
+        {:local_or_var, [?_ | _rest] = typed} ->
+          edit_begin_of_struct(env, typed)
       end
 
     {edit_begin, env.position.character}
   end
 
+  defp edit_begin_of_struct(env, typed_module_name) do
```
scohen, line 81: so, this is always a struct, so I'd just call it `beginning_of_edit`

apps/server/lib/lexical/server/code_intelligence/completion/translations/struct.ex
```diff
@@ -69,21 +69,25 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Struct do
...
-              {:ok, offset} ->
-                env.position.character - offset
+          edit_begin_of_struct(env, typed_module_name)
 
-              :error ->
-                env.position.character - length(typed_module_name)
-            end
-
-          edit_begin
+        {:local_or_var, [?_ | _rest] = typed} ->
```
scohen, line 74: you sure a single underscore is enough to decide that this is `__MODULE__`?

scottming: Yes, and only the call from `Translations.Macro` will arrive here.

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
scottming: yeah, you're right, I changed to starts_with `percent(%)`

so the case is  `%_.*`, I think we do not need to care about the thing after the cursor.

apps/server/lib/lexical/server/code_intelligence/completion/translations/struct.ex
scohen: ok.

apps/server/test/lexical/server/code_intelligence/completion/env_test.exs
```diff
@@ -365,10 +365,20 @@ defmodule Lexical.Server.CodeIntelligence.Completion.EnvTest do
...
+      env = new_env("def my_function(%_|)")
       assert in_context?(env, :struct_reference)
     end
 
+    test "is ture if a module reference start in a t type spec" do
+      env = new_env("@type t :: %_|")
+      assert in_context?(env, :struct_reference)
+    end
+
+    test "if false if module reference not starts with %" do
```
scohen, line 377:
```suggestion
    test "is false if module reference follows a struct reference" do
```

also, there should be tests for the opposite, like:
```elixir
new_env("def something(my_thing|, %Struct{})")
```

scohen (approved):
Small suggestion / fix. Then you're good to go.

PR #225, opened 2023-06-19
Second attempt to make struct completion more consistent

scottming:
I think this is what we want.

Two things here:

### 1. Make module completions translation more consistent

The execution steps of new new algorithm are as follows:

1. If there is only one module within three generations, only that module will be returned, e.g.: `%Lexical.Document.Changes{}`
CleanShot 2023-07-29 at 14 32 57@2x

2. If there are more structs within three generations and the first layer of the struct has ancestor structs, the ancestor structs and more will be mixedly returned, but not the ancestor module.
CleanShot 2023-07-29 at 14 32 39@2x

3. If there are no ancestor structs, only more will be returned.
CleanShot 2023-07-29 at 14 33 22@2x

In addition, we do not return any structs beyond two generations due to `dots_count + 1`.

### 2. Make struct completions translation consistent as module translation

CleanShot 2023-07-29 at 21 25 01@2x

When our completion is triggered by a dot, ElixirSense still provides us with struct completion, so we need a consistent approach.

Fixes part of #182

### DIFF
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
@@ -5,9 +5,14 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
   alias Lexical.Server.CodeIntelligence.Completion.Translations
   alias Lexical.Server.Project.Intelligence
 
-  use Translatable.Impl, for: [Candidate.Module, Candidate.Behaviour, Candidate.Protocol]
+  use Translatable.Impl,
+    for: [Candidate.Module, Candidate.Struct, Candidate.Behaviour, Candidate.Protocol]
 
   def translate(%Candidate.Module{} = module, builder, %Env{} = env) do
+    do_translate(module, builder, env)
+  end
+
+  def translate(%Candidate.Struct{} = module, builder, %Env{} = env) do
     do_translate(module, builder, env)
   end
 
@@ -23,21 +28,43 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
     if Env.in_context?(env, :struct_reference) do
       complete_in_struct_reference(env, builder, module)
     else
-      detail = builder.fallback(module.summary, module.name)
+      detail = builder.fallback(module.summary, module.full_name)
       completion(env, builder, module.name, detail)
     end
   end
 
-  defp complete_in_struct_reference(%Env{} = env, builder, module) do
+  defp complete_in_struct_reference(%Env{} = env, builder, %Candidate.Struct{} = struct) do
+    immediate_descendent_structs =
+      immediate_descendent_struct_modules(env.project, struct.full_name)
+
+    if Enum.empty?(immediate_descendent_structs) do
+      Translations.Struct.completion(env, builder, struct.name, struct.full_name)
+    else
+      do_complete_in_struct_reference(env, builder, struct, immediate_descendent_structs)
+    end
+  end
+
+  defp complete_in_struct_reference(%Env{} = env, builder, %Candidate.Module{} = module) do
     immediate_descendent_structs =
       immediate_descendent_struct_modules(env.project, module.full_name)
 
-    structs_map = Map.new(immediate_descendent_structs, fn module -> {module, true} end)
-    dot_counts = module_dot_counts(module.full_name)
+    do_complete_in_struct_reference(env, builder, module, immediate_descendent_structs)
+  end
+
+  defp do_complete_in_struct_reference(
+         %Env{} = env,
+         builder,
+         module_or_struct,
+         immediate_descendent_structs
+       ) do
+    structs_map =
+      Map.new(immediate_descendent_structs, fn module_or_struct -> {module_or_struct, true} end)
+
+    dot_counts = module_dot_counts(module_or_struct.full_name)
     ancestors = ancestors(immediate_descendent_structs, dot_counts)
 
     Enum.flat_map(ancestors, fn ancestor ->
-      local_name = local_module_name(module.full_name, ancestor, module.name)
+      local_name = local_module_name(module_or_struct.full_name, ancestor, module_or_struct.name)
 
       more =
         env.project
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
@@ -20,40 +20,55 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
   end
 
   defp do_translate(%_{} = module, builder, %Env{} = env) do
-    struct_reference? = Env.in_context?(env, :struct_reference)
+    if Env.in_context?(env, :struct_reference) do
+      complete_in_struct_reference(env, builder, module)
+    else
+      detail = builder.fallback(module.summary, module.name)
+      completion(env, builder, module.name, detail)
+    end
+  end
 
-    defines_struct? = Intelligence.defines_struct?(env.project, module.full_name)
-
+  defp complete_in_struct_reference(%Env{} = env, builder, module) do
     immediate_descendent_structs =
       immediate_descendent_struct_modules(env.project, module.full_name)
 
-    defines_struct_in_descendents? =
-      immediate_descendent_defines_struct?(env.project, module.full_name) and
-        length(immediate_descendent_structs) > 1
+    structs_map = Map.new(immediate_descendent_structs, fn module -> {module, true} end)
+    dot_counts = module_dot_counts(module.full_name)
+    ancestors = ancestors(immediate_descendent_structs, dot_counts)
 
-    cond do
-      struct_reference? and defines_struct_in_descendents? and defines_struct? ->
-        more = length(immediate_descendent_structs) - 1
+    Enum.flat_map(ancestors, fn ancestor ->
+      local_name = local_module_name(module.full_name, ancestor, module.name)
 
+      more =
+        env.project
+        |> Intelligence.collect_struct_modules(ancestor, to: :infinity)
+        |> Enum.count()
+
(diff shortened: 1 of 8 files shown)


apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
```diff
@@ -20,42 +20,54 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
...
+    end)
+  end
 
-      struct_reference? and
-          immediate_descendent_defines_struct?(env.project, module.full_name) ->
-        Enum.map(immediate_descendent_structs, fn child_module_name ->
-          local_name = local_module_name(module.full_name, child_module_name)
-          Translations.Struct.completion(env, builder, local_name, child_module_name)
+  defp ancestors_and_more(modules) do
+    ancestors_and_descentents =
```
scohen, line 47:
```suggestion
    ancestors_and_descendants =
```

apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
```diff
@@ -20,42 +20,54 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
...
+      for ancestor <- modules do
+        Enum.reduce(modules, {ancestor, 0, []}, fn maybe_descendant,
+                                                   {ancestor, count, descendents} ->
+          if descendent?(maybe_descendant, ancestor) do
+            {ancestor, count + 1, [maybe_descendant | descendents]}
+          else
+            {ancestor, count, descendents}
+          end
         end)
+      end
```
scohen, line 57: a reduce inside a comprehension? Why not use the `reduce` feature of comprehensions to make this easier to reason about? Or a flat_map? Or `group_by`?

apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
```diff
@@ -20,42 +20,54 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
...
+
+    descendents =
+      Enum.flat_map(ancestors_and_descentents, fn {_, _, descendents} -> descendents end)
 
-      true ->
-        detail = builder.fallback(module.summary, module.name)
-        completion(env, builder, module.name, detail)
+    for {ancestor, count, _} <- ancestors_and_descentents, ancestor not in descendents do
```
scohen, line 62: I'm totally lost here; why did you do everything in "one" loop (it's really not one loop, but a reduce in a comprehension, which is m*n) ?

apps/server/lib/lexical/server/code_intelligence/completion/translations/struct.ex
```diff
@@ -20,6 +20,10 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Struct do
     end
   end
 
+  def completion(%Env{} = env, builder, module_name, full_name, 0) do
+    completion(env, builder, module_name, full_name)
+  end
+
   def completion(%Env{} = env, builder, module_name, full_name, more) when is_integer(more) do
```
scohen, line 27: why not just make `more \\ 0` ?

scohen (changes requested):

apps/server/lib/lexical/server/code_intelligence/completion/translations/struct.ex
scottming: we have a function below, it will cause conflicts.

`  def completion(%Env{} = env, builder, struct_name, full_name) do`

apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
scottming: Yes, you're right. That calculation was very silly and lacked thinking.

apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
```diff
@@ -20,42 +20,54 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
...
+      |> List.wrap()
+    end)
+  end
 
-      struct_reference? and
-          immediate_descendent_defines_struct?(env.project, module.full_name) ->
-        Enum.map(immediate_descendent_structs, fn child_module_name ->
-          local_name = local_module_name(module.full_name, child_module_name)
-          Translations.Struct.completion(env, builder, local_name, child_module_name)
+  defp ancestors_and_more(modules) do
```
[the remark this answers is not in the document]
scottming: I changed the code to use only one reduce.

apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
```diff
@@ -20,40 +20,50 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
...
+    |> group_by_ancestor()
+    |> Enum.map(fn {ancestor, descendants} ->
+      {ancestor, length(descendants)}
+    end)
+  end
 
-      struct_reference? and defines_struct? ->
-        Translations.Struct.completion(env, builder, module.name, module.full_name)
+  defp group_by_ancestor(modules) do
+    sorted = Enum.sort(modules)
```
scohen, line 55: I still do not understand this approach. It feels very, very complicated

Correct me if I'm wrong, but: Say you're completing `Lexical.|`.  Let's focus on the `Lexical.Document` completion. Don't you have the following  non-exclusive cases:
1. Lexical.Document contains a struct. 
2. Lexical.Document is the ancestor of some structs

If 1 is true, then include the completion for `%Lexical.Document{}` in the response. If it isn't, don't do anything.
If 2 is true, ask intelligence how many structs have `Lexical.Document` as an ancestor. This is as simple as:
```elixir
def descendant_struct_count(%Project{} =project,  module) do 
  project
  |> Intelligence.collect_struct_modules(module, to: :infinity)
  |> Enum.count()
end
```

I tested this approach with the following code, and it worked: `Intelligence.collect_struct_modules(project, Lexical.Document, to: 500)` which returned `["Lexical.Document.Lines", "Lexical.Document.Range", "Lexical.Document",
 "Lexical.Document.Location", "Lexical.Document.Changes",
 "Lexical.Document.Position", "Lexical.Document.Edit"]`. 


(this would mean adding handling of `:infinity` to Inteliigence, but it seems a vastly simpler and more reusable approach)

Then after you have the ancestor count, and it's > 0, add the ancestor completions.

projects/lexical_shared/lib/lexical/formats.ex
```diff
@@ -88,4 +88,16 @@ defmodule Lexical.Formats do
...
+  def plural(count, singular, plural) do
+    case count do
+      0 -> templatize(count, plural)
+      1 -> templatize(count, singular)
+      _n -> templatize(count, plural)
+    end
+  end
+
+  defp templatize(count, template) do
+    String.replace(template, "${count}", "#{count}")
```
scohen, line 101: Maybe do `Integer.to_string(count)`, as it would be a little clearer.

apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
scottming: The problem is that they have ancestral relationships themselves, and we cannot guarantee where the first ancestor will appear. It could be in the first layer, like `ProjectStruct`, or it could be in the last layer like `Project.Structs.User`.

so all I want is a map like this:

```elixir
  @modules [
    "Lexical.Document.Lines",
    "Lexical.Document.Range",
    "Lexical.Document",
    "Lexical.Server.State",
    "Lexical.Server.Configuration",
    "Lexical.Document.Location",
    "Lexical.Document.Changes",
    "Lexical.Document.Position",
    "Lexical.Document.Edit",
    "Lexical.Project"
  ]

  @expect %{
    "Lexical.Document" => [
      "Lexical.Document.Range",
      "Lexical.Document.Position",
      "Lexical.Document.Location",
      "Lexical.Document.Lines",
      "Lexical.Document.Edit",
      "Lexical.Document.Changes"
    ],
    "Lexical.Project" => [],
    "Lexical.Server.Configuration" => [],
    "Lexical.Server.State" => []
  }
  
 assert group_by_ancestor(@modules) == @expect

```

scottming:
Many people have told me that they really like the current module completion effects, so I'm not sure if our changes are the right approach. I'd like to close this PR for now and try again in a while.

scohen:
@scottming, can we continue this please? I liked the way this was turning out, and I think we can improve things. Asking people what they prefer is not really the best way to make a product. 

There's an old saying that if you asked people what they wanted in a car, they'd ask for a faster horse.

apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
```diff
@@ -20,40 +25,77 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
...
+    |> Enum.uniq()
+  end
 
-      true ->
-        detail = builder.fallback(module.summary, module.name)
-        completion(env, builder, module.name, detail)
-    end
+  defp module_dot_counts(module_name) do
+    module_name |> String.graphemes() |> Enum.count(&(&1 == "."))
   end
```
scohen, line 99:
```elixir
# this skips grapheme translations
defp count_periods(module_name), do: module_dot_counts(module_name, 0) do 

defp count_periods(<<>>, count), do: count
defp count_periods(<<".", rest::binary>>, count), do: module_dot_counts(rest, count + 1)
defp count_periods(<<_::utf8, rest::binary>>, count), do: module_dot_counts(rest, count)
```

scohen (approved):
Just a couple of small comments. Can you rebase this with latest? I'd like to try.

scohen:
@scottming is this ready to merge?

PR #227, opened 2023-06-20
Support projects having the same directory name as a dependency

scohen:
In order to use the `Mix.Project.in_project`, you need to give a name for the project, and we were defaulting to the atomized name of their project directory. The problem with this approach is if you named your project "jason" and it depends on "jason", then your project will conflict with the `jason` app when it comes time to run `deps.compile`. By naming your project after the module defined in `mix.exs`, this avoids the conflict, but still keeps the naming consistent (mix defaults to this project name).

We can't use the project's `:app` key, because umbrellas don't define it.

Another approach would be to give the project a unique ID when it's created and to use that as the name of the project.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/bootstrap.ex
+++ b/apps/remote_control/lib/lexical/remote_control/bootstrap.ex
@@ -12,7 +12,7 @@ defmodule Lexical.RemoteControl.Bootstrap do
 
   def init(%Project{} = project, listener_pid, remote_control_config) do
     true = Code.append_path(hex_path())
-    RemoteControl.set_project(project)
+
     RemoteControl.set_project_listener_pid(listener_pid)
     Application.put_all_env(remote_control: remote_control_config)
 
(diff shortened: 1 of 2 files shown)


scottming (approved):

soundmonster:
It looks like I'm still running into this issue when running the formatter. It happens to me on the latest `main` as of 2024/05/06. The setup is the following: umbrella app in a folder `foo` that hass a dependency also called `foo`. When trying to find the formatter settings, Mix loads the `.formatter.exs` from the umbrella but the mix project from the dependency. What then happens is a formatter error that says "Unknown dependency :x_y_z given to :import_deps in the formatter configuration. Make sure the dependency is listed in your mix.exs for environment :test and you have run \"mix deps.get\"".

The rest of the LSP seems to work as expected.

I'm trying to find a fix myself but am very open to guidance. It looks that we first set up the project on LSP side, then we pass it to RemoteControl that then enriches it with the module name. We then mostly read the project name from ETS in remote control, but sometimes we get it passed over from the LSP and in that case it's missing the module. Is this correct?

scohen:
> I'm trying to find a fix myself but am very open to guidance. It looks that we first set up the project on LSP side, then we pass it to RemoteControl that then enriches it with the module name. We then mostly read the project name from ETS in remote control, but sometimes we get it passed over from the LSP and in that case it's missing the module. Is this correct?

Yep, this is the bug.

PR #236, opened 2023-06-23
WIP: Alias module

scohen:
Built an alias module with submodules that can look at a bit of source code and determine what the aliases are at a given position

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/code_mod/ast/aliases.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_mod/ast/aliases.ex
@@ -1,0 +1,300 @@
+defmodule Lexical.RemoteControl.CodeMod.Ast.Aliases do
+  defmodule Alias do
+    defstruct [:from, :to]
+
+    def new(from, to) do
+      %__MODULE__{from: from, to: to}
+    end
+  end
+
+  defmodule Scope do
+    defstruct [:end_position, :current_module, :aliases, :on_exit]
+
+    def new(end_position, current_module, on_exit \\ &Function.identity/1) do
+      %__MODULE__{
+        aliases: %{},
+        current_module: current_module,
+        end_position: end_position,
+        on_exit: on_exit
+      }
+    end
+
+    def global do
+      new({:infinity, :infinity}, nil)
+    end
+
+    def ended?(%__MODULE__{end_position: {:infinity, :infinity}}, _) do
+      false
+    end
+
+    def ended?(%__MODULE__{} = scope, {line, column}) do
+      {end_line, end_column} = scope.end_position
+
+      if line == end_line do
+        column >= end_column
+      else
+        line > end_line
+      end
+    end
+
+    def put_alias(%__MODULE__{} = scope, _, :skip) do
+      scope
+    end
+
+    def put_alias(%__MODULE__{} = scope, from, to) do
+      [first | rest] = from
+
+      # This allows a pre-existing alias to define another alias like
+      # alias Foo.Bar.Baz
+      # alias Baz.Quux
+      from =
+        case scope.aliases do
+          %{^first => to_alias} -> Module.split(to_alias.from) ++ rest
+          _ -> from
+        end
+
+      new_alias = Alias.new(ensure_alias(scope, from), ensure_alias(scope, to))
+      %__MODULE__{scope | aliases: Map.put(scope.aliases, new_alias.to, new_alias)}
+    end
+
+    defp ensure_alias(%__MODULE__{} = scope, [:__MODULE__ | rest]) do
+      Module.concat([scope.current_module | rest])
+    end
+
+    defp ensure_alias(%__MODULE__{}, alias_list) when is_list(alias_list) do
+      Module.concat(alias_list)
+    end
+
+    defp ensure_alias(%__MODULE__{}, alias_atom) when is_atom(alias_atom) do
+      alias_atom
+    end
+  end
+
+  defmodule Reducer do
+    defstruct scopes: []
+
+    def new do
+      %__MODULE__{scopes: [Scope.global()]}
+    end
+
+    def update(%__MODULE__{} = reducer, elem) do
+      reducer
+      |> maybe_pop_scope(elem)
+      |> apply_ast(elem)
+    end
+
+    def current_module(%__MODULE__{} = reducer) do
+      current_scope(reducer).current_module
+    end
+
+    def aliases(%__MODULE__{} = reducer) do
+      reducer.scopes
+      |> Enum.reverse()
+      |> Enum.flat_map(&Map.to_list(&1.aliases))
+      |> Map.new(fn {k, %Alias{} = scope_alias} -> {k, scope_alias.from} end)
+      |> Map.put(:__MODULE__, current_module(reducer))
+    end
+
+    # defmodule MyModule do
+    defp apply_ast(
+           %__MODULE__{} = reducer,
+           {:defmodule, metadata, [{:__aliases__, _, module_name}, _block]}
+         ) do
+      module_alias =
+        case current_module(reducer) do
+          nil ->
+            module_name
+
+          current_module ->
+            Module.split(current_module) ++ module_name
+        end
+
+      current_module_alias =
+        case module_name do
+          [current] -> current
+          _ -> :skip
+        end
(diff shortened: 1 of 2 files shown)


apps/remote_control/lib/lexical/remote_control/code_mod/ast/aliases.ex
```diff
@@ -0,0 +1,197 @@
...
+          module_alias =
+            case Map.fetch(state.aliases, :__MODULE__) do
+              {:ok, alias_struct} ->
+                Module.split(alias_struct.source) ++ module_alias
+
+              _ ->
+                module_alias
+            end
+
+          put_alias(state, module_alias, :__MODULE__)
```
scottming, line 70:
```suggestion
      state
          |> put_alias(List.last(module_alias), module_alias)
          |> put_alias(:__MODULE__, module_alias)

```

scottming: It can solve the problem of implicit aliases.

apps/remote_control/lib/lexical/remote_control/code_mod/ast/aliases.ex
```diff
@@ -0,0 +1,197 @@
...
+    end
+
+    # An alias with a specified name: alias Foo.Bar, as: FooBar
+
+    defp do_update(
+           %__MODULE__{} = state,
+           {:alias, _metadata,
+            [{:__aliases__, _, aliased_module}, [as: {:__aliases__, _, dest_alias}]]}
+         ) do
+      source = ensure_alias(aliased_module)
```
scottming, line 94: Since we have already performed the ensure action in put_alias, I think there are many places where we don't need to ensure anymore.

apps/remote_control/lib/lexical/remote_control/code_mod/ast/aliases.ex
```diff
@@ -0,0 +1,197 @@
...
+        dest = aliased_module |> List.last() |> ensure_alias()
+        put_alias(state, source, dest)
+      end)
+    end
+
+    defp do_update(%__MODULE__{} = state, elem) do
+      %__MODULE__{state | last_elem: elem}
+    end
+
+    defp put_alias(%__MODULE__{} = state, from, to) do
```
scottming, line 116: I think switching the `from` and `to` is better, because dest is the key.

apps/remote_control/test/lexical/remote_control/code_mod/ast/aliases_test.exs
```diff
@@ -0,0 +1,233 @@
...
+              alias Foo.Something
+              |
+            end
+          end
+        ]
+        |> aliases_at_cursor()
+
+      assert aliases[:Grandparent] == Foo.Grandparent
+      assert aliases[:__MODULE__] == Parent.Child
+      assert aliases[:Something] == Foo.Something
```
scottming, line 129: I think you missed one: 

```elixir
assert aliases[:Child] = Parent.Child
```

apps/remote_control/test/lexical/remote_control/code_mod/ast/aliases_test.exs
```diff
@@ -0,0 +1,233 @@
...
+            defmodule Child do
+              alias Foo.Something
+              |
+            end
+
+          end
+        ]
+        |> aliases_at_cursor()
+
+      refute aliases[:Child]
```
scottming, line 147: but how to complete this case:

```elixir
defmodule Parent do
  defmodule Child do
    defstruct [:a]

    def foo(%Child{|}) do
    end
  end
end
```

I think `aliases[:Child]` should be `Parent.Child`

scottming:
```elxiir
    test "an alias expand current module" do
      aliases = ~q[
        defmodule TopLevel do
          alias __MODULE__.Foo|
        end
      ]
      |> aliases_at_cursor() |> dbg()
    end
```

apps/remote_control/lib/lexical/remote_control/code_mod/ast/aliases.ex
```diff
@@ -0,0 +1,199 @@
...
+           {:ok, current_column} <- Keyword.fetch(metadata, :column),
+           [current_scope | scopes] <- state.scopes,
+           true <- Scope.ended?(current_scope, {current_line, current_column}) do
+        aliases =
+          Map.reject(state.aliases, fn {_, alias_struct} ->
+            alias_struct.ref == current_scope.ref
+          end)
+
+        %__MODULE__{state | scopes: scopes, aliases: aliases}
+        |> current_scope.on_exit.()
```
scottming, line 151: don't quite understand why `Function.identity/1` here.

scohen: it's the default `on_exit` callback. It just returns what's been passed in, i.e. it does nothing.

apps/remote_control/test/lexical/remote_control/code_mod/ast/aliases_test.exs
scohen: no, you cannot do this; the `Child` module hasn't been defined yet, you're still inside it. You need to use `__MODULE__`.

scohen: Hmm, this behavior has changed, though I'd still caution against using the name of the module _in_ the module, it's not refactor friendly.

apps/remote_control/test/lexical/remote_control/code_mod/ast/aliases_test.exs
```diff
@@ -0,0 +1,318 @@
...
+          end
+        ]
+        |> aliases_at_cursor()
+
+      assert aliases[:Grandparent] == Foo.Grandparent
+      assert aliases[:__MODULE__] == Parent.Child
+      assert aliases[:Something] == Foo.Something
+    end
+
+    test "a nested module is aliased inside the module" do
```
scottming, line 210: since there is already a `describe` about nested module, I think we can delete this one right now.

scottming (approved):

scottming:
A test failed in the CI, please address it, then you're good.

scohen:
it is unrelated.

apps/remote_control/test/lexical/remote_control/code_mod/ast/aliases_test.exs
```diff
@@ -0,0 +1,367 @@
...
+          defmodule GrandParent.Parent.Child do
+           |
+          end
+        ]
+        |> aliases_at_cursor()
+
+      refute Map.has_key?(aliases, :Child)
+    end
+
+    test "with children get their parents name" do
```
Sleepful, line 200: very nice

apps/remote_control/lib/lexical/remote_control/code_mod/ast/aliases.ex
```diff
@@ -0,0 +1,300 @@
...
+    defp normalize_from([{:__MODULE__, _, _} | rest]) do
+      [:__MODULE__ | rest]
+    end
+
+    defp normalize_from(from) do
+      from
+    end
+  end
+
+  alias Future.Code
```
Sleepful, line 266: which version of elixir will this be released as eventually?

scottming:
1.15

Jose Vargas ***@***.***>于2023年6月29日 周四21:08写道：

Sleepful (approved):
very fancy handling of scope, a little bit difficult to wrap my head around it but the code looks good

scohen:
> a little bit difficult to wrap my head around 

It's just a stack. When we encounter a new scope, we push another scope onto the stack, when we leave, we pop the current scope. 
When we get aliases, we start at the lowest scope, put its aliases in there, and keep adding aliases until we get to the current scope. Newer aliases overwrite older ones that way.

apps/remote_control/lib/lexical/remote_control/code_mod/ast/aliases.ex
scohen: This is because José made changes for us, so we vendored the 1.5 `Code` and associated modules and namespaced them in the `Future` module. The nice thing about this is we can just `alias Future.Code` and you have a `Code` module.... from the future.

PR #237, opened 2023-06-23
[Buggy] Snippet translations

Sleepful:
I only made this branch to test locally, this code is not working as it should


Sleepful:
re https://github.com/lexical-lsp/lexical/pull/208

PR #239, opened 2023-06-24
Fix the issue of project name being too long

scottming:
Currently, we only support mix projects. Therefore, `Elixir` and `MixProject` do not provide any additional information. Removing them can make the UI look better and the project name clearer.

### Before

image

### After

image

### DIFF
--- a/projects/lexical_shared/lib/lexical/project.ex
+++ b/projects/lexical_shared/lib/lexical/project.ex
@@ -51,25 +51,22 @@ defmodule Lexical.Project do
   @spec name(t) :: String.t()
 
   def name(%__MODULE__{project_module: nil} = project) do
-    project
-    |> root_path()
-    |> Path.split()
-    |> List.last()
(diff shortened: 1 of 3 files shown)


projects/lexical_shared/lib/lexical/project.ex
```diff
@@ -58,7 +58,18 @@ defmodule Lexical.Project do
...
+    with ["Elixir" | tail] <- String.split(module_string, "."),
+         true <- ends_with_mix_project?(tail) do
+      tail |> List.delete_at(-1) |> Enum.join(".")
+    else
+      _ -> module_string
+    end
+  end
+
+  defp ends_with_mix_project?(module_path) do
+    List.last(module_path) == "MixProject"
```
scohen, line 72: rather than using the `name` for two things, why not just make a `display_name` function that gives a short name?

scohen (approved):

PR #245, opened 2023-06-29
GitHub Actions improvements

scohen:
Our github actions were taking a while. This PR does the following:
1. Splits up static analysis, dialyzer checks and tests into three jobs that all run in parallel.
2. Adds a matrix to the test task so we can test in different versions of elixir and erlang.

The result is that tests used to take around 7 minutes to run, and now they run in 3. More importantly, by splitting things up into three jobs, we get more errors all at once. You see the static analysis results in 20 seconds, and dialyzer in about 40 seconds. You also get to see failures for all three jobs separately. Prior, you would need to have successful static analysis and dialyzer before your tests would even run, which meant you'd often have to go back to CI multiple times before a successful run.


.github/workflows/elixir.yml
```diff
@@ -84,21 +78,121 @@ jobs:
...
           make dialyzer.all
 
+  test:
+    runs-on: ubuntu-latest
+    name: Test on OTP ${{matrix.otp}} / Elixir ${{matrix.elixir}}
+    strategy:
+      # Specify the OTP and Elixir versions to use when building
+      # and running the workflow steps.
+      matrix:
+        include:
```
scottming, line 153: I like this idea, Run in the highest and lowest supported version

scottming (approved):
Look great to me, but I do not have much experience in the CI area, so maybe @Blond11516 can take a look.

PR #247, opened 2023-06-29
Update .tool-versions to match GH Actions

Sleepful:
Updated to have the same versions as:

https://github.com/lexical-lsp/lexical/blob/main/.github/workflows/release.yml#L18-L19


scohen (commented):
No, this is not correct; we can't use 14/25 as the defaults. We need to support elixir 1.13 / erl 24.

PR #249, opened 2023-06-29
Fixed boundary issue

scohen:
The plugins diagnostic result convertible was being defined in a project that did not have access to the plugins repo. Moving it to server fixes the dependency issues.


scohen:
FYI, i'm merging this, it's blocking me.

PR #252, opened 2023-06-30
Fixing flaky tests

scohen:
The tests here were failing occasionally, and the coordinator test was actually not cleaning up after itself *and* was not testing the right things. This fixes all of those problems.


scohen:
Going to merge this, as it's very annoying to see flakes happen.

PR #254, opened 2023-07-04
Raise timeout on plugin document diagnose

doughsay:
On my machine, which is relatively powerful, the `lexical_credo` plugin takes ~150ms to complete it's diagnostics on a `%Document{}`. With the timeout being 50ms it never receives results. I tested on larger files and small files and it didn't seem to make a difference.  250ms might still be too low; let me know what you think.

### DIFF
--- a/apps/plugin_runner/lib/lexical/plugin/runner.ex
+++ b/apps/plugin_runner/lib/lexical/plugin/runner.ex
@@ -36,7 +36,7 @@ defmodule Lexical.Plugin.Runner do
 
   @doc false
   def diagnose(%Document{} = document, on_complete) do
-    Runner.Coordinator.run_all(document, :diagnostic, on_complete, 50)
+    Runner.Coordinator.run_all(document, :diagnostic, on_complete, 250)
   end
 
   @doc false


doughsay:
I tested this on a larger project and it was a lot worse, so it seems that the time credo requires is just very project specific. In the larger project I had to raise the project diagnose timeout to 4000ms and the document diagnose to 1000ms for it to work.

scottming:
250ms, are you sure? It is just one file, shouldn't be so slow.

doughsay:
> 250ms, are you sure? It is just one file, shouldn't be so slow.

Ok, wow, I did some more digging 😅: if any of your checks have options given to them to exclude specific files, e.g. I have this in my `.credo.exs`:

```elixir
{Credo.Check.Consistency.MultiAliasImportRequireUse, files: %{excluded: ["lib/my_app_web.ex"]}},
```

...then it completely tanks the performance.

So I guess this isn't so much a credo plugin issue as a credo itself issue...

this is _with_ the exclude option set:

```
2023-07-04T21:48:11.805275-07:00 info: Plugins found 77 results in 810 ms
2023-07-04T21:48:11.988752-07:00 info: Plugins found 0 results in 170 ms
```

this is _without_:

```
2023-07-04T21:48:21.975363-07:00 info: Plugins found 81 results in 700 ms
2023-07-04T21:48:22.000224-07:00 info: Plugins found 0 results in 15 ms
```

so it dramatically affects the document diagnose, but not as much the project diagnose.

scottming:
You can look at the file `Credo.Sources` to see if the `read_from_stdin` is broken because of the `excluded` and become a check for the whole project.

scohen:
My suspicion is they're doing a `File.wildcard` somewhere, even though they don't have to.

doughsay:
Sorry, just haven't had any time yet to look into why credo is slow in this specific case.  If you're not keen on the idea of raising the timeout here, then I'll just close this for now.  I'll report back in #editor-tooling if I find anything out about credo.

scohen:
@doughsay did you file a bug in credo? I think that might be the best course of action here.

doughsay:
@scohen issue opened; I tried digging into it a little further and my best guess is that it's triggering this condition here:

https://github.com/rrrene/credo/blob/master/lib/credo/check/runner.ex#L45

and that else branch is a lot heavier, since it's expanding glob patterns and reading lists of files from disk.

PR #260, opened 2023-07-07
Completion improvements

scohen:
Added completion helper for describe blocks and ignored syntax functions / macros.


scohen:
Merging this because it's a small-ish change and I like it.

PR #261, opened 2023-07-08
Support Elixir 1.15

### LINKED ISSUE #101: Enhance `undefined` function/variable error diagnostics
image

When a compilation error occurs indicating that a value does not exist, it can be more precise. There are three messages here, but none of them are accurate enough. It is not difficult to be precise about errors like this down to the column level and make it simpler and clearer.

```
{:error,
 [
   {"/Users/scottming/Code/lexical/apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs",
    240,
    "** (CompileError) test/lexical/remote_control/code_intelligence/definition_test.exs:240: undefined function referenced_uri/0 (expected Lexical.RemoteControl.CodeIntelligence.DefinitionTest to define such a function or for it to be imported, but no
 are available)\n\n"}
 ],
 [
   {"/Users/scottming/Code/lexical/apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs",
    238,
    "variable \"ctx\" is unused (if the variable is not meant to be used, prefix it with an underscore)"},
   {"/Users/scottming/Code/lexical/apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs",
    240,
    "variable \"referenced_uri\" does not exist and is being expanded to \"referenced_uri()\", please use parentheses to remove the ambiguity or change the variable name"}
 ]}

```

### LINKED ISSUE #243: Support Elixir 1.15
There are a _ton_ of failures under elixir 1.15, likely due to all the compiler work they did.

scottming:
Fixes #101
Closes #243 

Starting with Elixir 1.15, we don't need to capture io for compiling individual files or quoted anymore,
because with_diagnostics returns all compile-time diagnostics.

So I have implemented two ways to compile and process the compiled results.

You can see from the tests that the location of the undefined diagnostic is more precise,
and I've also enhanced the unused and undefined messages to make them more concise and clear.

### Undefined

image
image

### Unused

image

## Known issues

* [x] ~`NAMESPACE` works very well, but `mix release lexical` won't work on `lexical` project.~ Fixed by 1.15.4
* [x] ~compile twice. no idea about this issue, because `mix deps.compile` in command line has this problem too.~ Fixed by 1.15.4
* [x] ~We should wait until 1.15.3, because there are some bugs in `1.15.2`, Elixir core team already fixed them: https://github.com/elixir-lang/elixir/pull/12742#issuecomment-1625377364.~


apps/common/mix.exs
```diff
@@ -12,7 +12,8 @@ defmodule Common.MixProject do
       elixir: "~> 1.13",
       elixirc_paths: elixirc_paths(Mix.env()),
       start_permanent: Mix.env() == :prod,
-      deps: deps()
+      deps: deps(),
+      compilers: [:yecc] ++ Mix.compilers()
```
scottming, line 16: 1.16 need this, so this line only for debugging.

apps/remote_control/lib/lexical/remote_control/build/error.ex
```diff
@@ -96,6 +111,55 @@ defmodule Lexical.RemoteControl.Build.Error do
...
+  end
+
+  def format("undefined" <> _ = message) do
+    parts = String.split(message, " (")
+    List.first(parts)
+  end
+
+  def format(message) when is_binary(message) do
+    maybe_format_unused(message)
+  end
```
scohen, line 142: why are these public? why not just return the formatted messages?

apps/remote_control/lib/lexical/remote_control/build/error.ex
```diff
@@ -96,6 +111,55 @@ defmodule Lexical.RemoteControl.Build.Error do
...
+  @doc """
+  Format a diagnostic message to make it more readable
+  """
+  def format(%Result{message: message} = diagnostic) do
+    %{diagnostic | message: format(message)}
+  end
+
+  def format("undefined" <> _ = message) do
+    parts = String.split(message, " (")
+    List.first(parts)
```
scohen, line 137:
```suggestion
    message
    |> String.split(" (")
    |> List.first()
```

apps/remote_control/lib/lexical/remote_control/build/error.ex
```diff
@@ -96,6 +111,55 @@ defmodule Lexical.RemoteControl.Build.Error do
...
+  end
+
+  @doc """
+  Format a diagnostic message to make it more readable
+  """
+  def format(%Result{message: message} = diagnostic) do
+    %{diagnostic | message: format(message)}
+  end
+
+  def format("undefined" <> _ = message) do
```
scohen, line 135: add a comment about what this is doing and why.

apps/remote_control/lib/lexical/remote_control/build/state.ex
```diff
@@ -257,7 +263,8 @@ defmodule Lexical.RemoteControl.Build.State do
     end
 
     with_progress "mix deps.compile", fn ->
-      Mix.Task.run("deps.safe_compile", ~w(--skip-umbrella-children))
+      deps_compile = if System.version() >= "1.15", do: "deps.compile", else: "deps.safe_compile"
```
scohen, line 266: I _really_ dislike this syntax; use the full if syntax. It's so much more readable.
Also, you cannot compare strings and versions like this, there's an entire module for versions
```suggestion
   compile_task = 
      if Version.match?(System.version(), ">= 1.15") do 
        "deps.compile"
      else
        "deps.safe_compile"
      end
```

apps/remote_control/lib/lexical/remote_control/build/state.ex
```diff
@@ -270,13 +277,65 @@ defmodule Lexical.RemoteControl.Build.State do
   defp safe_compile(%Document{} = document) do
     old_modules = ModuleMappings.modules_in_file(document.path)
 
-    compile = fn ->
-      case RemoteControl.Mix.in_project(fn _ -> compile_code(document) end) do
+    compile_code_func =
+      if System.version() >= "1.15" do
```
scohen, line 281:
```suggestion
      if Version.match?(System.version(), "> 1.15") do
```

apps/remote_control/lib/lexical/remote_control/build/state.ex
```diff
@@ -270,13 +277,65 @@ defmodule Lexical.RemoteControl.Build.State do
...
+      end
+
+    compile_func = fn document ->
+      case RemoteControl.Mix.in_project(fn _ -> compile_code_func.(document) end) do
         {:ok, result} -> result
         other -> other
       end
     end
 
+    if System.version() >= "1.15" do
```
scohen, line 294:
```suggestion
    if Version.match?(System.version, "> 1.15") do
```

apps/remote_control/lib/lexical/remote_control/build/state.ex
```diff
@@ -270,13 +277,65 @@ defmodule Lexical.RemoteControl.Build.State do
   defp safe_compile(%Document{} = document) do
     old_modules = ModuleMappings.modules_in_file(document.path)
 
-    compile = fn ->
-      case RemoteControl.Mix.in_project(fn _ -> compile_code(document) end) do
+    compile_code_func =
```
scohen, line 280: `code_compiler`

apps/remote_control/lib/lexical/remote_control/build/state.ex
```diff
@@ -332,6 +391,35 @@ defmodule Lexical.RemoteControl.Build.State do
     end
   end
 
+  @dialyzer {:nowarn_function, compile_code_after_1_15: 1}
+
+  defp compile_code_after_1_15(%Document{} = document) do
```
scohen, line 396: nit: `compile_code_with_diagnostics`

apps/remote_control/test/lexical/remote_control/build/error_test.exs
```diff
@@ -9,28 +9,79 @@ defmodule Lexical.RemoteControl.Build.ErrorTest do
   def compile(source) do
     doc = Document.new("file:///unknown.ex", source, 0)
 
-    case Code.string_to_quoted(source) do
+    case Code.string_to_quoted(source, columns: true) do
       {:ok, quoted_ast} ->
+        if System.version() >= "1.15" do
```
scohen, line 14:
```suggestion
        if Version.match?(System.version(), ">= 1.15" do
```

apps/remote_control/test/lexical/remote_control/build/error_test.exs
```diff
@@ -9,28 +9,79 @@ defmodule Lexical.RemoteControl.Build.ErrorTest do
...
+
+      error ->
+        Logger.warning("error is #{inspect(error)}")
+        {doc, error}
+    end
+  end
+
+  defp do_compile(doc, quoted_ast) do
+    try do
+      modules = for {m, _b} <- Code.compile_quoted(quoted_ast, "/unknown.ex"), do: m
```
scohen, line 28: this is an instance of the "short" form of a comprehension making code that's less readable.
```suggestion
      modules = for {module _bytecode} <- Code.compile_quoted(quoted_ast, "/unknown.ex") do 
        module 
      end      
```

apps/remote_control/test/lexical/remote_control/build/error_test.exs
```diff
@@ -9,28 +9,79 @@ defmodule Lexical.RemoteControl.Build.ErrorTest do
...
+      {:ok, modules}
+    rescue
+      exception ->
+        {filled_exception, stack} = Exception.blame(:error, exception, __STACKTRACE__)
+        Logger.warning("Exception #{inspect(filled_exception)} stack: #{inspect(stack)}")
+        {doc, {:exception, filled_exception, stack, quoted_ast}}
+    end
+  end
+
+  defp do_compile_after_1_15(doc, quoted_ast) do
```
scohen, line 38: `compile_with_diagnostics`

apps/remote_control/test/lexical/remote_control/build/error_test.exs
```diff
@@ -92,11 +143,13 @@ defmodule Lexical.RemoteControl.Build.ErrorTest do
         |> compile()
         |> parse_error()
 
-      assert [error, detail] = errors
+      # assert [error, detail] = errors
```
scohen, line 146: remove?

apps/remote_control/test/lexical/remote_control/build/error_test.exs
```diff
@@ -92,11 +143,13 @@ defmodule Lexical.RemoteControl.Build.ErrorTest do
...
+      assert [detail, error] = errors
+
       assert error.message =~ "unexpected reserved word: end"
       assert error.position == {15, 9}
 
-      assert detail.message =~ ~S[The "(" here is missing terminator ")"]
+      assert detail.message =~ ~S["(" here is missing terminator ")"]
```
scohen, line 152: why change the assertion, just downcase the message with the old assertion.

apps/remote_control/test/lexical/remote_control/build/error_test.exs
```diff
@@ -9,28 +9,79 @@ defmodule Lexical.RemoteControl.Build.ErrorTest do
   def compile(source) do
     doc = Document.new("file:///unknown.ex", source, 0)
 
-    case Code.string_to_quoted(source) do
+    case Code.string_to_quoted(source, columns: true) do
```
scohen, line 12: You've duplicated a lot of the code in `build.ex`. I think it's time to extract some of that code into its own modules; say `Build.Compile.File` and `Build.Compile.Project`, and you can use those modules here directly.

I also think you're going to need to add a tag like `@tag :with_diagnostics` that only runs in >= 1.15. 

This means you'll have to add the following to `test_helper.exs`
```elixir
ExUnit.configure(exclude: [with_diagnostics: function_exported?(Code, :with_diagnostics, 1)])
```

apps/remote_control/lib/lexical/remote_control/build/state.ex
scottming: we should not use `Version.match?(System.version(), ">= 1.15") `, we should use `~> 1.15`

```elixir
iex(1)> Version.match?(System.version(), ">= 1.15")
** (Version.InvalidRequirementError) invalid requirement: ">= 1.15"
    (elixir 1.15.3) lib/version.ex:441: Version.parse_requirement!/1
    (elixir 1.15.3) lib/version.ex:287: Version.match?/3
    iex:1: (file)
iex(1)> Version.match?(System.version(), "~> 1.15")
true
iex(2)> Version.match?("1.16.0", "~> 1.15")
true
```

apps/remote_control/lib/lexical/remote_control/build/state.ex
scottming: how about `compile_quoted_with_dianostics`

apps/remote_control/test/lexical/remote_control/build/error_test.exs
scottming: because they changed `The` to `the`

apps/remote_control/lib/lexical/remote_control/build/state.ex
```diff
@@ -270,13 +277,65 @@ defmodule Lexical.RemoteControl.Build.State do
...
     end
 
+    if System.version() >= "1.15" do
+      do_compile(compile_func, document, old_modules)
+    else
+      capture_compile_io(compile_func, document, old_modules)
+    end
+  end
+
+  defp do_compile(compile_func, document, old_modules) do
```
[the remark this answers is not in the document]
scottming: much simpler now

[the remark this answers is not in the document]
scottming: once we drop support for Elixir < 1.15, we just need to remove `do_compile_and_capture_io` this function.

apps/remote_control/lib/lexical/remote_control/build/error.ex
```diff
@@ -49,21 +49,36 @@ defmodule Lexical.RemoteControl.Build.Error do
...
 
-  defp uniq(diagnostics) do
+  def uniq(diagnostics) do
     # We need to uniq by position because the same position can be reported
     # and the `end_line_diagnostic` is always the precise one
     extract_line = fn
       %Result{position: {line, _column}} -> line
       %Result{position: {start_line, _start_col, _end_line, _end_col}} -> start_line
-      %Result{position: line} when is_integer(line) -> line
+      %Result{position: line} -> line
```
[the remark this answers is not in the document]
scottming: we can't always find the integer line.

for example: 

```elixir
        ~S[
          Module.create(
            Foo,
            quote do
              String.to_integer("")
            end,
            file: "")
      ]
```

Do you think it is better to crash here rather than the diagnostics genserver?

if so, I can change this test: `    test "can't find right line when use macro" do`

apps/remote_control/lib/lexical/remote_control/build/error.ex
```diff
@@ -49,21 +49,36 @@ defmodule Lexical.RemoteControl.Build.Error do
...
     end
 
-    Enum.uniq_by(diagnostics, extract_line)
+    # Note: Sometimes error and warning appear on one line at the same time
+    # So we need to uniq by line and severity,
+    # and :error is always more important than :warning
+    extract_line_and_severity = &{extract_line.(&1), &1.severity}
```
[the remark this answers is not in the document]
scottming: but what I really want is to be sorted by line and severity, and unique by line.

for example: 

```elixir
    test "reports multiple errors", %{project: project} do
      source = ~S[
        defmodule WithFiveErrors do
          def error(unused_1, unused_2) do
            unknown_fn()
            unused_3 = other_unknown()
          end
        end
      ]
```

if we only use `line_and_severity`, it will return a duplicated error in one line, which is noise. 

```elixir
    [
      %Lexical.Plugin.V1.Diagnostic.Result{
        details: nil,
        message: "undefined function
 other_unknown/0",
        position: 5,
        severity: :error,
        source: "Elixir",
        uri:
          "file:///Users/scottming/Code/lexical/apps/remote_control/test/fixtures/project_metadata/3/file.exs"
      },
      %Lexical.Plugin.V1.Diagnostic.Result{
        details: nil,
        message: "variable \"unused_3\" is unused",
        position: {5, 13},
        severity: :warning,
        source: "Elixir",
        uri:
          "file:///Users/scottming/Code/lexical/apps/remote_control/test/fixtures/project_metadata/3/file.exs"
      }
    ]
```

Having the editor show one diagnostics on a line is enough, and the user will naturally see the other one after the fix.

If multiple are on one line, then there is really a lot of noise, like in this example before 1.15.

image

apps/remote_control/test/lexical/remote_control/build/error_test.exs
scottming: Honestly, I don't quite understand why this tag `@tag :with_diagnostics` will runs in >= 1.15?

apps/remote_control/lib/lexical/remote_control/build/state.ex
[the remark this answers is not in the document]
scohen: That will be _years_ in the future.

apps/remote_control/lib/lexical/remote_control/build/file.ex
```diff
@@ -0,0 +1,137 @@
+defmodule Lexical.RemoteControl.Build.File do
```
scohen, line 1: wasn't this supposed to be `RemoteControl.Build.Compile.File`?
Where's `RemoteControl.Build.Compile.Project` ?

apps/remote_control/test/lexical/remote_control/build/error_test.exs
scohen: because we define it to be true in the call to `ExUnit.configure` that I pasted in above.

Maybe a better condition would be:
```elixir
ExUnit.configure(exclude: [with_diagnostics: Version.match?(System.version(), "< 1.15.3")])
```

apps/remote_control/lib/lexical/remote_control/build/file.ex
scottming: > Where's RemoteControl.Build.Compile.Project ?

This PR's change is mainly related to *File** compilation, **Project** compilation not changed too much, so if we want, we can extract the project compilation logic in another PR.

> RemoteControl.Build.Compile.File?

I tried this, but don't you think `Build.Compile.File.compile(document)` is slightly strange? two `compile` here.

apps/remote_control/test/lexical/remote_control/build/error_test.exs
scottming: Are you sure we need this? Did you check the latest test code? we don't call `with_diagnostics` directly now.

apps/remote_control/test/lexical/remote_control/build/error_test.exs
scottming: ok

apps/remote_control/test/lexical/remote_control/build/error_test.exs
```diff
@@ -1,41 +1,36 @@
...
-            {doc, {:exception, filled_exception, stack, quoted_ast}}
-        end
+  use ExUnit.Case
+  use Patch
 
-      error ->
-        Logger.warning("error is #{inspect error}")
-        {doc, error}
+  setup do
+    if not Version.match?(System.version(), "~> 1.15") do
```
scohen, line 13: is this necessary? You can start the capture server all the time and not use it.

apps/remote_control/test/lexical/remote_control/build_test.exs
```diff
@@ -205,7 +205,12 @@ defmodule Lexical.BuildTest do
       assert diagnostic.uri
       assert diagnostic.severity == :error
       assert diagnostic.message =~ ~S[undefined function doesnt_exist/0]
-      assert diagnostic.position == 2
+
+      if version_after_1_15?() do
```
scohen, line 209: nit: you're not checking if the version is after 1.15, you're checking if you're using `with_diagnostics`

apps/remote_control/test/lexical/remote_control/build_test.exs
```diff
@@ -592,4 +615,8 @@ defmodule Lexical.BuildTest do
       assert loaded?(project, WithAType)
     end
   end
+
+  defp version_after_1_15? do
+    Version.match?(System.version(), "~> 1.15")
```
scohen, line 620: this should be `>= 1.15.3` right?

scohen (commented):
Looking pretty good.
I don't like the if statements in the tests, but I don't know of a better way to do it.

Just  a couple comments.

apps/remote_control/test/lexical/remote_control/build_test.exs
scottming: So, Do you mean changing the name to `with_diagnostics?` is better?

apps/remote_control/test/lexical/remote_control/build_test.exs
scottming: yes

apps/remote_control/lib/lexical/remote_control/build/error.ex
```diff
@@ -17,6 +37,70 @@ defmodule Lexical.RemoteControl.Build.Error do
...
+    case String.split(message, "is unused (", parts: 2) do
+      [prefix, _] ->
+        prefix <> "is unused"
+
+      _ ->
+        message
+    end
+  end
+
+  defp reject_zero_line(diagnostics) do
```
scohen, line 75: obscure english: this would be the `reject_zeroth_line`

I cannot stress how odd of a convention this is.

apps/remote_control/lib/lexical/remote_control/build/error.ex
```diff
@@ -17,6 +37,70 @@ defmodule Lexical.RemoteControl.Build.Error do
...
+
+  defp format(%Result{} = result) do
+    %Result{result | message: format_message(result.message)}
+  end
+
+  defp format_message("undefined" <> _ = message) do
+    # All undefined messages explain the *same* thing inside the parentheses,
+    # like: `undefined function print/1 (expected Foo to define such a function or for it to be imported,
+    #        but none are available)`
+    # that makes no sense and just creates noise.
```
scohen, line 52: I disagree strongly that this makes no sense. It's a helpful detail message that can indicate something isn't working like you would expect. 
I get a little concerned about the maintainability of these fixup functions, they're very dependent on messages being consistent from release to release.

I would suggest omitting them.

apps/remote_control/lib/lexical/remote_control/build/file.ex
scohen: yeah, but it's about having some consistent type of interface between file and project operations.

If the naming isn't great, then maybe we have `Build.File.compile` and `Build.Project.compile` ?

apps/remote_control/test/lexical/remote_control/build_test.exs
scohen: correct

apps/remote_control/test/lexical/remote_control/build_test.exs
scohen: `~> 1.15` and `>= 1.15.3` are different.

apps/remote_control/lib/lexical/remote_control/build/error.ex
scottming: image
You prefer the above, right?

scottming:
> To that end, I suggest we write a Features module in common that's something like:
Good idea.

apps/remote_control/lib/lexical/remote_control/build/file.ex
scottming: `Build.File.compile` yeah, we already did this.

scohen: well, `Build.Project.compile` isn't there though.

apps/common/lib/elixir/features.ex
```diff
@@ -0,0 +1,9 @@
+defmodule Elixir.Features do
+  def with_diagnostics? do
+    Version.match?(System.version(), ">= 1.15.3")
+  end
+
+  def compile_wont_change_directory? do
```
scohen, line 6: flagging for a follow up commit: `compile_keeps_current_directory?`

scohen (approved):

PR #289, opened 2023-07-27
ci: Require strict versions from erlef/setup-beam

Blond11516:


scohen (approved):

scohen:
what about in the test job, @Blond11516 ?

PR #290, opened 2023-07-27
Fixed code unit / codepoint confusion

scohen:
There was a bit of confusion here about code units / codepoints, which is now fixed. We were always talking about code units, and utf16 either has one 16 bit code unit or two 16 bit code units.

### DIFF
--- a/apps/common/lib/lexical/code_unit.ex
+++ b/apps/common/lib/lexical/code_unit.ex
@@ -83,7 +83,7 @@ defmodule Lexical.CodeUnit do
   end
 
   defp do_count_utf16(<<c::utf8, rest::binary>>, count) do
-    do_count_utf16(rest, count + codepoint_size(c, :utf16))
+    do_count_utf16(rest, count + code_unit_size(c, :utf16))
   end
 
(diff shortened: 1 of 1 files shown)


apps/common/lib/lexical/code_unit.ex
```diff
@@ -180,21 +180,25 @@ defmodule Lexical.CodeUnit do
...
   end
 
   @unicode_range 0..0x10_FFFF
 
-  defp codepoint_size(c, :utf16) when c in @unicode_range do
-    if c in 0x0000..0xD7FF or c in 0xE000..0xFFFF,
-      do: 1,
-      else: 2
+  defp code_unit_size(c, :utf16) when c in @unicode_range do
+    case c do
```
scottming, line 192: So is this the only place where the logic was changed and else just was refactor?

scottming (approved):

apps/common/lib/lexical/code_unit.ex
scohen: yes, the other logic was about codepoints, not code units.

PR #293, opened 2023-07-27
Fixed node naming conflict

### LINKED ISSUE #291: Potential issue with beam process name collisions
I'm presently working on a couple of different elixir apps and have them connected via `--sname`. The `sname` values i'm using for my two apps are their actual app_names. This is also what lexical seems to use. So I ended up getting the following error to my lsp logging (using nvim at `master`):

```
[ERROR][2023-07-27 11:12:22] .../vim/lsp/rpc.lua:676	"rpc"	"/Users/me/.local/share/lsp/lexical/_build/dev/rel/lexical/start_lexical.sh"	"stderr"	"Protocol 'inet_tcp': the name retriever@127.0.0.1 seems to be in use by another Erlang node\r\n"

```

It tends to stop working from this point, and it doesn't show as being connected any longer to the buffer.

I'm using lexical at commit 8b084d1596c21aa3eac21c973975c24997079f2a

scohen:
A reported bug showed that lexical's node names can conflict with other nodes on the same computer. This PR adds a discriminator to the project node name so it won't conflict

Fixes #291

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control.ex
+++ b/apps/remote_control/lib/lexical/remote_control.ex
@@ -18,11 +18,10 @@ defmodule Lexical.RemoteControl do
     :ok = ensure_epmd_started()
     entropy = :rand.uniform(65_536)
     start_net_kernel(project, entropy)
-
     apps_to_start = [:elixir | @allowed_apps] ++ [:runtime_tools]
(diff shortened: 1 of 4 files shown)


scohen:
Closing in favor of #294

PR #296, opened 2023-07-28
Added support for per-file .eex compilation

scohen:
Refactor / improvement of per-file compilation. Prior, we were just shunting all files into the the elixir compiler, but as it turns out, there are other types of files that need to be compiled. This PR makes a compiler behaviour, and three implementations: elixir, eex and no-op.


apps/remote_control/lib/lexical/remote_control/build/document/compilers/quoted.ex
```diff
@@ -0,0 +1,127 @@
+defmodule Lexical.RemoteControl.Build.Document.Compilers.Ast do
+  alias Elixir.Features
+  alias Lexical.Document
+  alias Lexical.RemoteControl.Build
+  alias Lexical.RemoteControl.ModuleMappings
+
+  import Lexical.RemoteControl.Build.CaptureIO, only: [capture_io: 2]
+
+  def compile_quoted(%Document{} = document, quoted_ast, compiler_name) do
```
scottming, line 9: CleanShot 2023-07-29 at 19 32 58@2x
CleanShot 2023-07-29 at 19 32 53@2x
Don't understand why duplicate these two files.

scohen: It was a rename that went astray.

scottming (approved):

Issue #301, opened 2023-07-31
Lexical currently does not work with `config/*.exs`


HeapUnderfl0w:
Currently lexical does not handle editing the configuration in the `config/*.exs` files.
Opening such a file with `import Config` at the top results in the following lsp message:

```
could not set configuration via Config. This usually means you are trying to execute a configuration file directly, instead of reading it with Config.Reader
```

This is easily reproducable with any exs that includes `import Config`.
To reproduce this standalone:

1. Create a new project `mix new mre`
2. Create `mre/config/config.exs` with the content:
```elixir
import Config

config :mre, foo: :bar
```
3. Open the file with a editor that uses lexical as a lsp

System: `Nixos x64 (Kernel 6.3.13)`
Erlang: `Erlang (SMP,ASYNC_THREADS) (BEAM) emulator version 13.2.2.2`
Elixir: `Elixir 1.14.5 (compiled with Erlang/OTP 25)`
Lexical Rev: 1a2c891a7a438662dbd64aa205a28b7f7ce25e75
Running under `GNU Emacs 29.0.92` using `lsp-mode` on rev `3862b237a56b02ddd88c9dcfc74c9ee3d3f936a5`

PR #304, opened 2023-07-31
Config compiler

scohen:
We were treating configuration files in the project's config directory
like normal elixir source files, but they're slightly different, and
compiling them will cause errors to appear.

This change adds a config compiler that uses Config.Reader to compile
configuration.

Fixes https://github.com/lexical-lsp/lexical/issues/301

### DIFF
--- a/apps/common/lib/elixir/features.ex
+++ b/apps/common/lib/elixir/features.ex
@@ -1,6 +1,6 @@
 defmodule Elixir.Features do
   def with_diagnostics? do
-    Version.match?(System.version(), ">= 1.15.3")
+    function_exported?(Code, :with_diagnostics, 1)
   end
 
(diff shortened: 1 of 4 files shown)


scottming (approved):
I've tested this branch in some projects and found that the `config folder` under umbrella is not recognized, or using `import_config "#{config_env()}.exs"` is also problematic, but I think it might be a result of incorrect configuration.

scohen:
What do you mean? Errors are emitted just fine in the lexical project.

scottming:
Sorry, maybe I didn't explain it clearly enough. What I meant was that some `sub-apps` have already specified the config_path: "../../config/config.exs", but someone still created a config directory in the app, which caused recognition failure.

So never mind.

Issue #295, opened 2023-07-28
Request textDocument/completion failed


objectuser:
Compiled with Elixir 1.15.4, Erlang 26.0.2:

```sh
$ NAMESPACE=1 mix release lexical --overwrite
```

```elixir
[Error - 10:02:01 AM] Request textDocument/completion failed.
  Message: ** (FunctionClauseError) no function clause matching in LXical.Server.CodeIntelligence.Completion.complete/4
    (lx_server 0.1.0) lib/lexical/server/code_intelligence/completion.ex:30: LXical.Server.CodeIntelligence.Completion.complete(%LXical.Project{root_uri: "file:///...", mix_exs_uri: "file:///.../mix.exs", mix_project?: true, mix_env: nil, mix_target: nil, env_variables: %{}, project_module: nil}, #LXical.Document<path: "/.../...ex", version: 10, dirty?: false, lines: %Lines<"defmodule ... do..."(70 lines)>, ...>, #Protocol.Types.Position<[line: 51, character: 73]>, #Protocol.Types.Completion.Context<[trigger_kind: :invoked]>)
    (lx_server 0.1.0) lib/lexical/server/provider/handlers/completion.ex:12: LXical.Server.Provider.Handlers.Completion.handle/2
    (lx_server 0.1.0) lib/lexical/server/provider/queue.ex:99: anonymous fn/2 in LXical.Server.Provider.Queue.State.as_task/2
    (elixir 1.15.4) lib/task/supervised.ex:101: Task.Supervised.invoke_mfa/2
    (elixir 1.15.4) lib/task/supervised.ex:36: Task.Supervised.reply/4

  Code: -32603 
```

Recompiled without `NAMESPACE=1` and the error does not occur.

Related Discord conversation [here](https://discord.com/channels/269508806759809042/918183981344772117/1134517086853546024).

scohen:
Talked a lot about this on discord; 
This issue didn't recur after @objectuser killed extraneous BEAM processes on his laptop. Closing for now, as I was also experiencing this sporadically before the refactor of namespacing.

PR #314, opened 2023-08-09
Reworked lexical packaging

### LINKED ISSUE #255: Improve project packaging
We're using releases to generate lexical's build. This works well, but it has a couple of drawbacks:

  1. They are rather large
  2. They're architecture and version specific, which is the source of our not being able to have a different project version of elixir/erlang.

Other projects and LSPs seem to be using `mix.install` to install and compile the app globally, but I was unable to get this working. This is entirely due to namespacing. When we generate a namespaced app, we rewrite the app file, boot files, and script files of all of the apps that have namespacing applied to them.
I tried building a namespace compiler that would reach into mix.install's cache and apply namespacing to it,  which worked as far as the namespacing was applied successfully. The problem is that the install script that calls `mix.install` still refers to the original app names, and starts them, which isn't what we want. 

Worse, subsequent runs will recompile and reinstall those apps, which will cause problems if those apps are ever run inside lexical.

We need a packaging system that:
 
  1. Works with a wide variety of elixir and erlang combinations
  2. Doesn't need to be built for each variation of elixir / erlang. Releases are _very_ specific in this regard, failing at even minor version differences
  3. Can be installed from editor tooling easily
  4. Allow us to namespace our apps

scohen:
This commit moves away from releases for our packaging. What replaces them is a package directory that has a launcher, working config and priv directories and a bunch of erlang archives for storing our beam files.

New packaging enables the following:

  * We can compile on one version of erlang and run on another
  * We get a real priv directory experience, which isn't possible with escripts.
  * We get a "real" config directory
  * We get a boot script that's just a plain old elixir file
  * All of our files can be ported to windows fairly easily
  * No tricks for packaging scripts / config. They're just plain files, and we rely on the `:code` module to load them

I've tested this compiling in one version of elixir and erlang and running it in another, and it seems to work great. It makes sense though, they're just BEAM files, which should be cross platform.

Other approaches:

We all know why releases won't / don't work. I tried building an `escript`, and it kind of worked, but had the following drawbacks:

  * There is no priv directory. I was able to code around this by having a module that generates the port_mapper.sh file for you, but this was a bit cumbersome.
  * An escript is a binary file, and erlang loads all of its modules from it. This was a problem for the second VM, as we couldn't just point it at the binary file and have it look for modules there. I also tried using the 'inet' loader, but that had a bunch of extremely difficult to debug issues. Worse, I would still need some erlang archives, so why not just make it universal?

I also tried using `mix archive.build`, but it didn't support namespacing (obviously) and changing the task to do so would pretty much lead us to the path we have here.

Fixes #255


scohen:
@Blond11516 This will have an effect on the vscode extension, please take a look.

apps/remote_control/lib/mix/tasks/namespace/path.ex
```diff
@@ -24,7 +24,11 @@ defmodule Mix.Tasks.Namespace.Path do
         |> Namespace.Module.apply()
         |> Atom.to_string()
 
-      String.replace(path, string_name, namespaced_name)
+      if String.contains?(path, namespaced_name) do
```
scohen, line 27: this prevents double namespacing. I saw that we were generating `lx_lx_protocol.app`

pages/installation.md
```diff
@@ -101,22 +105,26 @@ Click on the extensions button on the sidebar, then search for
 download the latest version of Lexical.
 
 To change to a local executable, go to `Settings -> Extensions -> Lexical` and
-type `/my/home/projects/lexical/_build/dev/rel/lexical` into the text box in
+type `/my/home/projects/lexical/_build/dev/package/lexical` into the text box in
```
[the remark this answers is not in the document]
scohen: Thank you!
 ❤️

apps/server/lib/mix/tasks/package.ex
```diff
@@ -0,0 +1,311 @@
...
+      path: :string,
+      zip: :boolean
+    ]
+  ]
+
+  @execute_permisson 0o755
+
+  def run(args) do
+    Mix.Task.run(:compile)
+    {opts, _, _} = OptionParser.parse(args, @options)
```
Blond11516, line 82: Didn't know there's an `OptionParser` module in the standard library. Pretty neat!

apps/server/lib/mix/tasks/package.ex
```diff
@@ -0,0 +1,311 @@
...
+    default_path = Path.join([Mix.Project.build_path(), "package", "lexical"])
+    package_root = Keyword.get(opts, :path, default_path)
+
+    Mix.Shell.IO.info("Assembling buld in #{package_root}")
```
Blond11516, line 86:
```suggestion
    Mix.Shell.IO.info("Assembling build in #{package_root}")
```

Blond11516 (approved):
Don't have access to a device I can test this with right now, I'll try to do that later today.

That's a really cool packaging solution! Congrats on coming up with that, it's a load of manual work but I think it works out to something quite elegant given our requirements!

Regarding VSCode auto-install, if I understand correctly the only thing that would have to change is to launch through the `package` directory rather than `rel`? I'll be unavailable most of the weekend, but I should be able to update and release that by early next week at the latest, maybe tomorrow if I can find some time after work.

scohen:
@Blond11516 I pushed a PR to `vscode-lexical`. Be prepared to laugh.

apps/server/lib/mix/tasks/package.ex
scohen: Yeah, it's great. There's also a version comparator. Love it.

scohen:
> That's a really cool packaging solution! Congrats on coming up with that, it's a load of manual work but I think it works out to something quite elegant given our requirements!

Yeah, to say that I'm happy with this solution is an understatement. The meat of this PR is 235 lines of code, and the rest is mostly deletions and doc fixes.

Blond11516:
Did a quick test: compiled Lexical with Erlang 24.3.4.13 and Elixir 1.13.4, working great in a project using Erlang 25.1.2 and Elixir 1.14.2!

scohen:
@Blond11516 remind me which os you use

Blond11516:
@scohen I primarily use Linux

objectuser:
Been using this since yesterday afternoon and have not encountered #295 since.

scohen:
@Blond11516 awesome, I'm comfortable to merge this now, I think everyone else is on MacOS.

scohen:
@scottming I'm going to merge this. We've had verification that it works across OSes, and the approach is vastly superior to releases for our use case. 

Feel free to chime in with issues on this PR if you want something changed.

apps/remote_control/lib/lexical/remote_control/project_node.ex
```diff
@@ -69,12 +73,25 @@ defmodule Lexical.RemoteControl.ProjectNode do
       %{state | status: :stopped}
     end
 
-    def on_nodeup(%__MODULE__{} = state) do
-      %{state | status: :started}
+    def on_nodeup(%__MODULE__{} = state, node_name) do
+      if node_name == Project.node_name(state.project) do
```
scottming, line 77: Is this if-else mainly for `debug_shell`?

apps/remote_control/lib/lexical/remote_control/project_node/launcher.ex
```diff
@@ -0,0 +1,19 @@
...
+  @moduledoc """
+  A module that provides the path of an executable to launch another
+  erlang node via ports.
+  """
+  def path do
+    path(:os.type())
+  end
+
+  def path({:unix, _}) do
+    with :non_existing <- :code.where_is_file(~c"port_wrapper.sh") do
```
scottming, line 11: TIL

apps/server/lib/lexical/server/boot.ex
```diff
@@ -0,0 +1,62 @@
...
+      :non_existing ->
+        :error
+
+      path ->
+        {:ok, List.to_string(path)}
+    end
+  end
+
+  defp load_app_modules(app_name) do
+    Application.ensure_loaded(app_name)
```
scottming, line 56: So, we started `:server` app, and loaded deps' app, right?

I'm curious why we do not need to start dep's app?

apps/remote_control/lib/lexical/remote_control/project_node.ex
scohen: Yes

apps/server/lib/lexical/server/boot.ex
scohen: 'Good question, we are actually starting the server app and all the apps it depends on. `Application.ensure_all_started` starts the entire supervision tree of the app given and all apps that it depends on, in he proper order.

PR #317, opened 2023-08-10
Fix extraneous logging in test

scohen:
This PR is two commits, the first shuts up logging in test, but I don't understand why it's broken in the first place. More investigation is warranted.

The second cleans up the config files a bit and removes an unhandled message log message.

### DIFF
--- a/apps/plugin_runner/mix.exs
+++ b/apps/plugin_runner/mix.exs
@@ -28,6 +28,7 @@ defmodule Lexical.Plugin.Runner.MixProject do
   defp deps do
     [
       {:common, in_umbrella: true},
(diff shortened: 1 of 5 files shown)


Blond11516 (approved):
For what it's worth, `Logger.configure` [doesn't use app env to update the `level`](https://github.com/elixir-lang/elixir/blob/c521bdb91a77b36be16fdf18d632ad7719de4f91/lib/logger/lib/logger.ex#L618) (and only that option) so this might be in indication that something special is going on with it.

On the other hand, the doc explicitly says config files should work, so... :shrug:

scohen:
I know, I'm very confused about this.

PR #323, opened 2023-08-13
Support for HEEx compilation

scottming:
This PR adopts an approach that divides the compilation of HEEx into three steps. First, it focuses on the HTML portion, then the EEx part, where certain errors in EEx require evaluation to obtain, and finally, the compile_quoted step.
I think this layered approach is quite good, like peeling an onion.

Though there are some problems that I haven't thought of a good solution for now, such as

1. In a HEEx file, when a function component does not exist, `mix compile` will emit an error, but how can this be achieved for single file compilation? Jose suggested that we compile the relevant `ex` files directly, but it seems difficult because the latest content of those files may only exist in the `Document.Store` of the server node.
2. In a `.ex` file, like `core_components.ex`, how can the `~H` block be compiled separately?
3. The `~p` block in a `~H` block or `HEEx` file.

### as-you-type compilation

https://github.com/lexical-lsp/lexical/assets/12830256/4900bc51-b771-4c5f-9d7a-32c7140bd9f5

### after saving

https://github.com/lexical-lsp/lexical/assets/12830256/53afa6f2-8e26-4918-b0dc-192515acb9bd


apps/remote_control/lib/lexical/remote_control/build/document/compilers/heex.ex
```diff
@@ -0,0 +1,145 @@
...
+          document
+          |> Build.Error.error_to_diagnostic(exception, stack, quoted_ast)
+          |> Map.put(:source, "HEEx")
+
+        {:error, [converted]}
+
+      {{:ok, quoted_ast}, _} ->
+        # Ignore warnings for now
+        # because they will be handled by `compile_quoted/2`
+        # like: `assign @thing not available in EEx template`
```
scottming, line 88: Maybe instead of ignoring them, I should do something like `reject_undefined_assigns` in a unified place, for all the errors.

apps/remote_control/lib/lexical/remote_control/build/document/compilers/heex.ex
```diff
@@ -0,0 +1,145 @@
...
+          [converted | maybe_diagnostics]
+          |> Enum.reverse()
+          |> Build.Error.refine_diagnostics()
+          |> Enum.map(&Map.replace!(&1, :source, "HEEx"))
+
+        {:error, diagnostics}
+    end
+  end
+
+  defp eval_quoted_with_diagnostics(quoted_ast, path) do
```
scottming, line 105: Should we move this function and the below function to the `quoted.ex`?

Similar to the `compile_quoted`, I think we need the similar implementation  for `eval_quoted`

apps/remote_control/lib/lexical/remote_control/build/document/compilers/heex.ex
scottming: yes

scohen (approved):
This looks solid.

PR #331, opened 2023-08-14
Add support for `textDocument/hover` (for modules)

zachallaun:
This first commit is essentially just scaffolding, including the generated types for `textDocument/hover` and basic request/response handling. It currently replies with static content.

There's some additional WIP code for retrieving the docs for the entity being hovered, but I'm not sure how much of it will actually end up in the final feature.

---

@scohen, some thoughts/questions on the best way to proceed that I'd love your input on.

1. The hover response contains two things, an optional range, representing the entity that you retrieved hover content for, and the contents themselves as markdown. For this initial implementation, I'd thinking to handle modules, functions, and types -- see the `Project.Docs` fixture for what I'm planning to test with to start. Does this seem reasonable?
2. I'm planning to use `Code.fetch_docs/1` to retrieve docs. It retrieves docs for an entire module, all functions/types/etc. defined within. As per (#223), it'll eventually be worth caching this stuff, but for now I'm thinking to just scan the result each time.
3. Go-to-definition currently uses `ElixirSense` to retrieve the definition given a doc/line/character. I'm assuming that I should also use it for now to retrieve the range and module associated with the currently hovered entity?


zachallaun:
4. This shares a lot of behavior with `Definition` -- I'm inclined to co-locate them instead of adding a new `Docs` or something.

scohen:
@zachallaun 
1. Yes, this seems reasonable. I'd even say this sounds like a lot, perhaps start with modules and functions?
2. Go for it, I haven't made any progresson the intelligence and indexing features.
3. I would like to stay away from elixir_sense, as it's a dependency that is going away, and doesn't have a release cycle or versioning. I do understand that identifying stuff under the cursor is pretty tricky in elixir. I'd like your thoughts on what you feel comfortable doing.
4. Ok, maybe the name `Definition` needs work then?

zachallaun:
1. Let's see how much we get "for free" -- the hard part is figuring out the entity under cursor, after that it's just filtering a list. 
3. This is somewhat related to co-locating with `Definition`, which also relies on `elixir_sense`. So if we co-locate, I'll factor things such that when `elixir_sense` is removed, there's a minimal interface to replace. I'll keep an eye on opportunities to remove dependence on it, though. Using `Code.fetch_docs` directly is one such opportunity already, vs. `ElixirSense.docs`. 
4. Agreed, I was going to bring that up. `CodeEntity`? Just `Entity`? `Artifact`? I'm not really sure what the nomenclature should be -- basically any unit that can be defined somewhere and referred to either by an atom or module-function-arity.

scohen:
1. This is what's dangerous about using ElixirSense; you get a LOT for free, and then it becomes super hard to replace the whole lot. At this point, I'd almost favor going the other way and starting small with, say Modules, and working our way up. 
2. - 
3. -
4. I'd favor `SytaxEntity` or `CodeEntity`. Perhaps in a `Syntax.Entity` or `Code.Entity` module. Thoughts?

zachallaun:
1. Sounds good -- I'll start with modules and see if we can avoid elixir sense altogether. Basically these cases:

```elixir
Some.Module| # docs for Some.Module

alias Some.Module
Module| # docs for Some.Module

Some.Module|. # compile error, should still be able to identify Some.Module

Some|.Module # docs for Some
```

4. I think `Code.Entity` makes sense. This is pretty foundational, probably the kind of thing that will be in `shared` or `common` eventually (plugins will want access to it, methinks). So it'll basically be `Lexical.Code.Entity`, I think?

scohen:
Aliases can get pretty tricky, but luckily we have an alias tracker in `lib/lexical/code_mod/ast/aliases.ex` It allows you to give the ast and then do `Aliases.at(doc, position)` which in your second example, will give you a map of  `%{:Module => Elixir.Some.Module}`

scohen:
> So it'll basically be Lexical.Code.Entity, I think?

Yep, start off in `common` and if we need it in plugins, we'll move it to `lexical_shared`.

zachallaun:
I'm going to mark this PR as ready for review. Currently it is able to resolve modules under the cursor and display module docs. The code (and commits) need to be cleaned up a bit, however. That said, I think it's in a good-enough spot that it's useful to get feedback and make sure I'm not *too* far astray.

image

apps/remote_control/lib/lexical/remote_control/api.ex
```diff
@@ -60,4 +60,10 @@ defmodule Lexical.RemoteControl.Api do
       position
     ])
   end
+
+  def docs(%Project{} = project, module) when is_atom(module) do
+    with {:module, _} <- RemoteControl.call(project, Code, :ensure_compiled, [module]) do
```
zachallaun, line 65: I'm not sure that this `Code.ensure_compiled/1` call is necessary. I was seeing some strange behavior in the hover tests, where the first call to `Code.fetch_docs/1` was returning an error, but a second call was returning the docs -- in my experience, stuff like that usually happens when something hasn't been fully loaded/etc. So this might really just be a "fix" for a problem in the test setup for the hover provider.

apps/remote_control/lib/lexical/remote_control/code_mod/ast.ex
```diff
@@ -14,17 +12,7 @@ defmodule Lexical.RemoteControl.CodeMod.Ast do
...
-
-  defp parse(s) when is_binary(s) do
-    case Sourceror.string_to_quoted(s, token_metadata: true, columns: true) do
-      {:ok, code, _comments} ->
-        {:ok, code}
-
-      error ->
-        error
-    end
+    Sourceror.parse_string(s)
```
zachallaun, line 15: When Sourceror calls `string_to_quoted` internally, it sets a number of additional options that add metadata to parts of the AST and are later used with `Sourceror.get_range/1`, which I was relying on.

apps/server/lib/lexical/server/code_intelligence/ast.ex
zachallaun, line 1: Note to self: delete this

apps/server/lib/lexical/server/code_intelligence/entity.ex
```diff
@@ -0,0 +1,211 @@
...
+  defp current_module_aliases(nil), do: {:error, :missing_defmodule}
+
+  defp current_module_aliases(zipper) do
+    case Zipper.node(zipper) do
+      {:defmodule, _, [{:__aliases__, _, aliases} | _]} -> {:ok, aliases}
+      _ -> zipper |> Zipper.up() |> current_module_aliases()
+    end
+  end
+
+  defp innermost_zipper_at(ast, position) do
```
zachallaun, line 93: I spent some time trying to refactor this to get past the Credo check, but couldn't come up with anything I liked better. Suggestions welcome.

apps/server/test/lexical/server/code_intelligence/entity_test.exs
```diff
@@ -299,30 +304,224 @@ defmodule Lexical.Server.CodeIntelligence.DefinitionTest do
...
-  defp caller_position(subject_module) do
-    {line, character} = cursor_position(subject_module)
-    Position.new(line, character)
+  describe "resolve/2" do
+    test "syntax error", %{project: project} do
+      code = ~q[
+        Some.Modul|e.
+        ]
+
+      assert {:error, _} = resolve(project, code)
```
zachallaun, line 321: I originally wanted to support this case but was struggling to figure out how to handle it. I believe it should be possible, but I need to use `container_cursor_to_quoted` at the position of the syntax error, not the position of the cursor, when creating the fragment. I'm inclined to not support hover in the case of syntax errors to start (lexical immediately highlights the error anyways), and then add more robust entity resolution in a later PR.

apps/remote_control/lib/lexical/remote_control/code_mod/ast.ex
zachallaun: This was swapped back to a call to `string_to_quoted`, but includes all the options Sourceror still requires. See f58c0008ace899567092d41fbd79a53fa3e5cf30 for more information.

apps/remote_control/lib/lexical/remote_control/api.ex
scottming: I believe this is necessary because we cannot guarantee that the module is compiled when the user calls the hover feature, especially when we open the editor and the project compilation time is long.

apps/server/lib/lexical/server/code_intelligence/entity.ex
```diff
@@ -0,0 +1,240 @@
...
+  # Corrects dot-call issue
+  defp fetch_range({:., _, [left, atom]}) when is_atom(atom) do
+    with {:ok, range} <- fetch_range(left) do
+      # Unhandled edge case: quoted atoms, e.g. Foo."bar-baz"
+      len = atom |> Atom.to_string() |> String.length()
+      {:ok, update_in(range.end[:column], &(&1 + 1 + len))}
+    end
+  end
+
+  # There are certain AST nodes that fail on Sourceror.get_range/1, like
```
scottming, line 181: The current implementation appears to be overly complex to me, especially the logic for finding the innermost node and fetching the range.

This leads me to question whether it's really necessary to use Sourceror for this task. The power of the zipper lies in its ability to facilitate node modifications, but do we really need to make extensive modifications to nodes? We only need to find the entity under the current cursor, and currently, we only handle modules.

I'm not sure if `cursor_context` + `container_cursor_to_quoted` + `surround_context` can be helpful in this regard. For example, in a case like `AAA.{BBB.C|CC`, perhaps we could first use them to find AAA.C, and then combine it with the remaining `CC`. This approach might simplify the code. We can also move the cursor for other cases if needed.

apps/server/test/lexical/server/code_intelligence/entity_test.exs
```diff
@@ -299,30 +304,247 @@ defmodule Lexical.Server.CodeIntelligence.DefinitionTest do
...
+
+      assert {:ok, {:module, At.The.End}} = resolve(project, code)
+
+      code = ~q[
+        Beginning.Of.The|.End
+        ]
+
+      assert {:ok, {:module, Beginning.Of.The}} = resolve(project, code)
+
+      code = ~q[
```
scottming, line 351: I don't see the difference between this case and `At.The.En|d`.

I even feel that repeating such things is unnecessary. Tests require extreme focus and should have clear differentiation.

apps/server/test/lexical/server/code_intelligence/entity_test.exs
```diff
@@ -299,30 +304,247 @@ defmodule Lexical.Server.CodeIntelligence.DefinitionTest do
...
+          Modul|e
+        end
+        ]
+
+      assert {:ok, {:module, Long.Aliased.Module}} = resolve(project, code)
+
+      code = ~q[
+        defmodule Example do
+          alias Long.Aliased.Module
+          Module.Neste|d
```
scottming, line 439: Same problem, One example is sufficient; you are just duplicating some logic with the aliases tests. Moreover, I believe it is already addressed in the `aliases_tests.exs` file.

scottming (changes requested):

apps/remote_control/lib/lexical/remote_control/code_mod/ast.ex
```diff
@@ -17,16 +17,62 @@ defmodule Lexical.RemoteControl.CodeMod.Ast do
...
   defp parse(s) when is_binary(s) do
-    case Sourceror.string_to_quoted(s, token_metadata: true, columns: true) do
-      {:ok, code, _comments} ->
-        {:ok, code}
-
-      error ->
-        error
+    parsed =
+      s
+      |> Sourceror.string_to_quoted(
```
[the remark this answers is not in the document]
zachallaun: There's a fairly large comment above the function header that's intended to explain these options, but perhaps it's too long or unclear, or would benefit from a couple of additional comments where the options are used? For example:

```elixir
# ... existing long comment
defp parse(s) when is_binary(s) do
  parsed =
    s
    |> Sourceror.string_to_quoted(
      # specify a static atoms encoder to solve for the problem explained above
      static_atoms_encoder: &encode_static_atoms/2,
      # remaining options required to use full Sourceror API
      literal_encoder: &{:ok, {:__block__, &2, [&1]}},
      token_metadata: true,
      unescape: false,
      columns: true,
      warn_on_unnecessary_quotes: false,
      emit_warnings: false
    )

  ...
end
```

apps/server/lib/lexical/server/code_intelligence/entity.ex
```diff
@@ -0,0 +1,240 @@
...
+        {{_, _, _} = node, _} = zipper, {current, found_line?} ->
+          case fetch_range(node) do
+            {:ok, range} ->
+              same_line? = line == range.start[:line] and line == range.end[:line]
+              line_in_range? = line >= range.start[:line] and line <= range.end[:line]
+              # sourceror column ranges are inclusive at the start and exclusive at the end
+              char_in_range? = char >= range.start[:column] and char < range.end[:column]
+
+              cond do
+                same_line? and char_in_range? ->
```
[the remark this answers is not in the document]
zachallaun: Agreed with this an your other comments on other branches. I'll make sure that examples are present.

apps/server/lib/lexical/server/code_intelligence/entity.ex
zachallaun: I do not explain the motivation for this strategy in the code or in comments, and that's a mistake. I'll try to do so now, and then, if you all agree that this is a reasonable approach, I'll work on including better explanation in docs/comments such that it's more accessible to others in the future.

First, I'd like to address this comment:

> The power of the zipper lies in its ability to facilitate node modifications...

I don't agree; while zippers certainly do facilitate node modifications, the power of zippers is arbitrary movement. With standard tree traversal, a given node can easily inspect its children, but it cannot inspect its parent or siblings (except with the addition of a complex accumulator, at which point you're reinventing a zipper!).

I'll come back to zippers with a motivating example in a moment.

The ultimate goal of `resolve/2` is to resolve a "code entity" at a specific cursor position. While it currently handles only module aliases (at [@scohen's suggestion](https://github.com/lexical-lsp/lexical/pull/331#issuecomment-1678182334)), it's designed to be extended to many other use-cases. Consider the following AST, using `|` to represent various positions:

```elixir
defmodule |My.Module do # (1)
  alias My.Module.Neste|d # (2)

  @type some_type :: {:ok, Nested.|t()} | :error # (3)

  def fun(%Nested{} = arg) do
    ar|g # (4)
    |> Nested.|process_arg(%|{data: :literal}) # (5) (6)
    |> Nested.finish()
  end
end
```

The entities associated with these positions might be something like:

1. `{:module, My}`
2. `{:module, My.Module.Nested}`
3. `{:type, {My.Module.Nested, t, 0}}`
4. `{:argument, {My.Module, :fun, 1, 0}}`
5. `{:call, {My.Module.Nested, :process_arg, 2}}`
6. `{:literal, :map}`

For `textDocument/hover`, each of these entities might have docs or other useful information associated with them.

For instance, hovering (4) might show the relevant part(s) of the function header that specify that argument. In order to know that this is a bound argument to the function and not a local variable or import, you *must* inspect its context. This is where zipper shine. By exposing the innermost AST node for a position as a zipper, we can now inspect the contextual environment of that node to understand how its being used.

Getting back to modules -- the reason for the custom range stuff and changes to `Ast.parse/` is that, otherwise, this cannot be used for certain multi-line nodes, namely:

```elixir
# this is unusual and not idiomatic, but it's valid
Foo.
  Bar.
  Baz

# by default, this parses into
{:__aliases__, [line: 1, column: 1, last: [line: 3, column: 3], [:Foo, :Bar, :Baz]}
```

If we're hovering over `Bar`, there is no element more specific than the `:__aliases__` node to consider. By using `:static_atoms_encoder`, we get this parse tree, which allows us to consider `Bar` and the alias segments prior:

```elixir
{:__aliases__, [line: 1, column: 1, last: [line: 3, column: 3],
 [{:__block__, [line: 1, column: 1], [:Foo]},
  {:__block__, [line: 2, column: 3], [:Bar]},
  {:__block__, [line: 3, column: 3], [:Baz]}]}
```

In order to use this somewhat modified AST, we have to wrap `Sourceror.get_range`, which doesn't expect these additional `:__block__` nodes.

---

This is getting too long, so to summarize and make a few final points:

1. `resolve/2` is intended as a general-purpose API to discover the *kind of entity* at a position, and that requires understanding its surroundings/context.
2. Zippers are especially well-suited to inspecting the context of a given node and are, I think, appropriate to use here.
3. I think this code could/should be better organized. For instance, `fetch_range` is addressing changes that are happening in `Ast`, and should live/be documented next to that.

apps/server/test/lexical/server/code_intelligence/entity_test.exs
zachallaun: Agreed that only one example is sufficient here.

I do not agree that we should consider the `aliases_test.exs` as sufficient, however, as there's no guarantee this API continues to use `Aliases` in the future. `Aliases` is a unit that should be tested, and this is a unit, one level of abstraction up, that also needs to be tested.

apps/server/lib/lexical/server/code_intelligence/entity.ex
zachallaun: Disregard; I think this can be built using much of the existing work for completions. I'm going to study that a bit.

apps/remote_control/lib/lexical/remote_control/api.ex
scohen: these two functions feel a bit like business logic that should go inside `RemoteControl` somewhere.

apps/remote_control/lib/lexical/remote_control/code_mod/ast.ex
```diff
@@ -36,7 +82,11 @@ defmodule Lexical.RemoteControl.CodeMod.Ast do
           within_range?(node, position)
         end)
 
-      {:ok, zipper}
+      if zipper do
+        {:ok, zipper}
+      else
+        {:error, :not_found}
+      end
```
scohen, line 89: it feels like this should go inside the with. Perhaps extract the pipe into a function
```elixir
with {:ok, ast} <- from(document) do 
  zipper_at_position(ast, position)
end

defp zipper_at_position(ast, position) do 
  zipper = 
    ast 
    |> Zipper.zip()
    |> Zipper.find(&within_range?(&1, position)

  if zipper do 
    {:ok, zipper}
  else
    {:error, :not_found}
  end
end
```

apps/remote_control/lib/lexical/remote_control/code_mod/ast.ex
```diff
@@ -17,16 +17,62 @@ defmodule Lexical.RemoteControl.CodeMod.Ast do
...
   end
 
+  defp encode_static_atoms(atom_name, meta) do
+    meta = meta ++ [static_atom: true]
+    {:ok, {:__block__, meta, [String.to_atom(atom_name)]}}
+  end
+
+  defp remove_unnecessary_wrapped_static_atoms(ast) do
+    Sourceror.prewalk(ast, fn
+      {{:__block__, block_meta, [atom]}, meta, args} = node, state ->
```
scohen, line 64: I like to add comments to AST literals explaining what they are.

apps/server/lib/lexical/server/code_intelligence/entity.ex
```diff
@@ -0,0 +1,240 @@
...
+        error
+    end
+  end
+
+  defp resolve({:__block__, _, [atom]} = node, zipper, document, position) when is_atom(atom) do
+    parent = Zipper.up(zipper)
+
+    case Zipper.node(parent) do
+      {:__aliases__, _, _} = aliases ->
+        resolve(aliases, parent, document, position)
```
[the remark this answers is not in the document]
scohen: There are three functions in code that I think can help you here, `Code.Fragment.cursor_context`, `Code.Fragment.container_context` and `Code.Fragment.container_context_to_quoted`. 

I think most of this can be accomplished by trimming the string to the position and giving the result to `Code.Fragment.cursor_context` or by using `Code.Fragment.container_context`. Apologies if you've already tried these functions and rejected them, but I'd like to know why.

I played around a bit with `Code.Fragment.container_context` and it was particularly good at resolving modules, popping out an alias which you can use the alias resolver to expand.

apps/server/lib/lexical/server/code_intelligence/entity.ex
```diff
@@ -0,0 +1,240 @@
...
+  end
+
+  defp to_precise_range(text, line, column) do
+    case Code.Fragment.surround_context(text, {line, column}) do
+      %{begin: start_pos, end: end_pos} ->
+        to_range(start_pos, end_pos)
+
+      _ ->
+        # If the column is 1, but the code doesn't start on the first column, which isn't what we want.
+        # The cursor will be placed to the left of the actual definition.
```
scohen, line 227: does it matter if the column is 1 here? Code that's indented might not start on line 1 either

apps/server/lib/lexical/server/provider/handlers/hover.ex
```diff
@@ -0,0 +1,62 @@
...
+  alias Lexical.Server.CodeIntelligence.Entity
+  alias Lexical.Server.Provider.Env
+
+  require Logger
+
+  def handle(%Requests.Hover{} = request, %Env{} = env) do
+    maybe_hover =
+      with {:ok, {:module, module}} <- Entity.resolve(request.document, request.position),
+           {:ok, doc_content} <- module_doc_content(env.project, module) do
+        module_name = module |> to_string() |> String.replace_prefix("Elixir.", "")
```
scohen, line 16: I find we do this a lot in our code. Perhaps we should make a utility for splitting modules (both elixir and erlang) and turning them into strings.

apps/server/lib/lexical/server/provider/handlers/hover.ex
```diff
@@ -0,0 +1,62 @@
...
+        {:ok, doc}
+
+      {:ok, %{module_doc: :none}} ->
+        {:ok, "*This module is undocumented.*\n"}
+
+      {:ok, %{module_doc: :hidden}} ->
+        {:ok, "*This module is private.*\n"}
+
+      {:ok, other} ->
+        {:error, {:unexpected, other}}
```
scohen, line 46: when can this happen?

apps/server/lib/lexical/server/provider/handlers/hover.ex
```diff
@@ -0,0 +1,62 @@
...
+      error ->
+        error
+    end
+  end
+
+  defp fetch_docs(project, module) do
+    with {:docs_v1, _annotation, _lang, _format, module_doc, _meta, docs} <-
+           RemoteControl.Api.docs(project, module) do
+      {:ok, %{module_doc: parse_module_doc(module_doc), docs: docs}}
+    end
```
scohen, line 57: I feel like the API for `RemoteControl.API.docs(project, module)` needs a little help. We should make it our own, and return things we care about.

apps/server/test/lexical/server/code_intelligence/entity_test.exs
```diff
@@ -32,8 +32,13 @@ defmodule Lexical.Server.CodeIntelligence.DefinitionTest do
...
 
-    with :ok <- Document.Store.open(uri, content, 1) do
-      Document.Store.fetch(uri)
+    case Document.Store.open(uri, content, 1) do
+      :ok ->
+        Document.Store.fetch(uri)
+
+      {:error, :already_open} ->
+        :ok = Document.Store.close(uri)
+        subject_module(project, content)
```
scohen, line 41: why not just fetch it rather than closing and recursively calling this function?
```suggestion
        Document.Store.fetch(uri)
```

scohen: Also, this smells a little funny, why are you opening something twice in a unit test?

apps/server/test/lexical/server/code_intelligence/entity_test.exs
scohen: You might want to use `Code.Fragment.surround_context` to pull out the context instead of or in combination with `Code.Fragment.cursor_context`.

apps/server/test/lexical/server/code_intelligence/entity_test.exs
```diff
@@ -299,30 +304,217 @@ defmodule Lexical.Server.CodeIntelligence.DefinitionTest do
...
+      code = ~q[
+        Some.Modul|e.
+        ]
+
+      assert {:error, _} = resolve(project, code)
+    end
+
+    test "returns an error if the cursor is not over an entity", %{project: project} do
+      code = ~q[
+        Beyond.The.End|
```
scohen, line 326: I feel like this should return `Beyond.The.End`. The cursor is right next to a module name. I would expect the `{:error, :not_found}` to happen if there was a space between the end of the `End` module and the cursor.

apps/server/test/lexical/server/code_intelligence/entity_test.exs
```diff
@@ -299,30 +304,217 @@ defmodule Lexical.Server.CodeIntelligence.DefinitionTest do
...
+        ]
+
+      assert {:error, :not_found} = resolve(project, code)
+
+      code = ~q[
+        | Beyond.The.End
+        ]
+
+      assert {:error, :not_found} = resolve(project, code)
```
scohen, line 335: This is a different test, it should be `| Before.The.Beginning`

I really like having a single thing tested in each unit test. It makes failures much easier to debug and fix.

apps/server/test/lexical/server/code_intelligence/entity_test.exs
```diff
@@ -299,30 +304,217 @@ defmodule Lexical.Server.CodeIntelligence.DefinitionTest do
...
+      assert {:ok, {:module, Example.Nested}} = resolve(project, code)
+
+      code = ~q[
+        defmodule Example do
+          __MODULE_|_.Nested
+        end
+        ]
+
+      assert {:ok, {:module, Example}} = resolve(project, code)
+    end
```
scohen, line 488: I like how concise these tests are. Good work on that. I really would prefer having a test block test a single thing though.

apps/server/lib/lexical/server/code_intelligence/entity.ex
scohen: pretty much the only way to fix cyclomatic complexity errors is to pull out functions. I actually think credo calculates CC incorrectly. This doesn't feel crazy to me, but the nested function + case + cond is likely what's bumping up the CC.

apps/server/lib/lexical/server/code_intelligence/entity.ex
zachallaun: iirc, this comment was copied here from another file during some reorganization. I'll look more carefully at this logic though and make sure that whatever special case is being handled makes sense.

apps/server/lib/lexical/server/provider/handlers/hover.ex
zachallaun: Any thoughts on where to put this and what to name the utility module?

apps/server/lib/lexical/server/provider/handlers/hover.ex
zachallaun: This was cruft from when I was experimenting. It's gone now.

apps/common/lib/lexical/ast.ex
```diff
@@ -15,10 +65,13 @@ defmodule Lexical.Ast do
 
   @type t :: any()
 
+  @type cursor_context :: any()
+  @type surround_context :: any()
```
zachallaun, line 69: Note to self: actually write these out. And possibly move the fragment api to `Lexical.Ast.Fragment`

apps/remote_control/lib/lexical/remote_control/code_intelligence/docs.ex
```diff
@@ -0,0 +1,75 @@
...
+  @spec for_module(module()) :: {:ok, t} | {:error, any()}
+  def for_module(module) when is_atom(module) do
+    with :ok <- ensure_ready(module),
+         {:docs_v1, _anno, _lang, _fmt, module_doc, _meta, docs} <- Code.fetch_docs(module) do
+      {:ok, normalize_docs(module, module_doc, docs)}
+    end
+  end
+
+  defp normalize_docs(module, module_doc, element_docs) do
+    elements_by_kind = Enum.group_by(element_docs, &(&1 |> elem(0) |> elem(0)))
```
scohen, line 30: this is a baffling function ;) What are the first elements of the tuples?

apps/server/lib/lexical/server/code_intelligence/entity.ex
```diff
@@ -0,0 +1,142 @@
...
+  alias Lexical.Document.Location
+  alias Lexical.Document.Position
+  alias Lexical.Document.Range
+  alias Lexical.Project
+  alias Lexical.RemoteControl
+  alias Lexical.Text
+
+  require Logger
+
+  @type resolved :: {:module, module()}
```
scohen, line 14: this means we only resolve modules presently, correct?

apps/server/lib/lexical/server/code_intelligence/entity.ex
```diff
@@ -0,0 +1,142 @@
...
+    |> Module.concat()
+    |> Ast.expand_aliases(document, position)
+  end
+
+  @doc """
+  Returns the source location of the entity at the given position in the document.
+  """
+  def definition(%Project{} = project, %Document{} = document, %Position{} = position) do
+    maybe_location = RemoteControl.Api.definition(project, document, position)
+    parse_location(maybe_location, document)
```
scohen, line 99: project
```suggestion
    project
    |> RemoteControl.Api.definition(document, position)
    |> parse_location(document)
```

apps/server/lib/lexical/server/provider/handlers/hover.ex
scohen: I'd put it in common, I'm flexible on the name, `Lexical.Ast.Module` maybe?

apps/server/test/lexical/server/provider/handlers/hover_test.exs
```diff
@@ -0,0 +1,117 @@
...
+    Handlers.Hover.handle(request, %Env{project: project})
+  end
+
+  describe "module hover" do
+    @describetag uri_for: "docs.ex"
+
+    test "module with docs", %{project: project, uri: uri} do
+      # alias Project.Docs.PublicModule
+      #                    ^
+      {:ok, request} = hover_request(uri, 64, 21)
```
scohen, line 64: 🤔 I would like to think about tests like this a little more; Tests with hard-coded magic numbers (here referring to files in a different place) are extremely brittle and difficult to debug if they fail.

Is there a way we can build a file on the fly and annotate it inside the tests as we do with the `CompletionCase`?

scohen (commented):
Wow, what an improvement. I'm really liking the direction here (and I can't wait until we can get `Ast.fragment` merged so I can use it in my indexer PR.

I would like to call out the tests though, I think (hope) we can improve them so that each test clearly shows both the module being defined (and its docs, or lack thereof) and the module being hovered over.

apps/common/lib/lexical/ast.ex
```diff
@@ -15,10 +65,13 @@ defmodule Lexical.Ast do
 
   @type t :: any()
```
scottming, line 66: Can we use another `type` instead of `any`?

apps/common/lib/lexical/ast.ex
```diff
@@ -42,19 +95,98 @@ defmodule Lexical.Ast do
...
+    document
+    |> Document.fragment(position)
+    |> cursor_context()
+  end
+
+  @doc """
+  Returns the cursor context of the fragment.
+  """
+  @spec cursor_context(String.t()) :: {:ok, cursor_context()} | {:error, :cursor_context}
+  def cursor_context(fragment) when is_binary(fragment) do
```
scottming, line 150: the variable name is a fragment, but its type is string, which feels strange because the fragment function above returns t

apps/common/lib/lexical/ast.ex
```diff
@@ -42,19 +95,98 @@ defmodule Lexical.Ast do
...
+  Returns the surround context of the document at a position.
+  """
+  @spec surround_context(
+          Document.t() | String.t(),
+          Position.t() | {Position.line(), Position.character()}
+        ) ::
+          {:ok, surround_context()} | {:error, :surround_context}
+  def surround_context(%Document{} = document, %Position{} = position) do
+    %{line: line, character: column} = position
+    surround_context(document, {line, column})
```
scottming, line 164: I prefer to convert the document to string here, then we can avoid this recursive call.

apps/common/lib/lexical/ast.ex
```diff
@@ -42,19 +95,98 @@ defmodule Lexical.Ast do
...
+  end
+
+  def surround_context(string, %Position{} = position) do
+    %{line: line, character: column} = position
+    surround_context(string, {line, column})
```
scottming, line 169: I think `do_surround_context(string, {line, column})` is enough, right?

scottming (changes requested):

apps/common/lib/lexical/ast.ex
zachallaun: Yeah, maybe we should just use `Macro.t()` directly instead and remove `Ast.t()`. That's what Sourceror does, for instance, which is a good example to follow. WDYT?

apps/common/lib/lexical/ast.ex
zachallaun: Good points. I was copying the naming from `Code.Fragment.cursor_context`, but I agree that it's confusing given the presence of another `fragment` function. I'm inclined to just rename to `s` or `string`.

apps/common/lib/lexical/ast.ex
zachallaun: Will address this and the other case below.

apps/remote_control/lib/lexical/remote_control/code_intelligence/docs.ex
zachallaun: Hah! I meant to add a comment, but I actually thing I should just be a little more verbose here in the interest of clarity.

```elixir 
elements_by_kind =
  Enum.group_by(element_docs, fn {{kind, _name, _arity}, ...} ->
    kind
  end)
```

apps/server/lib/lexical/server/code_intelligence/entity.ex
zachallaun: Correct. The idea is that this will (quickly) expand as other entities are supported.

apps/server/test/lexical/server/provider/handlers/hover_test.exs
zachallaun: Agreed. When I first wrote this test file, I started by copying the go_to_definition_test.exs and then adapting it, because I wasn't yet familiar with what was required to setup the test environment. 

Now that I have a better understanding, I'll take another pass at this.

apps/common/lib/lexical/ast.ex
scottming: I think `Macro.t()` is good.

zachallaun:
> I noticed that this PR also supports built-in Elixir modules like Enum, so I think we might need to add support for Hexdocs links.

Those aren't embedded in the module docs, but added after the fact. I do think it is worth adding them (Elixir LS does as well), but I'd rather that happen in another PR.

> Additionally, I found that when the cursor is on a dot, it displays the documentation for the previous module, which is great. However, it doesn't work when there are no functions or submodules following it, like Lexical.|, but Lexical.|RemoteControl will work.

Will look into it!

apps/common/lib/lexical/ast.ex
zachallaun: I'm actually not sure how/whether I want to handle this. It feels a little silly to fully duplicate the return types from `Code.Fragment` (especially since they define the return fully in the function spec, not as separate types). I'm sort of more inclined to just document these as representing the return values from those functions. 

```elixir
@typedoc "Return value from `Code.Fragment.cursor_context/2`"
@type cursor_context :: any()
```

zachallaun: @scohen @scottming WDYT?

scohen: I don't have a strong opinion, though I shy away from `any()` by default. I think the typedoc clears it up.

scottming: I'm on the same page as steve, I think typedoc is clear enough here.

apps/server/test/lexical/server/code_intelligence/entity_test.exs
```diff
@@ -299,30 +299,249 @@ defmodule Lexical.Server.CodeIntelligence.DefinitionTest do
...
+      assert {:error, :not_found} = resolve(project, code)
+    end
+
+    test "resolves module segments at and before the cursor", %{project: project} do
+      code = ~q[
+        In.|The.Middle
+      ]
+
+      assert {:ok, {:module, In.The}, range} = resolve(project, code)
+      assert range.start.line == 1
```
scottming, line 357: To be honest, I prefer the range assert pattern as the same as the definition** used, rather than specific lines and characters. This allows us to see what we are resolving clearly.

assert resolved_range = "In.«The».Middle"

@scohen What do you think?

zachallaun: Found the existing usage [here](https://github.com/lexical-lsp/lexical/blob/d9dcae41e03254f589a7c44e058013fe098043b4/apps/server/test/lexical/server/code_intelligence/definition_test.exs#L258) -- I’ll happily extract this and use it in these tests.

zachallaun:
@scohen @scottming I believe everything that's been mentioned so far has been addressed. Please let me know what else y'all would like to see prior to merge!

apps/server/lib/lexical/server/code_intelligence/entity.ex
```diff
@@ -0,0 +1,152 @@
...
+  Returns `{:ok, resolved, range}` if successful and `{:error, error}`
+  otherwise. The `range` includes the resolved node and the
+  Resolved entities are one of:
+
+    * `{:module, module}`
+
+  """
+  @spec resolve(Document.t(), Position.t()) :: {:ok, resolved, Range.t()} | {:error, term()}
+  def resolve(%Document{} = document, %Position{} = position) do
+    with {:ok, %{context: context, begin: begin_pos, end: end_pos}} <-
```
scohen, line 30: can ast return anything other than this? I think matching out all the pieces is a little noisy. 
I'm also wondering if we should wrap the map in a struct. I really really dislike maps as data structures. You have no idea what's in them.

Maybe this is for a subsequent review.

scohen (approved):

apps/server/lib/lexical/server/code_intelligence/entity.ex
zachallaun: > can ast return anything other than this?

It can also return `{:ok, :none}` or `{:error, parse_error}`.

Right now it's a pretty thin wrapper over `Code.Fragment.surround_context/3`. We can play with the return value, however. Instead of a struct, I'd be more inclined to return `{:ok, context, range}` and turn the `:none` case into an `{:error, ...}` of some kind. But, doing that, we may want to change the name of the function as well so that there isn't an expectation that it'll return the same thing as `Code.Fragment.surround_context`.

scottming (approved):

scohen (approved):

apps/server/lib/lexical/server/code_intelligence/entity.ex
zachallaun: Will address this in followup PR.

Issue #324, opened 2023-08-13
bug: `could not load module :future_elixir_interpolation`


chukinas:
I tried creating a minimal neovim setup using https://github.com/scottming/nvim-mini-for-lexical/, but when I open an elixir file, I get the error `Client 1 quit with exit code 1 and signal 0`.

Is there any way to debug or see logs so I can see more detail around what's going on? I'm not even sure at this point if that error is return by the language server or something in neovim.

chukinas:
Oh cool, so I found the `:LspLog` command and found this error in the neovim LSP logs:
```
[START][2023-08-13 08:46:07] LSP logging initiated
[ERROR][2023-08-13 08:46:07] .../vim/lsp/rpc.lua:734    "rpc"   "/home/jc/projects/opensource/lexical/_build/dev/package/lexical/bin/start_lexical.sh"  "stderr"        '** (Argume
ntError) could not load module :future_elixir_interpolation due to reason :badfile\n    (elixir 1.15.4) lib/code.ex:1779: Code.ensure_loaded!/1\n    (elixir 1.15.4) lib/enum.ex:98
4: Enum."-each/2-lists^foreach/1-0-"/2\n    lib/lexical/server/boot.ex:17: LXical.Server.Boot.start/0\n    nofile:1: (file)\n    (stdlib 4.0) erl_eval.erl:748: :erl_eval.do_apply/
7\n'
```

I saw something about this when I was following the Lexical setup steps. I'll go back to those steps and see what I can learn.

zachallaun:
I ran into the same thing yesterday, specifically when switching the OTP version that I was packaging lexical with. In my case, it was fixed by deleting my `_build` directory (in lexical) and recompiling/packaging. Hope the same works for you to get things up and running!

Would of course be good to figure out what's going on here -- maybe `mix package` needs to be forcing compilation to ensure that old build artifacts aren't being used?

scohen:
@chukinas, Did you build on a newer version of erlang than your project uses?

scohen:
@chukinas Did you get a chance to try the branch in #332?

scohen:
any updates, @chukinas ?

scohen:
@chukinas I'm closing due to non-response, and am assuming it works for you now.

PR #332, opened 2023-08-14
Added version compatibility check

### LINKED ISSUE #328: Tag build artifacts
I've encountered some problems when switching between versions of erlang / elixir (though mostly erlang). If an artifact is compiled on a later version, it will not work inside an earlier version of the VM. This is to be expected, but it's clear that we need to tag our artifacts with the VM version they were built on. This way, we can ensure that we re-build all of our artifacts if the VM changes. 

This occurs in two places. First, inside the artifacts created during the package task, and secondly in the `.lexical/build` directory.

If a mismatch is detected in `.lexical/build`, we should delete all artifacts and rebuild there. 
If a mismatch is detected in the package, the VM should exit and emit a warning to the user to recompile on a different version.

scohen:
Lexical now outputs the erlang and elixir versions of the VM that was used to create them. It now checks the build and packaging directories for these versioning files and emits an error message if the vm starts with beam files that won't work.

Fixes #328 and #324

### DIFF
--- a/apps/common/lib/lexical/vm/versions.ex
+++ b/apps/common/lib/lexical/vm/versions.ex
@@ -1,0 +1,222 @@
+defmodule Lexical.VM.Versions do
+  @moduledoc """
+  Reads and writes version tags for elixir and erlang
+
+  When compiling, it is important to node which version of the VM and elixir runtime
+  were used to build the beam files, as beam files compiled on a newer version of the
+  VM cannot be used on older versions.
+
+  This module allows a directory to be tagged with the versions of elixir and erlang
+  used as compilation artifacts, and also allows the user to ask if a certain version
+  is compatible with the currently running VM.
+  """
+
+  @type version_string :: String.t()
+
+  @type t :: %{elixir: version_string(), erlang: version_string()}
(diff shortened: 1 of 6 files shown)


scottming:
```elixir
~/Co/lexical tagged_artifacts *1 ❯ iex -S mix                                                    46s  25.3.2.4  1.15.4-otp-25
Erlang/OTP 25 [erts-13.2.2.2] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit:ns]

** (Mix) Could not start application server: exited in: Lexical.Server.Application.start(:normal, [])
    ** (EXIT) an exception was raised:
        ** (MatchError) no match of right hand side value: :error
            (server 0.1.0) lib/lexical/server/application.ex:33: Lexical.Server.Application.check_version/0
            (server 0.1.0) lib/lexical/server/application.ex:16: Lexical.Server.Application.start/2
            (kernel 8.5.4.1) application_master.erl:293: :application_master.start_it_old/4
            ```
            
            After `mix package`, then run `iex -S mix` will encounter this error.

scottming (approved):

scottming:
@scohen yeah, you're right, I tried `mix clean` first, then everything works good.

PR #336, opened 2023-08-16
Context-aware "use" completions

### LINKED ISSUE #330: Improved `use` completions
We should enhance the intelligence around `use |` to only show modules that have a `__using__` macro.

scohen:
When using `use`, we would complete any module, even those without `__using__` macros defined.

This PR filters any suggestions inside a `use` directive to ensure that any suggested modules (or a submodule) contain a __using__ macro of arity 1.

One thing that cropped up is that not all modules are loaded when queried, so I had to use `:code.all_available` to generate the module list. This wasn't super duper fast, so I had to put a cache in front of it, which is the large part of this PR.

Another oddity is that I couldn't just send anonymous functions as predicates, so I had to come up with a way to implement filtering logic across distribution, so I did a matchspec-like thingy, which was a bit ugly, so I made some syntax to make it look like function captures.

Fixes #330

### DIFF
--- a/apps/server/lib/lexical/server/code_intelligence/completion.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion.ex
@@ -195,7 +195,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
               RemoteControl.Api.modules_with_prefix(
                 env.project,
                 full_name,
-                predicate(&macro_exported?(:"$1", :__using__, 1))
+                predicate(&macro_exported?(&1, :__using__, 1))
               )
 
             not Enum.empty?(with_prefix)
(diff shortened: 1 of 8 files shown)


scohen:
Note: this is intended to land _after_ we get 0.3 out.

apps/server/lib/lexical/server/code_intelligence/completion.ex
```diff
@@ -186,6 +188,22 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
...
         ]
 
+      Env.in_context?(env, :use) ->
+        case result do
+          %{full_name: full_name} ->
+            with_prefix =
+              RemoteControl.Api.modules_with_prefix(
+                env.project,
+                full_name,
+                predicate(&macro_exported?(:"$1", :__using__, 1))
```
scottming, line 198: I like this API.

scottming (approved):

scottming:
I tried this branch, works great.

BTW, Perhaps it is unrelated to this PR, but should we let `use` to complete `use |`, removing the existing two `use` completions?

I mean: `use(module)` and `use(module, opts)`

CleanShot 2023-08-16 at 20 24 07@2x

apps/server/lib/lexical/server/code_intelligence/completion.ex
scohen: Heh, I'd much rather use a predicate function

scohen:
Agree with your screenshot, we should emit a snippet

PR #338, opened 2023-08-16
Suggest a module name for defmodule completion

scohen:
The defmodule completion snippet now does a best-effort attempt to guess the module name.

The guessing takes into account lib, test and test/support directories, and if it doesn't detect any of them, just uses the file name.

### DIFF
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
@@ -1,4 +1,5 @@
 defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
+  alias Lexical.Document
   alias Lexical.RemoteControl.Completion.Candidate
   alias Lexical.Server.CodeIntelligence.Completion.Env
   alias Lexical.Server.CodeIntelligence.Completion.Translatable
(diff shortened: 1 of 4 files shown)


scohen:
@scottming, thoughts?

apps/server/test/lexical/server/code_intelligence/completion/translations/macro_test.exs
```diff
@@ -118,7 +118,58 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.MacroTest do
...
+             """
+    end
+
+    test "defmodule for test paths", %{project: project} do
+      assert {:ok, completion} =
+               project
+               |> complete("defmodule|", path: "test/foo/project/my_test.exs")
+               |> fetch_completion("defmodule ")
+
+      assert completion.detail
```
scottming, line 133: I found there is no detail here, Just like the struct completions, if we put the `module` name to the detail. that would be helpful.

scottming (approved):

apps/server/test/lexical/server/code_intelligence/completion/translations/macro_test.exs
scohen: Do you need the detail? The snippet shows the suggested module name, and allows you to change it by typing. Also, nested module names can be quite long, which won't look good in the completion ui

scottming: I see, I'm fine without `detail`

apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
```diff
@@ -50,11 +51,15 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
     |> builder.boost(8)
   end
 
+  require Logger
+
   def translate(%Candidate.Macro{name: "defmodule"} = macro, builder, env) do
     label = "defmodule (Define a module)"
+    suggestion = suggest_module_name(env.document)
+    Logger.info("Suggestion is #{suggestion}")
```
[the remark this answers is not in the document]
scohen: Nope!

PR #339, opened 2023-08-16
Improved launcher script

scohen:
Seems ERL_LIBS wasn't picked up universally across versions, so I swapped that out for the -pa flag

### DIFF
--- a/bin/start_lexical.sh
+++ b/bin/start_lexical.sh
@@ -1,15 +1,14 @@
 #!/usr/bin/env bash
 
 set_up_version_manager() {
-    if asdf which elixir > /dev/null -eq 0; then
+    if command -v asdf > /dev/null && asdf which elixir > /dev/null -eq 0; then
         VERSION_MANAGER="asdf"
-    elif rtx which elixir > /dev/null -eq 0; then
+    elif command -v rtx > /dev/null &&  rtx which elixir > /dev/null -eq 0; then
         VERSION_MANAGER="rtx"
     else
         VERSION_MANAGER="none"
     fi
 }
-
 
 set_up_version_manager
 
--- a/bin/start_lexical.sh
+++ b/bin/start_lexical.sh
@@ -18,7 +18,7 @@ case "$VERSION_MANAGER" in
         asdf env erl exec "$@" &
         ;;
     rtx)
-        eval "$(rtx env -s bash erl)"
+        eval "$(rtx env -s bash)"
         ;;
     *)
         ;;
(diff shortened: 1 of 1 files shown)


bin/start_lexical.sh
```diff
@@ -1,16 +1,15 @@
 #!/usr/bin/env bash
 
 set_up_version_manager() {
-    if [ -e $HOME/.asdf ] &&  asdf which elixir > /dev/null -eq 0; then
+    if command -v asdf > /dev/null && asdf which elixir > /dev/null -eq 0; then
         VERSION_MANAGER="asdf"
-    elif [ -e $HOME/.rtx ] && rtx which elixir > /dev/null -eq 0; then
+    elif command -v rtx > /dev/null &&  rtx which elixir > /dev/null -eq 0; then
```
Blond11516, line 6: So trying this out made me realize rtx detection didn't even work because there is no `.rtx` folder in the home directory. It was just using the globally defined versions.

Now, this version doesn't quite work either. Here's first part of a fix:
```suggestion
    elif command -v rtx > /dev/null &&  rtx which elixir > /dev/null 2>&1; then
```
The current version fails for me because `-e` is passed as an argument to `rtx which`, which obviously doesn't work.

After that, the "Start the program in the background" case breaks things because `rtx env` doesn't do the same thing as `asdf env` it seems: it just outputs env vars for the command passed in (`erl` in this case). This output causes the language server to fail to start according to vscode, I'm guessing because of the unexpected content in stdout.

The thing is, I'm not sure how to fix it because I don't even understand what this part of the script does. From what I understand, it's trying to run the command passed in as argument (which is nothing in the normal case) in the asdf/rtx env in the background? Why are we even trying to do that? We're running the command on line 44, and the fact we don't do anything if there's no version manager confuses me even more...

Sorry for the long comment, it's late and I'm tired but I want to write this down before I forget half of it tomorrow morning.

bin/start_lexical.sh
```diff
@@ -19,7 +18,7 @@ case "$VERSION_MANAGER" in
         asdf env erl exec "$@" &
         ;;
     rtx)
-        rtx env -s bash erl exec "$@" &
+        eval "$(rtx env -s bash erl)"
```
Blond11516, line 21:
```suggestion
        eval "$(rtx env -s bash)"
```
`rtx env`'s argument serves to override a given tool with a specific version. `erl` does nothing in this case

Blond11516 (approved):
Seems to work fine on my machine 👍

PR #343, opened 2023-08-23
Add Vanilla Emacs with eglot instruction

dalugm:
Eglot has a built-in checking mechanism where the first item in each sub-list of eglot-server-programs cannot be repeated. Which means, when Eglot already has a elixir-mode cons, users add a new entry will lead to an error. So we had to change the original list. So I don't know whether we should keep the elixir-ls's command.

If you have any ideas, please let me know.


scohen (approved):

scohen:
Thank you, this is very helpful!

PR #345, opened 2023-08-24
Add elixir boot script to support having spaces in the package path

Blond11516:
The start_lexical.sh launch script didn't properly handle having spaces in the path to the package; it would fail to construct paths to the erlang archives and add them to the erlang path. Properly handling spaces in the bash script is fairly painful because it requires careful escaping to prevent paths with spaces from being interpreted as two separate arguments.

This moves this work to an elixir script, where there is no need to manage escaping spaces in paths (because they're strings).

With an exs script, we don't need any of Lexical's code to be available prior. In fact, we could even use this script to add `consolidated`, `config` and `priv` to the path if we wanted.

## Testing in vscode auto-install-like conditions

To make sure this works with spaces in the path, I package lexical and manually copy the package to vscode's global storage (path is for MacOS):
```bash
mix package
cp -r _build/dev/package/* "/Users/[YOUR_USER]/Library/Application Support/Code/User/globalStorage/lexical-lsp.lexical/lexical_install"
```

Then restart vscode to restart the extension and see if Lexical starts properly


apps/server/priv/boot.exs
```diff
@@ -0,0 +1,17 @@
+[script_dir] = System.argv()
+
+[script_dir, "..", "lib", "*.ez"]
+|> Path.join()
+|> Path.wildcard()
+|> Enum.each(fn archive_path ->
+  lib =
+    archive_path
+    |> Path.basename()
+    |> String.replace_suffix(".ez", "")
```
scohen, line 10: does this work? I think you need the whole file with the `.ez` extension and everything.

apps/server/priv/boot.exs
```diff
@@ -0,0 +1,17 @@
+[script_dir] = System.argv()
```
scohen, line 1: maybe we just put this file in `bin` at the top level and add that to the code path?

apps/server/priv/boot.exs
```diff
@@ -0,0 +1,17 @@
...
+
+[script_dir, "..", "lib", "*.ez"]
+|> Path.join()
+|> Path.wildcard()
+|> Enum.each(fn archive_path ->
```
scohen, line 6: i _think_ this can be accomplished by 
```elixir
[System.cwd!(), "lib", "*.ez"]
|> Path.join()
|> Path.wildcard()
|> Enum.map(&Path.join(&1, "ebin"))
|> :code.add_patha()
```

scohen: This will require adding the bin directory via `-pa`

bin/start_lexical.sh
```diff
@@ -35,16 +35,10 @@ esac
...
-                    -pa "${SCRIPT_DIR}/../consolidated" \
-                    -pa "${SCRIPT_DIR}/../config/" \
-                    -pa "${SCRIPT_DIR}/../priv/" \
-                    --eval "LXical.Server.Boot.start" \
-                    --cookie "lexical" \
-                    --no-halt
+"${ELIXIR_COMMAND}" \
+    -pa "${SCRIPT_DIR}/../consolidated" \
+    -pa "${SCRIPT_DIR}/../config/" \
+    -pa "${SCRIPT_DIR}/../priv/" \
```
scohen, line 41: do we still need to set these via `-pa` or can we use `:code.add_patha?`?

scohen: looks like you already answered that in the comment. Let's add them in the script then.

scohen:
This should go in the top-level `bin` directory rather than mucking about with `priv`.

scohen:
I'm a big fan of this approach. Much better than wrangling with bash.

apps/server/priv/boot.exs
Blond11516: As discussed on Discord, the `.ez` part is still contained in `archive_path`, so it gets included in the ebin path starting on line 12

apps/server/priv/boot.exs
Blond11516: This doesn't work because we need the name of the lib betwen `.ez` and `ebin`

Blond11516:
I've added instructions to test with spaces in the path as if lexical was auto-installed by vscode to the PR description.

scohen (approved):

PR #347, opened 2023-08-25
Indexing features for modules

scohen:
This commit introduces indexing for module references. It currently will only work for files to which we have the source code, turns them into AST and then applies one or more "extractors" on it. An extractor runs over AST and indetifies things it's looking for, and then emits entries, which will be used to build a searchable index.

Currently, there is one extractor implemented: Module references. This does a fairly exhaustive search for module references, wherever they're hiding. It then uses the aliases module to find the original source module.

Note that most of the higher level functions are just placeholders. An indexing strategy needs to find things inside of .exs files as well as .ex files, and we currently have no way to index core elixir (this will be highly dependent on how elixir was installed and if we can get the source code to the version we're running on).


scohen:
@zachallaun ☝️

apps/remote_control/lib/lexical/remote_control/search/indexer/module.ex
```diff
@@ -0,0 +1,24 @@
...
+
+  def index(module) do
+    with true <- indexable?(module),
+         {:ok, path, source} <- source_file_path(module) do
+      Indexer.Source.index(path, source)
+    end
+  end
+
+  def source_file_path(module) do
+    with {:ok, %{file: file_path}} <- BeamFile.debug_info(module),
```
zachallaun, line 12: Playing around with this locally, this `debug_info` call takes quite a long time. If all we care about is the source path and contents, this is available on the module directly using `__info__(:compile)`:

```elixir
iex(1)> Lexical.RemoteControl.__info__(:compile)
[
  version: '8.2.6.3',
  options: [:no_spawn_compiler_process, :from_core, :no_core_prepare,
   :no_auto_import],
  source: '/home/z/dev/lexical/apps/remote_control/lib/lexical/remote_control.ex'
]
```

Would also make `BeamFile` dep unnecessary.

apps/remote_control/lib/lexical/remote_control/search/indexer/source.ex
```diff
@@ -0,0 +1,33 @@
...
+  alias Lexical.RemoteControl.Search.Indexer.Source.Reducer
+
+  require Logger
+
+  @to_quoted_opts [columns: true, token_metadata: true]
+
+  def index(path, source) do
+    document = Document.new(path, source, 1)
+
+    case Code.string_to_quoted(source, @to_quoted_opts) do
```
zachallaun, line 13: Could use `Ast.from(source)` here to not have to specify the `@to_quoted_opts`.

zachallaun (commented):
Didn't quite finish reading through all of this, but made a few comments on things that I noticed.

apps/remote_control/lib/lexical/remote_control/search/indexer/module.ex
scohen: fantastic. I was originally using BeamFile as a way to reconstitute the source code from certain beam files, but that didn't pan out (the code it generated was definitely incorrect). I'll swap these out.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
```diff
@@ -0,0 +1,166 @@
...
+      :error ->
+        :ignored
+    end
+  end
+
+  def extract(_, _) do
+    :ignored
+  end
+
+  defp resolve_alias(%Reducer{} = reducer, unresolved_alias) do
```
zachallaun, line 106: While it's using `at_ast` now, this function is still extremely similar to [`Ast.expand_alises`](https://github.com/lexical-lsp/lexical/blob/b25b988c074de13279027bb646408fa218f0d19e/apps/common/lib/lexical/ast.ex#L221). From a contributor's perspective, it's a lot easier to understand existing code and implement new features when common APIs are used -- can we either modify `expand_aliases` to use `at_ast` or create a `expand_alises_at_ast` that can be used instead of this function?

apps/common/lib/lexical/ast.ex
```diff
@@ -55,6 +55,49 @@ defmodule Lexical.Ast do
...
+              current_line > end_line ->
+                {:halt, zipper, acc}
+
+              true ->
+                {_element, new_acc} = prewalk_fn.(element, acc)
+                {:cont, zipper, new_acc}
+            end
+
+          element ->
+            {_element, new_acc} = prewalk_fn.(element, acc)
```
zachallaun, line 93: It's a little odd that the element returned from `prewalk_fn` is ignored in each branch here. I'd say the prewalk function should either only return the new accumulator, or the zipper can be updated in-place with the returned element. 

```elixir
{new_element, new_acc} = prewalk_fn.(element, acc)
new_zipper = Zipper.replace(zipper, new_element)
{:cont, new_zipper, new_acc}
```

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
scohen: I saw that you did some work in this area in your other PR, so I kept this here until that lands. `Aliases.at` does things a little differently than `Ast` (it uses `Code.Fragment`, for example), and I'd like to unify the two.

apps/common/lib/lexical/ast.ex
scohen: agreed, you're seeing some evolution from using `Macro.prewalk` to zippers. I hadn't decided what I wanted to do yet.  I'll remove.

apps/remote_control/lib/lexical/remote_control/search/indexer/source/reducer.ex
```diff
@@ -0,0 +1,145 @@
...
+
+  defp do_reduce(%__MODULE__{} = reducer, element) do
+    case Metadata.location(element) do
+      {:block, position, block_start, block_end} ->
+        block = Block.new(block_start, block_end)
+
+        reducer
+        |> update_position(position)
+        |> maybe_pop_block()
+        |> push_block(block)
```
scottming, line 58: I haven't fully read through this PR, but I feel that the core should be this module. Here, I don't quite understand why we initially pop a block and then push it back (with a parent ref added).

I would like to know the reason behind this. Could you provide an example to help me better understand it?

scohen: Sure. This call happens when we encounter a new block. We're _maybe_ popping the existing block, and then pushing the new block. This allows us to differentiate between:
```elixir
defmodule First do 
end
defmodule Second do <position> 
end
```
and 
```elixir
defmodule Parent do 
  defmodule Child do <position>
  end 
end
```

In the first, `maybe_pop_block` pops the block representing the `First` module, and in the second, `maybe_pop_block` doesn't do anything, keeping the parent block in-scope and child block is pushed.

Blond11516 (approved):
Most of the AST stuff is lost on me (I need to brush up on that) but the overall approach looks good.

I noticed the indexer doesn't appear to use any concurrency at the moment; is that on purpose? Unless I'm mistaken, parallelizing around files should be transparent to the rest of the indexing process?

There's obviously the issue of choosing how many concurrent processes to use, which would ideally depend on the available cores. Probably not worth doing now to keep this PR focused on the core algorithm, was just wondering if you'd given it some thought.

scohen:
> I noticed the indexer doesn't appear to use any concurrency at the moment; is that on purpose? Unless I'm mistaken, parallelizing around files should be transparent to the rest of the indexing process?

Yes, that's intentional. 
1. Make it work
2. Make it pretty
3. Make it fast
 
Still working on #1.

scohen:
BTW, this is a draft PR, I'd like feedback on the overall approach

apps/remote_control/lib/lexical/remote_control/search/indexer.ex
```diff
@@ -0,0 +1,40 @@
+defmodule Lexical.RemoteControl.Search.Indexer do
+  alias Lexical.Project
+  alias Lexical.RemoteControl.Search.Indexer
+
+  def index(%Project{} = project) do
```
Blond11516, line 5: What's the use case for this function? Won't all modules be available anyway after compilation?

apps/remote_control/lib/lexical/remote_control/search/indexer/source/reducer.ex
```diff
@@ -0,0 +1,145 @@
...
+
+  defstruct [:entries, :document, :quoted_document, :position, :ends_at, :blocks]
+
+  @extractors [Extractors.Module]
+
+  def new(%Document{} = document, quoted_document) do
+    %__MODULE__{
+      document: document,
+      quoted_document: quoted_document,
+      entries: [],
```
Blond11516, line 22: Will a list be able to elegantly support incremental indexing (when documents change)? We'll need to find and remove entries from the list on top of adding new ones, so I'm not quite sure a flat list is going to scale very far.

The first alternative that comes to mind is a map, though finding the right key for that might be a challenge.

I know we're not at the optimization phase yet, but since this is kind of the core of the system it'd be nice not to paint ourselves into a corner.

zachallaun: I may be mistaken, but I take this to be a sort of basic "collect all the things", not the final storage mechanism.

apps/remote_control/lib/lexical/remote_control/search/indexer.ex
scohen: So no, nothing in the test directories, nor any .exs files will be available.

apps/remote_control/lib/lexical/remote_control/search/indexer/source/reducer.ex
scohen: That's correct, @zachallaun. 
I'm currently experimenting with ETS and sqlite as storage mechanisms for the on-disk index, and yes, incremental indexing is very important.

apps/remote_control/lib/lexical/remote_control/search/indexer.ex
Blond11516: That's true

apps/remote_control/lib/lexical/remote_control/search/indexer/source/reducer.ex
Blond11516: Ok so the reducer would only built a list of entries for a single indexing pass, and then those entries will get passed somewhere else for storage in a search-friendly format. Makes sense!

scohen: That's correct, @Blond11516. Check out my latest commit for where it all goes.

apps/common/lib/lexical/ast.ex
```diff
@@ -55,6 +55,50 @@ defmodule Lexical.Ast do
     end
   end
 
+  @doc """
+  Traverses the given ast until the given position.
+  """
+  def traverse_until(ast, acc, prewalk_fn, %Position{} = position) do
```
zachallaun, line 61: The first time I saw this function, it wasn't clear to me that it would only turn the accumulator, since `traverse*` functions usually return `{thing_being_traversed, acc}`. Adding a spec would help, but maybe it's also worth calling this `walk_until`? Just to break the expectation compared to other traversal functions.

scohen: yeah, but it's using `Zipper.traverse` `walk` is pretty generic, the elixir functions specify `prewalk` or `postwalk`. 

IIRC you made a suggestion earlier to only return the accumulator, prior I was returning both the element and the accumulator.

zachallaun: > IIRC you made a suggestion earlier to only return the accumulator, prior I was returning both the element and the accumulator.

How dare you make this about _me_!

😉 You're almost right though, but my earlier suggestion was to change the signature of the passed function since we were ignoring the returned node, e.g.  `{_node, acc} = fun.(node)`. Maybe instead we should not ignore it and update the zipper using `Zipper.update/2`.

apps/remote_control/lib/lexical/remote_control/search/indexer.ex
```diff
@@ -2,39 +2,99 @@ defmodule Lexical.RemoteControl.Search.Indexer do
...
+
+  # Note: I tried doing something very similar using
+  # Task.async_stream, but it was barely faster than just
+  # doing things with Enum (16s vs 20s on my M2 max). This
+  # function takes around 4 seconds to reindex all Lexical's
+  # modules, making it 5x faster than Enum and 4x faster than
+  # Task.async_stream
+  defp async_chunks(data, processor, timeout \\ 20_000) do
+    data
+    |> Stream.chunk_every(round(System.schedulers_online() / 2))
```
scohen, line 68: Note: remove the halving.

apps/remote_control/lib/lexical/remote_control/search/fuzzy.ex
```diff
@@ -0,0 +1,163 @@
...
+
+    %__MODULE__{
+      fuzzy
+      | subject_to_refs: subject_to_refs,
+        path_to_refs: path_to_refs,
+        preprocessed_subjects: preprocessed_subjects
+    }
+  end
+
+  def update(%__MODULE__{} = fuzzy, entries) do
```
Blond11516, line 144: I think this would be better off as another head of `add/2` or `add_list/2` if we want to keep them separate since there's really no difference in outcome between the two.

apps/remote_control/lib/lexical/remote_control/search/fuzzy/scorer.ex
```diff
@@ -0,0 +1,189 @@
+defmodule Lexical.RemoteControl.Search.Fuzzy.Scorer do
```
Blond11516, line 1: This whole module is super interesting! It'll also be interesting to see how it performs in the real world and how we can tune and adapt it over time.

apps/remote_control/test/lexical/remote_control/search/fuzzy/scorer_test.exs
```diff
@@ -0,0 +1,62 @@
...
+               score_and_sort("Enumerable", ~w(E Enum Enumerab Enumera))
+    end
+
+    test "patterns that match the start of a string win" do
+      assert ~w(Enum nume era ble) = score_and_sort("Enumerable", ~w(ble era nume Enum))
+    end
+
+    test "patterns that match consecutive characters win" do
+      assert ~w(Enum erb nml) = score_and_sort("Enumerable", ~w(Enum erb nml))
+    end
```
Blond11516, line 55:
```suggestion
    test "patterns that match consecutive characters win" do
      assert ~w(mer erb nml) = score_and_sort("Enumerable", ~w(erb nml mer))
    end
```
`Enum` matches both the most consecutive characters and the beginning of the subject, making the heuristic under test ambiguous.

apps/remote_control/test/lexical/remote_control/search/fuzzy_test.exs
```diff
@@ -0,0 +1,60 @@
...
+  describe "housekeeping" do
+    test "it can add an entry", %{fuzzy: fuzzy} do
+      refute Fuzzy.has_subject?(fuzzy, Other)
+      entry = reference(subject: Other)
+
+      fuzzy = Fuzzy.add(fuzzy, entry)
+      assert Fuzzy.has_subject?(fuzzy, Other)
+    end
+
+    test "a refernce can be removed", %{fuzzy: fuzzy, entries: [to_remove | _]} do
```
Blond11516, line 30:
```suggestion
    test "a reference can be removed", %{fuzzy: fuzzy, entries: [to_remove | _]} do
```

Blond11516 (commented):
I've only read the fuzzy and scorer modules so far, but I think they're definitely on the right track! This will be a very useful feature for lexical.

apps/remote_control/lib/lexical/remote_control/search/fuzzy/scorer.ex
scohen: It performs pretty well right now. I can search over all of the module references (~50,000) in around 30ms.

scohen:
@Blond11516 please re-review; the unit tests for scorer were actually backwards.

apps/remote_control/lib/lexical/remote_control/search/fuzzy.ex
```diff
@@ -0,0 +1,163 @@
...
+          {score, references}
+
+        :error ->
+          nil
+      end
+    end)
+    |> Stream.reject(&is_nil/1)
+    |> Enum.to_list()
+    |> List.keysort(0)
+    |> Enum.flat_map(&elem(&1, 1))
```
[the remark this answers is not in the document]
scohen: I don't think so.

apps/remote_control/lib/lexical/remote_control/search/fuzzy.ex
scohen: add_all?

apps/remote_control/test/lexical/remote_control/search/fuzzy/scorer_test.exs
scohen: These tests are actually backwards, we should be matching agains multiple subjects, not multiple patterns.

apps/remote_control/lib/lexical/remote_control/search/fuzzy/scorer.ex
scohen: it also builds and updates quickly.

apps/remote_control/lib/lexical/remote_control/search/indexer.ex
```diff
@@ -0,0 +1,100 @@
...
+        []
+    end
+  end
+
+  # Note: I tried doing something very similar using
+  # Task.async_stream, but it was barely faster than just
+  # doing things with Enum (16s vs 20s on my M2 max). This
+  # function takes around 4 seconds to reindex all Lexical's
+  # modules, making it 5x faster than Enum and 4x faster than
+  # Task.async_stream
```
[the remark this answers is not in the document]
scohen: Yes, the bottleneck here is cpu, not io

apps/remote_control/lib/lexical/remote_control/search/fuzzy.ex
```diff
@@ -2,42 +2,89 @@ defmodule Lexical.RemoteControl.Search.Fuzzy do
...
   @type subject :: String.t()
-  @type entry :: {String.t(), reference()}
-  @type t :: %__MODULE__{
-          subject_to_refs: %{subject() => [reference()]},
-          path_to_refs: %{Path.t() => [reference()]}
-        }
+  @type extracted_subject :: term()
+  @type key :: term()
+  @type value :: term()
+  @type extracted_subject_key_value :: {extracted_subject(), key(), value()}
```
scohen, line 26: Better type name: fuzzy mapping

apps/remote_control/lib/lexical/remote_control/search/fuzzy/scorer.ex
```diff
@@ -0,0 +1,193 @@
...
+      subject
+      |> String.graphemes()
+      |> List.to_tuple()
+
+    subject(original: subject, graphemes: graphemes, normalized: normalize(subject))
+  end
+
+  def preprocess(subject) do
+    subject
+    |> inspect()
```
scohen, line 53:
```suggestion
    |> stringify()
```
But we should be using the function passed in to fuzzy. Maybe scorer should only accept strings as subjects

apps/remote_control/lib/lexical/remote_control/search/store.ex
```diff
@@ -0,0 +1,159 @@
...
+    {:ok, state, {:continue, :load}}
+  end
+
+  def handle_continue(:load, %State{} = state) do
+    case State.load(state) do
+      {:ok, state} ->
+        {:noreply, state}
+
+      {:error, _} ->
+        Logger.warn("Could not load nor build search state")
```
scottming, line 86:
```suggestion
        Logger.warning("Could not load nor build search state")
```

apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
```diff
@@ -0,0 +1,90 @@
...
+  @type column :: non_neg_integer
+  @type position :: {line, column}
+  @type version :: String.t()
+  @type entry_reference :: reference() | nil
+
+  defstruct [
+    :application,
+    :elixir_version,
+    :erlang_version,
+    :finish,
```
zachallaun, line 16: Is this the end position? Can we call it `:end` instead for consistency with ranges?

apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
```diff
@@ -0,0 +1,90 @@
...
+  @type subject :: String.t()
+  @type entry_subtype :: :reference | :definition
+  @type indexer :: :beam | :source
+  @type line :: non_neg_integer()
+  @type column :: non_neg_integer
+  @type position :: {line, column}
+  @type version :: String.t()
+  @type entry_reference :: reference() | nil
+
+  defstruct [
```
zachallaun, line 12: Something that may be missing from this is public/private. There are a _lot_ more `defp` than `def` in the average codebase, for instance. Within a module, we'd want results for private functions defined in that module. Looking up results from a remote module, though, we might not want to retrieve all private things too and have to filter them out after the fact.

zachallaun (commented):
I've made some comments specific to `Entry`, which really needs to be nailed down as changes to it will have knock-on effects in a lot of other places.

apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
scohen: I really hate calling things 'end' as it conflicts with elixir's syntax for `end` and throws my editor into knots, though I might just use `Lexical.Document.Range` to represent them. I think that's the better idea.

apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
```diff
@@ -0,0 +1,90 @@
+defmodule Lexical.RemoteControl.Search.Indexer.Entry do
+  @type entry_type :: :module | :function | :macro | :field
```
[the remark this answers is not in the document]
scohen: > I'm not sure that :function and :macro should be separate, since they exist in the same namespace.

What's the harm of being specific here. For find references, this is right, but I think it might be helpful to differentiate between the parts of speech of elixir. Calling something a function a macro seems wrong to me. If you want both, we should make it possible to query for both, possibly by passing a list into the query function.

apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
scohen: i'm not sure I follow; for find references, I'd like things to be exhaustive. I don't want to click on a module in some remote module and not seeing usages in private functions.

apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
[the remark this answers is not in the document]
scohen: for the `entry_type`,  maybe I should just remove everything except `:module`

apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
zachallaun: That makes sense. I was thinking from a different perspective, like fuzzy completions instead of find references. So you might have something like `MyModule.thing|` and a possible completion might be `MyModule.get_thing` -- in that case, you're looking up definitions, not references, but it seems like it shouldn't/won't be hard to exclude private things after the fact.

apps/remote_control/test/lexical/remote_control/search/indexer/source_test.exs
```diff
@@ -0,0 +1,321 @@
...
+        ~q[
+        defmodule Root do
+          @something :timer
+        end
+      ]
+        |> index()
+
+      assert erlang_module.type == :module
+      assert erlang_module.parent == module_def.ref
+      assert erlang_module.subject == :timer
```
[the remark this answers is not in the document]
scottming: BTW, the `:timer` position is `{2, 14}` to `{2, 19}`

apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
scohen: Correct, given how many times functions are called in a typical app, removing private functions will be better accomplished via a filtering step. 
I agree in general though, I'll add a `visibility` field that defaults to `public`.

apps/remote_control/test/lexical/remote_control/search/indexer/source_test.exs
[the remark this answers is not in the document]
scohen: Good point, @scottming, moving these tests to use the markers caught a bunch of issues.

apps/remote_control/test/lexical/remote_control/search/indexer/source_test.exs
```diff
@@ -0,0 +1,321 @@
...
+      {:ok, [entry]} =
+        ~q[
+        defmodule Simple do
+        end
+        ]
+        |> index()
+
+      assert entry.type == :module
+      assert entry.parent == :root
+      assert entry.start == {1, 1}
```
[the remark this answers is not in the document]
scohen: they're where the reference was found, actually.

apps/remote_control/lib/lexical/remote_control/search/fuzzy.ex
Blond11516: I see you went with a guard. Works for me 👍

apps/remote_control/lib/lexical/remote_control/search/fuzzy/scorer.ex
Blond11516: Sorry I wasn't very clear. By performance I meant the accuracy of results, not "cpu cycles" performance. Glad to know it's fast though!

apps/remote_control/lib/lexical/remote_control/search/store/ets.ex
```diff
@@ -0,0 +1,213 @@
...
+  """
+  defrecord :metadata,
+    schema_version: @version,
+    types: [:module],
+    subtypes: [:definition, :reference]
+
+  @doc """
+  A header for the records stored in the ETS table / disk.
+  The default values are initialized to :_ to allow this to be used for querying,
+  as :_ means wildcard. This _does_ mean you need to
```
Blond11516, line 23: Looks like this was cut off

Blond11516 (commented):
I still have to go through the ETS backend and store state but the updates to fuzzy and scorer look good!

scottming (approved):
I didn't look closely at the fuzzy part; I like the indexer part.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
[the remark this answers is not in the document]
zachallaun: Thinking through this more, I honestly question the utility of the `Reducer` if there's a switch to using zippers. I think a lot of code could be deleted and there could be fewer abstractions involved.

apps/remote_control/lib/lexical/remote_control/search/store/state.ex
```diff
@@ -0,0 +1,154 @@
...
+      project: project,
+      index_path: index_path(project),
+      loaded?: false,
+      update_index: update_index
+    }
+    |> ensure_index_directory_exists()
+  end
+
+  def drop(%__MODULE__{}) do
+    Ets.drop()
```
zachallaun, line 23: Since the backend is now a behaviour, maybe it's worth referencing `Ets` only once and using a module attribute?

```elixir
@backend Lexical.RemoteControl.Search.Store.Ets

...

def drop(%__MODULE__{}) do
  @backend.drop()
end
```

This would make experimenting with alternative backends a bit easier.

apps/remote_control/lib/lexical/remote_control/search/fuzzy.ex
```diff
@@ -0,0 +1,251 @@
...
+      |> Map.new(fn subject -> {subject, Scorer.preprocess(subject)} end)
+
+    %__MODULE__{
+      subject_to_values: subject_to_values,
+      grouping_key_to_values: grouping_key_to_values,
+      preprocessed_subjects: preprocessed_subjects,
+      mapper: mapper,
+      subject_converter: subject_converter
+    }
+  end
```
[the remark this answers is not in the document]
scohen: this is for performance; Calling Map.update tens of thousands of times is vastly, vastly slower than building a list of tuples and creating new maps.

apps/remote_control/lib/lexical/remote_control/search/store/state.ex
scohen: That's for the next PR. I'm going to make an mnesia backend and see how that goes.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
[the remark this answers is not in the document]
scohen: I was avoiding wide use of zippers because my last benchmarks showed them to be significantly slower than using the functions in `Macro`. This is highly critical code, and slowness here will really tank the server. 

Perhaps I should make a benchmarks project (or put benchmarks in server) so that we have something concrete to compare.

apps/common/lib/lexical/ast.ex
```diff
@@ -69,6 +69,14 @@ defmodule Lexical.Ast do
   @typedoc "Return value from `Code.Fragment.surround_context/3`"
   @type surround_context :: any()
 
+  @type quoted_elixir ::
```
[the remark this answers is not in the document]
scohen: I swear I did this and dialyzer complained. On re-check, it looks like it was angry about something else.

apps/remote_control/lib/lexical/remote_control/search/fuzzy/scorer.ex
```diff
@@ -0,0 +1,193 @@
...
+  @consecutive_character_bonus 5
+
+  defp consecutive_match_bonus(matched_positions) do
+    # This function checks for consecutive matched characters, and
+    # makes matches with more consecutive matched characters worth more.
+    # This means if I type En, it will match Enum more than it will match
+    # Something
+
+    [-50 | matched_positions]
+    |> Enum.chunk_every(2, 1, [-1])
```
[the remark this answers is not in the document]
scohen: Yep, you're right.

apps/remote_control/lib/lexical/remote_control/search/fuzzy.ex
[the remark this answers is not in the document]
zachallaun: Makes sense!

Would it be worth defining add in terms of new, then? Add could create a new fuzzy struct and then combine them. My main concern is that the "prep" logic is non-trivial enough that it would be nice not to have to essentially duplicate it. 

Defining add in terms of new would probably also be a perf hit if we're adding a lot of single items, but it may improve perf when adding bulk items.

apps/remote_control/lib/lexical/remote_control/search/fuzzy/scorer.ex
```diff
@@ -0,0 +1,193 @@
...
+
+  defp do_case_match_boost(matched_positions, <<_::utf8, rest::binary>>, graphemes, boost) do
+    do_case_match_boost(matched_positions, rest, graphemes, boost)
+  end
+
+  defp grapheme_to_utf8(graphemes, position) do
+    <<c::utf8>> = elem(graphemes, position)
+
+    c
+  end
```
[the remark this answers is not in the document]
scohen: performance again, Your example creates two additional lists, each the size of the string. Peeling off characters is much more efficient (though harder to read. I also realized that if I put the pattern first, we'll enable `bin_opt_info` on these strings).

apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
```diff
@@ -0,0 +1,83 @@
...
+          elixir_version: version(),
+          erlang_version: version(),
+          subject: subject(),
+          parent: entry_reference(),
+          path: Path.t(),
+          range: Lexical.Document.Range.t(),
+          ref: entry_reference(),
+          subtype: entry_subtype(),
+          type: entry_type(),
+          updated_at: pos_integer()
```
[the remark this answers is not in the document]
scohen: it's so gross though.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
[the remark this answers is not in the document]
zachallaun: An alternative to zippers would be to maintain and pass a stack of ancestors, which should still be as efficient but give the extractors all the context they need. (E.g. aliases/atoms can check if their parent is a defmodule)

apps/remote_control/lib/lexical/remote_control/search/store/state.ex
zachallaun: Sounds good!

apps/remote_control/lib/lexical/remote_control/search/fuzzy/scorer.ex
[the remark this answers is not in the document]
zachallaun: Makes sense! Fine to leave it. Double check the third do_case_match_boost clause though, I don't think it can ever be called?

apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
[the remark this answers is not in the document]
zachallaun: Hah! I don't think it's that bad and is imo a bit simpler and easier to introspect (Unix timestamps are kinda a pain to understand and compare visually). I'd defer to you on this though -- just a small suggestion I saw that might get rid of some code.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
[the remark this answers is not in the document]
scohen: aren't blocks a stack of ancestors?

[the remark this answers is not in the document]
zachallaun: Not quite, because blocks are only added when you enter a `do`, unless I'm remembering incorrectly.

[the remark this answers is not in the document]
scohen: Another difference I'll point out is that what you suggested can traverse things multiple times, once in the parent reducer / zipper / whatever, and another time in the extractors. We're going to have a lot of extractors, and I think it's desirable to ensure that we only go over the document once.

apps/remote_control/lib/lexical/remote_control/search/fuzzy.ex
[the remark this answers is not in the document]
scohen: > Defining add in terms of new would probably also be a perf hit

Yes, exactly. The only way it'd be faster is if you were adding more items than already exist in the data structure.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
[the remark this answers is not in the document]
zachallaun: While I agree with your concern around making sure things are traversed once, I don't think my suggestion makes anything "worse". Extractors can already do extra traversals on the nodes they get; the only difference is that they could start from higher up.

If you wanted to ensure that extractors couldn't restart traversal from one of the ancestors in the stack, I suppose you could store a stack of call/meta with no children instead. 

Basically here's what my argument boils down to: things are simpler if extractors can inspect the ancestors of the current node.

I'm also fine with this turning into a case of "show, don't tell" -- if you want to move forward as is, I can do an experimental refactor later and see how things pan out and you can decide at that point whether it's an improvement.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
```diff
@@ -0,0 +1,175 @@
...
+
+        {:ok, entry}
+
+      _ ->
+        :ignored
+    end
+  end
+
+  # This matches an erlang module, which is just an atom
+  def extract(atom_literal, %Reducer{} = reducer) when is_atom(atom_literal) do
```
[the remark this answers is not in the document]
scohen: errm, I don't think this is the case; When I prewalk the AST, I see stuff like: 
```elixir
:__block__, [line: 1, column: 21], [:do]}
```
which isn't an atom.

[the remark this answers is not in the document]
zachallaun: It is an atom though, because `do` desugars into a keyword list where the key is `:do`. 

```elixir
if true do
  :ok
end

if(true, [{:do, :ok}])
```

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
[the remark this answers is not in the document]
scohen: >  Extractors can already do extra traversals on the nodes they get; the only difference is that they could start from higher up.

That's not the same thing as using `Zipper./up/down/left/right`. Those do a lot more thorough traversals at the cost of a lot of speed. 

I would like to get this merged, so I can work on integrating this into lexical. I'm fine with a zipper based attempt, but we have to be very cognizant about the speed of things. Elixir's `Code` module is _not_ fast, and dorgan himself has complained that zippers are even slower. This stuff processes a _lot_ of data, and we really need to keep it quick.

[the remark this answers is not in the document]
zachallaun: Sounds good! Once merged, I'll experiment a bit and share any major breakthroughs. I think it's possible to simplify things pretty dramatically while retaining the current performance, but the onus is on me to show that.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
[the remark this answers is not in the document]
scohen: yes, but it isn't an atom in the sense anyone would ever care to find all references to "do"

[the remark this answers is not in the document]
scohen: unless they want to stress the fuzzy finder.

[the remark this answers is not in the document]
scohen: though with the module check, it won't be extracted.

Issue #171, opened 2023-05-19
Format won't work after using `% s/old/new/g` in neovim


scottming:
This is the error

```elixir
lexical: -32803: %SyntaxError{file: "/Users/scottming/Code/dummy/lib/dummy.ex", line: 9, column: 16, snippet: %{content: "    |>
 Enum.ffï¢´(&String.contains?(&1, \"a\"))", offset: 15}, description: "unexpected token: Codepoint failed identifier tokenizatio
n, but a simpler form was found.\n\nGot:\n\n    \"ffï¢\" (code points 0x00066 0x00066 0x000EF 0x000A2)\n\nHint: You could write
the above in a similar way that is accepted by Elixir:\n\n    \"ffïc̸\" (code points 0x00066 0x00066 0x000EF 0x00063 0x00338)\n\n
See https://hexdocs.pm/elixir/unicode-syntax.html for more information."}
```

```elixir
defmodule Dummy do
  @moduledoc false

  def hello do
    :world

    ["a", "b", "c"]
    |> Enum.filter(&String.contains?(&1, "x"))
    |> Enum.filter(&String.contains?(&1, "a"))
  end
end
```

and then `% s/filter/f/g`, then do format.

I'm not sure if this is a lexical bug or not, will add more information later.

scottming:
To make this error replicated easier, this is the request of neovim:

first you need to create a file:

```elixir
defmodule Dummy.Document do
  2
  2
end
```

then do 

```lua
[DEBUG][2023-05-19 14:58:49] .../vim/lsp/rpc.lua:284	"rpc.send"	{
  jsonrpc = "2.0",
  method = "textDocument/didChange",
  params = {
    contentChanges = { {
        range = {
          ["end"] = {
            character = 3,
            line = 2
          },
          start = {
            character = 2,
            line = 1
          }
        },
        rangeLength = 5,
        text = "1\n  1"
      } },
    textDocument = {
      uri = "file:///Users/scottming/Code/dummy/lib/dummy/document.ex",
      version = 3
    }
  }
}
[DEBUG][2023-05-19 14:58:49] .../vim/lsp/rpc.lua:284	"rpc.send"	{
  jsonrpc = "2.0",
  method = "textDocument/didChange",
  params = {
    contentChanges = { {
        range = {
          ["end"] = {
            character = 3,
            line = 2
          },
          start = {
            character = 2,
            line = 1
          }
        },
        rangeLength = 5,
        text = "1\n  1"
[...]

PR #353, opened 2023-09-01
Convert utf8->latin1 before decoding JSON-RPC payloads

### LINKED ISSUE #287: Unicode diffing bug
I typed

a = "你好

then delete`好`, and append `"`

Then I see an unexpected token error:

CleanShot 2023-07-26 at 22 15 41@2x

This is all the related client log:

```lua
[DEBUG][2023-07-26 22:24:16] .../vim/lsp/rpc.lua:284	"rpc.send"	{
  jsonrpc = "2.0",
  method = "textDocument/didChange",
  params = {
    contentChanges = { {
        range = {
          ["end"] = {
            character = 4,
            line = 16
          },
          start = {
            character = 4,
            line = 16
          }
        },
        rangeLength = 0,
        text = 'a = "你好'
      } },
    textDocument = {
      uri = "file:///Users/scottming/Code/dummy/lib/dummy.ex",
      version = 5
    }
  }
}

[DEBUG][2023-07-26 22:24:54] .../vim/lsp/rpc.lua:284	"rpc.send"	{
  jsonrpc = "2.0",
  method = "textDocument/didChange",
  params = {
    contentChanges = { {
        range = {
          ["end"] = {
            character = 11,
            line = 16
          },
          start = {
            character = 10,
            line = 16
          }
        },
        rangeLength = 1,
        text = ""
      } },
    textDocument = {
      uri = "file:///Users/scottming/Code/dummy/lib/dummy.ex",
      version = 6
    }
  }
}
[DEBUG][2023-07-26 22:24:56] .../vim/lsp/rpc.lua:284	"rpc.send"	{
  jsonrpc = "2.0",
  method = "textDocument/didChange",
  params = {
    contentChanges = { {
        range = {
          ["end"] = {
            character = 10,
            line = 16
          },
          start = {
            character = 10,
            line = 16
          }
        },
        rangeLength = 0,
        text = '"'
      } },
    textDocument = {
      uri = "file:///Users/scottming/Code/dummy/lib/dummy.ex",
      version = 7
    }
  }
}

[DEBUG][2023-07-26 22:24:56] .../vim/lsp/rpc.lua:387	"rpc.receive"	{
  jsonrpc = "2.0",
  method = "textDocument/publishDiagnostics",
  params = {
    diagnostics = { {
        message = 'unexpected token: " " (column 12, code point U+00A0)',
        range = {
          ["end"] = {
            character = 0,
            line = 17
          },
          start = {
            character = 10,
            line = 16
          }
        },
        severity = 1,
        source = "Elixir"
      } },
    uri = "file:///Users/scottming/Code/dummy/lib/dummy.ex"
  }
}
```

zachallaun:
This (hopefully) fixes #287.

I'm not completely sure why this is occurring, but non-ascii characters were seemingly double-encoded, at least on my system (Ubuntu 22.04 via WSL, VS Code desktop on Windows). I've confirmed with `:io.getopts()` right before `IO.read/2` is called that the encoding is set to `:latin1`, but the result after `JsonRpc.decode/1` was that text would be utf8 double-encoded -- that is, it was what you would expect from:

    latin1_data
    |> :unicode.characters_to_binary(:latin1, :utf8)
    |> :unicode.characters_to_binary(:latin1, :utf8)

This would result in `Document.Line` text containing more bytes than it should, which causes text edits to fail, leading behind random extra bytes.

I don't love this "fix" because it feels very much like a band-aid, not addressing whatever the root issue is, but I'd also rather things work in documents containing multi-byte characters.

### DIFF
--- a/apps/server/lib/lexical/server/transport/std_io.ex
+++ b/apps/server/lib/lexical/server/transport/std_io.ex
@@ -86,7 +86,7 @@ defmodule Lexical.Server.Transport.StdIO do
 
         with {:ok, content_length} <-
                header_value(headers, "content-length", &String.to_integer/1),
-             {:ok, data} <- read(device, content_length),
+             {:ok, data} <- read_body(device, content_length),
              {:ok, message} <- JsonRpc.decode(data) do
           callback.(message)
         end
@@ -112,10 +112,16 @@ defmodule Lexical.Server.Transport.StdIO do
     end
   end
 
-  defp read(device, amount) do
+  defp read_body(device, amount) do
     case IO.read(device, amount) do
-      data when is_binary(data) or is_list(data) -> {:ok, data}
-      other -> other
+      data when is_binary(data) or is_list(data) ->
+        # This is a bit "magical" and is likely a symptom of a bug elsewhere.
+        # See https://github.com/lexical-lsp/lexical/issues/287 for context.
+        data = :unicode.characters_to_binary(data, :utf8, :latin1)
+        {:ok, data}
+
+      other ->
+        other
     end
   end
 
--- a/apps/server/lib/lexical/server/transport/std_io.ex
+++ b/apps/server/lib/lexical/server/transport/std_io.ex
@@ -115,7 +115,7 @@ defmodule Lexical.Server.Transport.StdIO do
   defp read_body(device, amount) do
     case IO.read(device, amount) do
       data when is_binary(data) or is_list(data) ->
-        # This is a bit "magical" and is likely a symptom of a bug elsewhere.
+        # Ensure that incoming data is latin1 to prevent double-encoding to utf8 later
         # See https://github.com/lexical-lsp/lexical/issues/287 for context.
         data = :unicode.characters_to_binary(data, :utf8, :latin1)
         {:ok, data}


scottming:
I tried it and it seems to fix the Chinese encoding issue and it fixes #171  as well:

I can get more people to test it.

scottming (approved):

apps/server/lib/lexical/server/transport/std_io.ex
```diff
@@ -112,10 +112,16 @@ defmodule Lexical.Server.Transport.StdIO do
     end
   end
 
-  defp read(device, amount) do
+  defp read_body(device, amount) do
     case IO.read(device, amount) do
-      data when is_binary(data) or is_list(data) -> {:ok, data}
-      other -> other
+      data when is_binary(data) or is_list(data) ->
+        # This is a bit "magical" and is likely a symptom of a bug elsewhere.
```
scohen, line 118: I take slight issue with this comment. This is the lowest level of lexical. If the bug is elsewhere, where is it? You established that the binaries decoded here were not
Utf8.  Where else would the bug be?

zachallaun: "Bug" may be the wrong word -- "misunderstanding" is probably more accurate.

How about:

```elixir
# Ensure that incoming data is latin1 to prevent double-encoding to utf8 later
```

zachallaun:
```suggestion
        # Ensure that incoming data is latin1 to prevent double-encoding to utf8 later
```

PR #355, opened 2023-09-04
Fix markdown formatting for supported versions

reisub:
The markdown rendering of README is confusing because of the `>` being interpreted as quote start.

**Before this change:**
Screenshot 2023-09-04 at 09 43 42

**After this change:**
Screenshot 2023-09-04 at 09 44 30


scohen (approved):

scohen:
Thank you!

PR #357, opened 2023-09-07
Refactor: Moved Dispatch from server to remote_control

scohen:
This change moves the dispatcher for messages to remote control, and out of server. It also unifies remote control's event emitting logic (which was spread between using Server's dispatcher and RemoteControl's own Compilation.Dispatch module).

The result is a much slimmed down dispatch module that no longer needs to take a project as a parameter for all its function calls.

I also added a helper function for unit tests so that they don't need to start a remote control node in order to test dispatching logic.


zachallaun (approved):
This looks great! The new structure makes a lot more sense.

Issue #113, opened 2023-04-23
Go to definition can't jump to the correct position if the target is editing


scottming:

scohen:
i don't understand what this means.

scottming:
<img width="703" alt="image" src="https://user-images.githubusercontent.com/12830256/237035306-64ba1a40-3cfb-4aa4-b22a-263aedc98a0b.png">

What I mean is that I opened both the module and the defined module at the same time. If I am editing the defined module(not saved yet), then when I jump, it is likely to go to the wrong position, like in the picture below.

My cursor was on `Dummy.|print` in the test file, but the peek definition jumped to `|>`, which is incorrect. Therefore, this is a bug of ElixirSense.

scohen:
This is fixed with the new indexer.

PR #359, opened 2023-09-11
Integrated indexing into language server

scohen:
This commit has the following changes:

  1. RemoteControl now emits an event when a document has been quoted.
  2. Added a handler to listen for the above event, and update the index when a newly quoted document is emitted
  3. Wired up workspace file changes so that when a file is created or deleted, it emits an event to RemoteControl
  4. When a filesystem create or delete is detcted by the indexing handler, it is either indexed or its information is removed from the index.

I also had to make Document.Store global. This required moving `fetch` back into the GenServer

After the above changes, I can now edit documents and find module definitions and references via the console.

### DIFF
--- a/apps/common/lib/lexical/ast.ex
+++ b/apps/common/lib/lexical/ast.ex
@@ -394,8 +394,8 @@ defmodule Lexical.Ast do
     end
   end
 
-  def expand_aliases(_, _, {line, column}) do
-    Position.new(line, column)
+  def expand_aliases(segments, doc, {line, column}) do
+    expand_aliases(segments, doc, Position.new(line, column))
   end
 
   def expand_aliases(_, _, empty) do
--- a/apps/remote_control/lib/lexical/remote_control/dispatch/handlers/indexing.ex
+++ b/apps/remote_control/lib/lexical/remote_control/dispatch/handlers/indexing.ex
@@ -25,7 +25,7 @@ defmodule Lexical.RemoteControl.Dispatch.Handlers.Indexing do
   end
 
   def reindex(%Document{} = document, quoted_ast) do
-    with :ok <- up_to_date?(document),
+    with :ok <- ensure_latest_version(document),
          {:ok, entries} <- Indexer.Quoted.index(document, quoted_ast) do
       Search.Store.update(document.path, entries)
     end
@@ -36,7 +36,7 @@ defmodule Lexical.RemoteControl.Dispatch.Handlers.Indexing do
     Search.Store.clear(path)
   end
 
-  defp up_to_date?(%Document{version: version, uri: uri}) do
+  defp ensure_latest_version(%Document{version: version, uri: uri}) do
     case Document.Store.fetch(uri) do
       {:ok, %Document{version: ^version}} ->
         :ok
--- a/apps/remote_control/lib/lexical/remote_control/dispatch/handlers/indexing.ex
+++ b/apps/remote_control/lib/lexical/remote_control/dispatch/handlers/indexing.ex
@@ -1,0 +1,51 @@
+defmodule Lexical.RemoteControl.Dispatch.Handlers.Indexing do
+  alias Lexical.Document
+  alias Lexical.RemoteControl.Api.Messages
+  alias Lexical.RemoteControl.Dispatch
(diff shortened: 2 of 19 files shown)


apps/remote_control/lib/lexical/remote_control/dispatch/handlers/indexing.ex
```diff
@@ -0,0 +1,51 @@
...
+    end
+  end
+
+  def delete_path(uri) do
+    path = Document.Path.ensure_path(uri)
+    Search.Store.clear(path)
+  end
+
+  defp up_to_date?(%Document{version: version, uri: uri}) do
+    case Document.Store.fetch(uri) do
```
scottming, line 40: That's why we need the `Document.Store` to be global.

I think this way can solve this issue: #113

apps/remote_control/lib/lexical/remote_control/search/store.ex
```diff
@@ -71,6 +75,10 @@ defmodule Lexical.RemoteControl.Search.Store do
     GenServer.start_link(__MODULE__, [project, create_index, refresh_index], name: __MODULE__)
   end
 
+  def start_link([create_index, refresh_index]) do
```
zachallaun, line 78: This is a small thing, but something I noticed when I saw this being started in `RemoteControl.Application`: can we make this a kw list or a tuple?

```suggestion
  def start_link(opts) do
    create_index = Keyword.fetch!(opts, :create)
    refresh_index = Keyword.fetch!(opts, :update)
```

zachallaun (commented):
Looks good! Made a couple small comments. 

I'm wondering if the document store should be named per-project -- possibly started with the project supervisor and available through `project.store` or `project.document_store` (as opposed to referencing `Document.Store` directly), but I think that's tangential to this PR.

apps/common/lib/lexical/ast.ex
```diff
@@ -394,8 +394,8 @@ defmodule Lexical.Ast do
     end
   end
 
-  def expand_aliases(_, _, {line, column}) do
-    Position.new(line, column)
+  def expand_aliases(segments, doc, {line, column}) do
+    expand_aliases(segments, doc, Position.new(line, column))
```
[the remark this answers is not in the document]
scohen: Yeah, it grabbed me as being "obvoiusly wrong". This is why I pinged you yesterday on discord.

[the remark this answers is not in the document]
zachallaun: Sorry I missed it last time! Good catch though.

apps/remote_control/lib/lexical/remote_control/dispatch/handlers/indexing.ex
```diff
@@ -0,0 +1,51 @@
...
+      Search.Store.update(document.path, entries)
+    end
+  end
+
+  def delete_path(uri) do
+    path = Document.Path.ensure_path(uri)
+    Search.Store.clear(path)
+  end
+
+  defp up_to_date?(%Document{version: version, uri: uri}) do
```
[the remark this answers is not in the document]
scohen: doh, this got refactored right before I pushed. Good catch.

apps/remote_control/lib/lexical/remote_control/dispatch/handlers/indexing.ex
scohen: yes, maybe?

apps/remote_control/lib/lexical/remote_control/search/store.ex
scohen: i've always seen init_args as a list to mimic positional args. 🤔 

This seemingly small change has a fairly large change in tests and IMO, makes the ergonomics of `start_supervised!` a bit worse, now we'll have to type keywords into each call to it, where we can do something like, misspell keys (I did this very thing). 

If this was something that was a top-level API, i'd probably make that change, but I think that this only occurs in a couple places and it's not worth it.

scohen:
> I'm wondering if the document store should be named per-project

That really depends on if you can have the same document visible across multiple projects. My thinking there is that you have one file on disk, and you effectively need a single representation of that file and any changes to it in one place.

PR #363, opened 2023-09-12
Entry LSP conversion improvements

scohen:
While integrating indexing into the language server, I encountered a pretty severe performance issue. Find references can return a _lot_ of things, and all of those references include ranges that need to be converted into LSP form before we can emit them. This poses a problem, we will likely need to open up dozens (possibly hundreds) of files in order to do the conversion. This will cause a massive amount of File I/O, and just can't work.

However, if we can store the lines along with the entries, then we can create a sparse Document.Lines document and use _that_ in the conversion step (which is already supported in conversions.ex), then we won't need to do _any_ file I/O at all.

This change introduces a `sparse` constructor for Document.Lines and changes the `Entry` constructors to take documents rather than paths for their first arguments so we can pull out the lines in question.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
@@ -1,4 +1,6 @@
 defmodule Lexical.RemoteControl.Search.Indexer.Entry do
+  alias Lexical.Document
+
   @type entry_type :: :module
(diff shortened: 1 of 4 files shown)


scohen:
Note: If we accept #364, then we can just store ranges in the entry, which will be a lot cleaner, _and_ perform better.

scohen:
Closing, since we accepted #364

PR #364, opened 2023-09-12
Position refactor

scohen:
@zachallaun and I realized that if you track the line that a position occurs on, you no longer need the document to do conversions. Watch this space.


apps/protocol/lib/lexical/protocol/convertibles/lexical.document.range.ex
```diff
@@ -2,8 +2,8 @@ defimpl Lexical.Convertible, for: Lexical.Document.Range do
   alias Lexical.Document
   alias Lexical.Protocol.Conversions
 
-  def to_lsp(%Document.Range{} = position, context_document) do
-    Conversions.to_lsp(position, context_document)
+  def to_lsp(%Document.Range{} = position) do
+    Conversions.to_lsp(position)
```
zachallaun, line 6: This is a typo that was already there but we should just fix it: position -> range (looks like same typo is in the clause below as well)

apps/server/test/lexical/convertibles/lexical.plugin.diagnostic.result_test.exs
```diff
@@ -20,7 +20,7 @@ defmodule Lexical.Convertibles.Lexical.Plugin.V1.Diagnostic.ResultTest do
         def fun_two do
           "🎸hello"
         end
-      end
+      endpl
```
zachallaun, line 23: Typo?

scohen: I was thinking: how did the tests even pass with this?

zachallaun (approved):
I'm pretty confident in this. As you mentioned, there may be more we can do by looking further into errors.ex, but I think these changes are a good start and we're going to be building on them imminently, so there will be more opportunity to surface any weird behavior.

scohen:
agreed, @zachallaun. I built this branch and have been using it as my main LSP for a couple days. No issues so far.

projects/lexical_test/lib/lexical/test/position_support.ex
```diff
@@ -0,0 +1,72 @@
...
+  def normalize(%Range{} = range) do
+    range
+    |> update_in([:start], &normalize/1)
+    |> update_in([:end], &normalize/1)
+  end
+
+  def normalize(%Position{} = position) do
+    %Position{
+      position
+      | context_line: nil,
```
scottming, line 41: I actually don't quite understand the reason for normalization. Is it for more convenient testing?

apps/protocol/lib/lexical/protocol/conversions.ex
```diff
@@ -121,31 +117,32 @@ defmodule Lexical.Protocol.Conversions do
...
-  defp extract_lsp_character(%ElixirPosition{} = position, line(ascii?: true, text: text)) do
+  defp extract_lsp_character(
+         %ElixirPosition{context_line: line(ascii?: true, text: text)} = position
+       ) do
     character = min(position.character - 1, byte_size(text))
     {:ok, character}
   end
 
-  defp extract_lsp_character(%ElixirPosition{} = position, line(text: utf8_text)) do
+  defp extract_lsp_character(%ElixirPosition{context_line: line(text: utf8_text)} = position) do
```
scottming, line 145: I think I get it, mainly because the additional information is needed here

scottming (approved):

projects/lexical_test/lib/lexical/test/position_support.ex
scohen: It’s so you can assert by making a position that doesn’t have all the fields set. Otherwise you’d need to copy the line, the offset, and other fields into the position, when all you care about is the line number and character.

apps/protocol/lib/lexical/protocol/conversions.ex
scohen: Yes

PR #367, opened 2023-09-14
Better typespecs

scohen:
This PR has three commits, the first is the biggest, and adds (very) detailed type information to our generated modules. Thanks to this, i found a couple legit bugs in the code. 

I also had to nudge the generators slightly in order to get them to generate base types, which were missing from the codebase. Since the missing things were just aliases to base types, stuff would still work.


apps/proto/lib/lexical/proto/macros/message.ex
```diff
@@ -9,15 +9,15 @@ defmodule Lexical.Proto.Macros.Message do
...
     parse_fn =
       if Keyword.get(opts, :include_parse?, true) do
         Parse.build(types)
       end
 
     quote do
-      unquote(Struct.build(types))
-      unquote(Typespec.build())
+      unquote(Struct.build(types, env))
+      unquote(Typespec.t())
```
zachallaun, line 20: This would become:

```elixir
@type t :: unquote(Typespec.typespec())
```

apps/proto/lib/lexical/proto/alias.ex
```diff
@@ -1,12 +1,15 @@
...
   alias Lexical.Proto.CompileMetadata
   alias Lexical.Proto.Field
+  alias Lexical.Proto.Macros.Typespec
 
   defmacro defalias(alias_definition) do
     caller_module = __CALLER__.module
     CompileMetadata.add_type_alias_module(caller_module)
 
     quote location: :keep do
+      unquote(Typespec.t(alias_definition, __CALLER__))
```
zachallaun, line 11: I found `Typespec.t` a little confusing at first. What if we just call `typespec` directly?

```elixir
@type t :: unquote(Typespec.typespec(alias_definition, __CALLER__))
```

apps/proto/lib/lexical/proto/response.ex
```diff
@@ -13,15 +13,15 @@ defmodule Lexical.Proto.Response do
...
       result: quote(do: optional(unquote(response_type)))
     ]
 
     quote location: :keep do
       alias Lexical.Proto.LspTypes
       unquote(Access.build())
-      unquote(Struct.build(jsonrpc_types))
-      unquote(Typespec.build())
+      unquote(Struct.build(jsonrpc_types, __CALLER__))
+      unquote(Typespec.t())
```
zachallaun, line 24:
```elixir
@type t :: unquote(Typespec.typespec())
```

apps/proto/lib/lexical/proto/type.ex
```diff
@@ -20,8 +20,8 @@ defmodule Lexical.Proto.Type do
       unquote(Json.build(caller_module))
       unquote(Inspect.build(caller_module))
       unquote(Access.build())
-      unquote(Struct.build(types))
-      unquote(Typespec.build(types))
+      unquote(Struct.build(types, __CALLER__))
+      unquote(Typespec.t(types, __CALLER__))
```
zachallaun, line 24:
```elixir
@type t :: unquote(Typespec.typespec(types, __CALLER__))
```

apps/proto/lib/lexical/proto/macros/struct.ex
```diff
@@ -23,7 +25,10 @@ defmodule Lexical.Proto.Macros.Struct do
     quote location: :keep do
       @enforce_keys unquote(required_keys)
       defstruct unquote(keys)
+      unquote(Typespec.keyword_constructor_options(opts, env, :option, :options))
```
zachallaun, line 28: This would get hoisted above the `quote`, so it would be:

```elixir
{option_typespec, options_typespec} = Typespec.keyword_constructor_options(opts, env)

quote location: :keep do
  @enforce_keys unquote(required_keys)
  defstruct unquote(keys)
  @type option :: unquote(option_typespec)
  @type options :: unquote(options_typespec)
```

zachallaun (commented):
I think this looks good. My only real suggestion is that I think things would be a bit more obvious if `Typespec` were only responsible for constructing the bits on the right of `::`. See various comments.

scohen:
> I think things would be a bit more obvious if Typespec were only

I considerd this, your comment cinches my hunch, I'll change it.

apps/proto/lib/lexical/proto/macros/typespec.ex
```diff
@@ -1,7 +1,213 @@
...
+
+  defp do_typespec({:integer, _, _}, _env) do
+    quote(do: integer())
+  end
+
+  defp do_typespec({:float, _, _}, _env) do
+    quote(do: float())
+  end
+
+  defp do_typespec({:__MODULE__, [line: 740], nil}, env) do
```
zachallaun, line 65: Where does like 740 come from? 😝

scohen: bad copy paste

apps/proto/lib/lexical/proto/enum.ex
```diff
@@ -25,6 +45,18 @@ defmodule Lexical.Proto.Enum do
     end
   end
 
+  defp determine_type(i) when is_integer(i) do
```
zachallaun, line 48: Now that Typespec has a `literal_type` function, do we want to make that public and use it here?

scohen: Typespec's function is too broad; Enums can only have string or integer values. Everything else is an error.

zachallaun: Ah got it

apps/proto/lib/lexical/proto/macros/typespec.ex
```diff
@@ -1,7 +1,213 @@
...
+      _ ->
+        quote do
+          unquote(expanded_alias).t()
+        end
+    end
+  end
+
+  defp do_typespec({:literal, _, value}, _env) when is_atom(value) do
+    quote do
+      unquote(value)
```
zachallaun, line 116: There are a handful of quote blocks like this one of the form:

```elixir
quote do
  unquote(thing)
end
```

All of these are equivalent to just returning `thing` without any additional quoting/unquoting. If it's a stylistic choice to make it clear that we're returning something that can be used as a quoted form, that's fine!

zachallaun (approved):

PR #368, opened 2023-09-14
Aliases was confused by nested non-module blocks

scohen:
Aliases would get confused if a module had another block that didn't define another module (like in a DSL). This was because it only popped a single scope at a time, and if several scopes ended on the same line, only the inner scope would be popped. Calling the pop funcion recursively solves the issue.


zachallaun (approved):
Your explanation and the fix make sense to me!

PR #371, opened 2023-09-16
Async indexing

zachallaun:
Depending on the computer and size of the project being indexed, it was possible for indexing to take longer than the previous 20 second timeout. Additionally, since indexing was blocking the store GenServer, any calls made to the Store during a long indexing would be likely to time out (default GenServer call timeout is 5 seconds).

This PR updates the `Store.State` API to load asynchronously so that the store can continue to respond while indexing occurs in a task, while the indexing itself has had its timeout changed to `:infinity`.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/store.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/store.ex
@@ -92,6 +92,7 @@ defmodule Lexical.RemoteControl.Search.Store do
     {:ok, state}
   end
 
+  # handle the result from `State.async_load/1`
   def handle_info({ref, result}, %State{async_load_ref: ref} = state) do
     {:noreply, State.async_load_complete(state, result)}
   end
--- a/apps/remote_control/lib/lexical/remote_control/search/store.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/store.ex
@@ -42,6 +42,10 @@ defmodule Lexical.RemoteControl.Search.Store do
     GenServer.call(__MODULE__, :all)
   end
 
+  def loaded? do
+    GenServer.call(__MODULE__, :loaded?)
+  end
+
   def replace(entries) do
     GenServer.call(__MODULE__, {:replace, entries})
   end
@@ -80,20 +84,20 @@ defmodule Lexical.RemoteControl.Search.Store do
   end
 
   def init([%Project{} = project, create_index, update_index]) do
-    state = State.new(project, create_index, update_index)
+    state =
+      project
+      |> State.new(create_index, update_index)
+      |> State.async_load()
 
-    {:ok, state, {:continue, :load}}
+    {:ok, state}
   end
 
-  def handle_continue(:load, %State{} = state) do
-    case State.load(state) do
-      {:ok, state} ->
-        {:noreply, state}
+  def handle_info({ref, result}, %State{async_load_ref: ref} = state) do
+    {:noreply, State.async_load_complete(state, result)}
+  end
 
-      {:error, _} ->
-        Logger.warning("Could not load nor build search state")
-        {:noreply, state}
-    end
+  def handle_info(_, state) do
+    {:noreply, state}
   end
 
   def handle_call({:replace, entities}, _from, %State{} = state) do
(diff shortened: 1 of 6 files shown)


philipgiuliani:
I've tested it in a project which I couldn't use with Lexical because it was in an indexing loop. Using this branch it worked now and the LSP seems to be fully functional.

scohen (commented):
Note: if you want to change a name, I'm fine with that, but that should be done in a separate PR. A lot of this one is just renaming `refresh` to `update`, which makes it a lot harder to review.

zachallaun:
> Note: if you want to change a name, I'm fine with that, but that should be done in a separate PR. A lot of this one is just renaming refresh to update, which makes it a lot harder to review.

Yep, good point. I tend to best understand code by refactoring it, because it helps me understand what the code is doing and helps develop my mental model. In this particular case, I was initially thrown off by the use of refresh and update somewhat interchangeably, so I decided to normalize it to `update`. I agree that it should have been a separate PR, though.

apps/server/lib/lexical/server/iex/helpers.ex
```diff
@@ -131,6 +131,21 @@ defmodule Lexical.Server.IEx.Helpers do
     |> Lexical.Server.Project.Supervisor.start()
   end
 
+  def time(fun) when is_function(fun, 0) do
+    start = System.monotonic_time()
+    result = fun.()
+    stop = System.monotonic_time()
+    duration = System.convert_time_unit(stop - start, :native, :millisecond)
```
scohen, line 138:
```suggestion
   {duration_micros, result} = :timer.tc(fun)
    duration = duration_micros / 1000
```

apps/remote_control/lib/lexical/remote_control/search/store.ex
```diff
@@ -66,34 +70,34 @@ defmodule Lexical.RemoteControl.Search.Store do
...
 
-    {:ok, state, {:continue, :load}}
+    {:ok, state}
   end
 
-  def handle_continue(:load, %State{} = state) do
-    case State.load(state) do
-      {:ok, state} ->
-        {:noreply, state}
+  def handle_info({ref, result}, %State{async_load_ref: ref} = state) do
```
scohen, line 95: put a commnt here that this is coming from `Task`. You also need to handle the possible `DOWN` message mentioned here: https://hexdocs.pm/elixir/1.13/Task.html#await/2-compatibility-with-otp-behaviours

apps/server/lib/lexical/server/iex/helpers.ex
```diff
@@ -131,6 +131,21 @@ defmodule Lexical.Server.IEx.Helpers do
...
+
+    IO.puts([
+      IO.ANSI.format([:cyan, :bright, "Time: "]),
+      to_string(duration),
+      "ms"
```
scohen, line 143: this is also easy to do with `Lexical.Formats.time(microseconds)`
So this whole function can be reduced to:
```elixir
{elapsed_us, result} = :timer.tc(fun)
IO.puts(IO.ANSI.format([:cyan, :bright, "Time: ", Lexical.Format.time(elaped_us)])
result
```

scohen (approved):

apps/remote_control/lib/lexical/remote_control/search/store.ex
zachallaun: Okay so handling `DOWN` is actually a little tricky and may not be necessary.

Since I'm using `Task.async/1` to start the tasks, the caller is linked to them. That means that if the task *crashes*, the store is going to crash. (Note that this retains the same behavior as before the PR, since the `create_index` and `update_index` functions were just being run directly in the store process.)

So the only `DOWN` messages we'll be able to receive are those where the task completed, which means we'll receive the result and handle it.

If we *don't* want a crashed task to crash the store, we'll have to start them in a `Task.Supervisor`, which lets us use `Task.Supervisor.async_nolink`. I don't think that's necessarily a bad idea, but it may be something that can be addressed later and not in this PR.

Thoughts?

scohen (approved):

PR #373, opened 2023-09-17
Fix normalize_path on Windows

CharlesOkwuagwu:
This fixes `normalize_path ` on Windows.

### DIFF
--- a/apps/server/lib/mix/tasks/package.ex
+++ b/apps/server/lib/mix/tasks/package.ex
@@ -325,7 +325,8 @@ defmodule Mix.Tasks.Package do
         case :os.type() do
           {:win32, _} -> 
             orig
-           _ ->
+
+          _ ->
             path
             |> Path.dirname()
             |> Path.join(orig)
--- a/apps/server/lib/mix/tasks/package.ex
+++ b/apps/server/lib/mix/tasks/package.ex
@@ -269,7 +269,6 @@ defmodule Mix.Tasks.Package do
       case priv_dir(app_name) do
         {:ok, priv_source_dir} ->
           File.cp_r!(priv_source_dir, priv_dest_dir)
-
         _ ->
           :ok
       end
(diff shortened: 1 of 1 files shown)


apps/server/lib/mix/tasks/package.ex
```diff
@@ -269,7 +269,6 @@ defmodule Mix.Tasks.Package do
       case priv_dir(app_name) do
         {:ok, priv_source_dir} ->
           File.cp_r!(priv_source_dir, priv_dest_dir)
-
```
zachallaun, line 272: Please keep this newline.

apps/server/lib/mix/tasks/package.ex
```diff
@@ -323,11 +322,16 @@ defmodule Mix.Tasks.Package do
...
     case File.read_link(path) do
       {:ok, orig} ->
-        path
-        |> Path.dirname()
-        |> Path.join(orig)
-        |> Path.expand()
-        |> Path.absname()
+        case :os.type() do
+          {:win32, _} -> 
+            origin
```
zachallaun, line 327:
```suggestion
            orig
```

apps/server/lib/mix/tasks/package.ex
```diff
@@ -323,11 +322,16 @@ defmodule Mix.Tasks.Package do
...
+           _ ->
```
zachallaun, line 328:
```suggestion

          _ ->
```

zachallaun (changes requested):
Thank you for trying to address this!

There's a compilation error in this PR (`origin` doesn't exist). I'm also wondering if we can use CI to try to run `mix package` in a windows environment and make sure it succeeds.

apps/server/lib/mix/tasks/package.ex
CharlesOkwuagwu: Cool. I guess we can close this then.

scohen:
What was the bug? There's no description in the PR, and it looks to me like this won't handle relative paths on windows.

scohen:
I'm closing this, as I don't think it will have any meaningful effect on whether or not lexical runs on windows.

PR #374, opened 2023-09-17
Add Support for `Renaming module`

scottming:
We need to consider two typical scenarios: 

1. One is renaming an exact entry, 
2. And the other is renaming both the entry and its descendants.

I differentiate between these two scenarios based on the cursor's position. If there are no more child modules after the cursor, for example, in the case of `defmodule TopLevel.|Entry`, then rename the `Entry` module. 

However, if there are child modules after the cursor, then rename both the Entry module and its descendants. 
For example, if the cursor is at `defmodule TopLevel.|Entry.Child`, and we want to rename it to `defmodule TopLevel.Renamed.Child`, then `TopLevel.Entry.AnotherChild` will also be renamed to `TopLevel.Renamed.AnotherChild`

### TODO

* [ ] Add some tests for `handlers/rename.ex`

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
@@ -1,0 +1,178 @@
+defmodule Lexical.RemoteControl.CodeIntelligence.Rename do
+  alias Lexical.Ast
+  alias Lexical.Ast.Analysis
+  alias Lexical.Document
+  alias Lexical.Document.Edit
+  alias Lexical.Document.Line
+  alias Lexical.Document.Position
+  alias Lexical.RemoteControl.CodeIntelligence.Entity
+  alias Lexical.RemoteControl.Search.Store
+  require Logger
+
+  import Line
+
+  @spec rename(Analysis.t(), Position.t(), String.t()) ::
+          {:ok, %{Lexical.uri() => [Edit.t()]}} | {:error, term()}
+  def rename(%Analysis{} = analysis, %Position{} = position, new_name) do
+    with {:ok, entity, range} <- resolve_module(analysis, position) do
+      edits =
+        analysis.document
+        |> search_related_candidates(position, entity, range)
+        |> to_uri_with_changes(new_name)
+
+      {:ok, edits}
+    end
+  end
+
+  defp resolve_module(analysis, position) do
+    case Entity.resolve(analysis, position) do
+      {:ok, {module_or_struct, module}, range} when module_or_struct in [:struct, :module] ->
+        {:ok, module, range}
+
+      {:ok, other, _} ->
+        {:error, {:unsupported_entity, other}}
+
+      {:error, reason} ->
+        {:error, reason}
+    end
+  end
+
+  defp search_related_candidates(document, position, entity, range) do
+    cursor_entity_string = cursor_entity_string(range)
+
+    entities =
+      exacts(entity, cursor_entity_string)
+
+    # Users won't always want to rename descendants of a module.
+    # For instance, when there are no more submodules after the cursor.
+    # like: `defmodule TopLevel.Mo|dule do`
+    # in most cases, users only want to rename the module itself.
+    #
+    # However, there's an exception when the cursor is in the middle,
+    # such as `Top.Mo|dule.ChildModule`. If we rename it to `Top.Renamed.Child`,
+    # it would be natural to also rename `Module.ChildModule` to `Renamed.Child`.
+    if at_the_middle_of_module?(document, position, range) do
+      entities ++ descendants(entity, cursor_entity_string)
+    else
+      entities
+    end
+  end
+
+  defp at_the_middle_of_module?(document, position, range) do
+    range_text = range_text(range)
+
+    case Ast.surround_context(document, position) do
+      {:ok, %{context: {:alias, alias}}} ->
+        String.length(range_text) < length(alias)
+
+      _ ->
+        false
+    end
+  end
+
+  defp cursor_entity_string(range) do
+    # Parent.|Module -> Module
+    range
+    |> range_text()
+    |> String.split(".")
+    |> List.last()
+  end
+
+  defp descendants(entity, cursor_entity_string) do
+    entity_string = inspect(entity)
+    prefix = "#{entity_string}."
+
+    case Store.prefix(prefix, subject: [:definition, :reference]) do
+      {:ok, results} ->
+        filtered =
+          results
+          |> Enum.filter(fn result ->
+            range_text = range_text(result.range)
+            String.contains?(range_text, cursor_entity_string)
+          end)
+          |> adjust_range(entity)
+
+        filtered
+
+      _ ->
+        []
+    end
+  end
+
+  defp exacts(entity, cursor_entity_string) do
+    entity_string = inspect(entity)
+
+    case Store.exact(entity_string, subject: [:definition, :reference]) do
+      {:ok, results} ->
+        filtered =
+          Enum.filter(results, fn result ->
+            range_text = range_text(result.range)
+            String.contains?(range_text, cursor_entity_string)
+          end)
+
+        filtered
+
+      _ ->
+        []
+    end
+  end
+
+  defp adjust_range(entries, entity) do
+    for entry <- entries do
+      location = {entry.range.start.line, entry.range.start.character}
+      uri = Document.Path.ensure_uri(entry.path)
+
+      case resolve_entity_range(uri, location, entity) do
+        {:ok, range} ->
+          %{entry | range: range}
+
+        :error ->
+          :error
+      end
+    end
+    |> Enum.reject(&(&1 == :error))
+  end
+
+  defp resolve_entity_range(uri, location, entity) do
+    {line, character} = location
+
+    with {:ok, document} <- Document.Store.open_temporary(uri),
+         position = Position.new(document, line, character),
+         analysis = Ast.analyze(document),
+         {:ok, result, range} <- resolve_module(analysis, position) do
+      if result == entity do
+        {:ok, range}
+      else
+        result_length = result |> inspect() |> String.length()
+        # Move the cursor the next part:
+        # `|Parent.Next.Target.Child` -> 'Parent.|Next.Target.Child' -> 'Parent.Next.|Target.Child'
+        resolve_entity_range(uri, {line, character + result_length + 1}, entity)
+      end
+    else
+      _ ->
+        Logger.error("Failed to find entity range for #{inspect(uri)} at #{inspect(location)}")
+        :error
+    end
+  end
+
+  defp to_uri_with_changes(results, new_name) do
+    Enum.group_by(
+      results,
+      fn result -> Document.Path.ensure_uri(result.path) end,
+      fn result ->
+        cursor_entity_length = result.range |> cursor_entity_string() |> String.length()
+        # e.g: `Parent.|ToBeRenameModule`, we need the start position of `ToBeRenameModule`
+        start_character = result.range.end.character - cursor_entity_length
+        start_position = %{result.range.start | character: start_character}
+
+        new_range = %{result.range | start: start_position}
+        Edit.new(new_name, new_range)
+      end
+    )
+  end
+
+  defp range_text(range) do
+    line(text: text) = range.end.context_line
+    String.slice(text, range.start.character - 1, range.end.character - range.start.character)
+  end
+end
--- a/apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
@@ -18,7 +18,7 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Rename do
       edits =
         analysis.document
         |> search_related_candidates(position, entity, range)
-        |> to_uri_with_changes(new_name)
+        |> to_edits_by_uri(new_name)
 
       {:ok, edits}
     end
@@ -82,80 +82,62 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Rename do
     entity_string = inspect(entity)
     prefix = "#{entity_string}."
 
-    case Store.prefix(prefix, subject: [:definition, :reference]) do
-      {:ok, results} ->
-        filtered =
-          results
-          |> Enum.filter(fn result ->
-            range_text = range_text(result.range)
-            String.contains?(range_text, cursor_entity_string)
-          end)
-          |> adjust_range(entity)
-
-        filtered
-
-      _ ->
-        []
-    end
+    prefix
+    |> Store.prefix(subject: [:definition, :reference])
+    |> Enum.filter(&entry_matching?(&1, cursor_entity_string))
+    |> adjust_range(entity)
   end
 
   defp exacts(entity, cursor_entity_string) do
     entity_string = inspect(entity)
 
-    case Store.exact(entity_string, subject: [:definition, :reference]) do
-      {:ok, results} ->
-        filtered =
-          Enum.filter(results, fn result ->
-            range_text = range_text(result.range)
-            String.contains?(range_text, cursor_entity_string)
-          end)
+    entity_string
+    |> Store.exact(subject: [:definition, :reference])
+    |> Enum.filter(&entry_matching?(&1, cursor_entity_string))
+  end
 
-        filtered
-
-      _ ->
-        []
-    end
+  defp entry_matching?(entry, cursor_entity_string) do
+    range_text = range_text(entry.range)
+    String.contains?(range_text, cursor_entity_string)
   end
 
   defp adjust_range(entries, entity) do
-    for entry <- entries do
-      location = {entry.range.start.line, entry.range.start.character}
-      uri = Document.Path.ensure_uri(entry.path)
-
-      case resolve_entity_range(uri, location, entity) do
-        {:ok, range} ->
-          %{entry | range: range}
-
-        :error ->
-          :error
-      end
+    for entry <- entries,
+        uri = Document.Path.ensure_uri(entry.path),
+        {:ok, range} = resolve_entity_range(uri, entry.range.start, entity) do
+      %{entry | range: range}
     end
-    |> Enum.reject(&(&1 == :error))
   end
 
-  defp resolve_entity_range(uri, location, entity) do
-    {line, character} = location
-
+  defp resolve_entity_range(uri, position, entity) do
     with {:ok, document} <- Document.Store.open_temporary(uri),
-         position = Position.new(document, line, character),
          analysis = Ast.analyze(document),
          {:ok, result, range} <- resolve_module(analysis, position) do
       if result == entity do
         {:ok, range}
       else
-        result_length = result |> inspect() |> String.length()
-        # Move the cursor the next part:
+        last_part_length = result |> last_part() |> String.length()
+        # Move the cursor to the next part:
         # `|Parent.Next.Target.Child` -> 'Parent.|Next.Target.Child' -> 'Parent.Next.|Target.Child'
-        resolve_entity_range(uri, {line, character + result_length + 1}, entity)
+        character = position.character + last_part_length + 1
+        position = Position.new(document, position.line, character)
+        resolve_entity_range(uri, position, entity)
       end
     else
       _ ->
-        Logger.error("Failed to find entity range for #{inspect(uri)} at #{inspect(location)}")
+        Logger.error("Failed to find entity range for #{inspect(uri)} at #{inspect(position)}")
         :error
     end
   end
 
-  defp to_uri_with_changes(results, new_name) do
+  defp last_part(entity) when is_atom(entity) do
+    entity
+    |> inspect()
+    |> String.split(".")
+    |> List.last()
+  end
+
+  defp to_edits_by_uri(results, new_name) do
     Enum.group_by(
       results,
       fn result -> Document.Path.ensure_uri(result.path) end,
--- a/apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
@@ -24,14 +24,15 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Rename do
     end
   end
 
-  @spec supported?(Analysis.t(), Position.t()) :: boolean()
-  def supported?(%Analysis{} = analysis, %Position{} = position) do
+  @spec prepare(Analysis.t(), Position.t()) :: {:ok, String.t(), Range.t()} | {:error, term()}
+  def prepare(%Analysis{} = analysis, %Position{} = position) do
     case resolve_module(analysis, position) do
-      {:ok, _, _} ->
-        true
+      {:ok, _, range} ->
+        cursor_entity = cursor_entity_string(range)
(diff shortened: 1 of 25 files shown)


apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -0,0 +1,151 @@
+defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
```
scottming, line 1: I copied this module from the server app.

The reason we put it on the server in the first place was because of `Document.Store`, but now it's time to move it over, which will make it easier for us to test, and a lot of remote control functionality will depend on it as well.

I'll be happy to move relative tests to `remote_control` app after @zachallaun 's PR merging.

zachallaun: Sounds good! Agreed that it makes sense in remote control.

apps/server/lib/lexical/server/iex/helpers.ex
zachallaun, line 1: I'm not sure these changes are correct. For instance, if you do `iex -S mix`, instead of running the debug repl, Lexical won't have been namespaced yet and this will fail. 

@scohen had an idea a little while ago that I don't think ever made it in, but maybe now's a good time. IIRC, it was to alias `LXical` to `Lexical` if `LXical` is loaded. You'd have to check that it's loaded in the macro and conditionally include it. Something like:

```elixir
namespace_alias =
  if Code.ensure_loaded?(LXical) do
    quote(do: alias LXical, as: Lexical)
  end

quote do
  unquote(namespace_alias)
  ...
end
```

apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
```diff
@@ -0,0 +1,164 @@
...
+  @spec rename(Document.t(), Position.t(), String.t()) ::
+          {:ok, %{Lexical.uri() => [Edit.t()]}} | {:error, term()}
+  def rename(%Document{} = document, %Position{} = position, new_name) do
+    with {:ok, entity, range} <- resolve(document, position),
+         {:ok, results} <- search(document, position, entity, range) do
+      to_uri_with_changes(results, new_name)
+    end
+  end
+
+  defp resolve(document, position) do
```
zachallaun, line 22:
```suggestion
  defp resolve_module(document, position) do
```

apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
```diff
@@ -0,0 +1,164 @@
...
+    case Entity.resolve(document, position) do
+      {:ok, {:module, module}, range} -> {:ok, module, range}
+      {:error, error} -> {:error, error}
```
zachallaun, line 25: While the other PR hasn't landed yet, I think this should be a catch-all so that it doesn't have to change (much) when it does. 

```suggestion
      _ -> {:error, :unsupported_entity}
```

zachallaun: Also, just as a note/reminder, when the other PR lands, resolve will also potentially return `:struct`, which should be included for renaming. (It returns `:struct` for cursor positions like `%|MyStruct{}`.)

zachallaun (commented):
Haven't finished review yet, but have to run for a bit and wanted to get this thought down: Does this handle renaming an alias? For instance, if I have:

```elixir
alias Future.Code, as: Code
```

And then I want to rename `Code` to `MyFutureCode` (contrived example), the rename should only occur within the scope that the alias is valid. I think this is probably something that needs to be handled/considered by the indexer -- just want to make sure it is.

zachallaun:
Please ignore this:

> Does this handle renaming an alias?

Made my way to the tests and I see that it does :) Apologies for not reading those first.

apps/remote_control/test/lexical/remote_control/code_intelligence/rename_test.exs
zachallaun, line 1: The test coverage here looks really good, but I found some of them hard to read/interpret. I think that's primarily because the assertions use `=~`, and while I can't find any cases where it's actually ambiguous, it can be a little hard to see exactly which part is being asserted on. IMO this would be more clear if we were more verbose:

```elixir
test "example" do
  code = ~q[
    defmodule TopLevel do
      alias Foo.Bar, as: FooBar
      |FooBar
    end
  ]

  renamed = ~q[
    defmodule TopLevel do
      alias Foo.Bar, as: Renamed
      Renamed
    end
  ]

  assert rename(code, "Renamed") == expected
end
```

apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
```diff
@@ -0,0 +1,164 @@
...
+    range
+    |> range_text()
+    |> String.split(".")
+    |> List.last()
+  end
+
+  defp search_descendants(entity, cursor_entity_string) do
+    entity_string = inspect(entity)
+
+    case Store.fuzzy(entity_string, subject: [:definition, :reference]) do
```
scohen, line 62: I think you should be using `exact` rather than fuzzy here. Fuzzy will return a lot more results.

apps/remote_control/test/lexical/remote_control/code_intelligence/rename_test.exs
scottming: I think this is a good point, and I will try to make the testing more focused.

For example, in some cases, there is no need for both the definition and reference to appear at the same time. The new name should be uniformly `Renamed`. 

And do you think in the below pictures, the bottom is easier to read/interpret than the top?

image

To be honest I don't think it will, because it's a bit faster for us to locate `alias` and `defmodule` than to locate specific modules.

zachallaun: In that particular case, I do agree that the first one is probably a little easier to read. For shorter ones, I think being fully written out is helpful. I trust your judgement!

scottming: @zachallaun yeah, I pushed a change, when for the nested module renaming, I used this way.

zachallaun: Looks good!

apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
scottming: But I'm looking for `descendants`, is there a better way than fuzzy?

scottming:
I think this pull request is ready for review. The core of the renaming process is to identify eligible candidates, adjust the range, and combine them into a structure of `%{ uri => changes }`.

apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
```diff
@@ -0,0 +1,178 @@
...
+  alias Lexical.RemoteControl.CodeIntelligence.Entity
+  alias Lexical.RemoteControl.Search.Store
+  require Logger
+
+  import Line
+
+  @spec rename(Analysis.t(), Position.t(), String.t()) ::
+          {:ok, %{Lexical.uri() => [Edit.t()]}} | {:error, term()}
+  def rename(%Analysis{} = analysis, %Position{} = position, new_name) do
+    with {:ok, entity, range} <- resolve_module(analysis, position) do
```
zachallaun, line 17: What do you think about replacing `resolve_module` with a new `Entity.resolve/3` API that would work like:

```suggestion
    with {:ok, entity, range} <- Entity.resolve(analysis, position, [:module, :struct]) do
```

Inside `Entity`, I think the implementation would just be:

```elixir
def resolve(%Analysis{} = analysis, %Position{} = position, kinds) when is_list(kinds) do
  case resolve(analysis, position) do
    {:error, _} = error ->
      error

    entity when is_tuple(entity) and elem(entity, 0) in kinds ->
      entity

    _ ->
      {:error, :not_found}
  end
end
```

apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
```diff
@@ -0,0 +1,178 @@
...
+    prefix = "#{entity_string}."
+
+    case Store.prefix(prefix, subject: [:definition, :reference]) do
+      {:ok, results} ->
+        filtered =
+          results
+          |> Enum.filter(fn result ->
+            range_text = range_text(result.range)
+            String.contains?(range_text, cursor_entity_string)
+          end)
```
zachallaun, line 92: This particular filter happens at least three times in this file, I think -- probably worth extracting into a helper.

```suggestion
          |> entries_matching(cursor_entity_string)
```

apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
```diff
@@ -0,0 +1,178 @@
...
+      case resolve_entity_range(uri, location, entity) do
+        {:ok, range} ->
+          %{entry | range: range}
+
+        :error ->
+          :error
+      end
+    end
+    |> Enum.reject(&(&1 == :error))
+  end
```
zachallaun, line 134: Two suggestions here:

1. I don't think `location` serves any purpose -- `resolve_entity_range` can just take `entry.range.start`.
2. This can be cleaned up nicely by using the filtering feature of `for`.

```suggestion
  defp adjust_range(entries, entity) do
    for entry <- entries,
        uri = Document.Path.ensure_uri(entry.path),
        {:ok, range} = resolve_entity_range(uri, entry.range.start, entity) do
      %{entry | range: range}
    end
  end
```

apps/remote_control/test/lexical/remote_control/code_intelligence/rename_test.exs
```diff
@@ -0,0 +1,319 @@
...
+  import CursorSupport
+  import Lexical.Test.EventualAssertions
+  import Fixtures
+
+  use ExUnit.Case
+
+  setup_all do
+    project = project()
+    RemoteControl.set_project(project)
+    start_supervised!(Document.Store)
```
zachallaun, line 22: This will be necessary to use `Document.Store.fetch(uri, :analysis)` as I suggested elsewhere.

```suggestion
    start_supervised!(Lexical.Server.Application.document_store_child_spec())
```

apps/server/lib/lexical/server/provider/handlers/rename.ex
```diff
@@ -0,0 +1,39 @@
...
+  alias Lexical.Protocol.Requests.Rename
+  alias Lexical.Protocol.Responses
+  alias Lexical.Protocol.Types.Workspace.Edit
+  alias Lexical.RemoteControl.Api
+  alias Lexical.Server.Provider.Env
+  require Logger
+
+  def handle(%Rename{} = request, %Env{} = env) do
+    case Document.Store.fetch(request.document.uri, :analysis) do
+      {:ok, _document, %Ast.Analysis{} = analysis} ->
```
zachallaun, line 13: Analysis always succeeds, so we need to look at `valid?` on the analysis to actually get to the error case here.

```suggestion
      {:ok, _document, %Ast.Analysis{valid?: true} = analysis} ->
```

apps/server/lib/lexical/server/provider/handlers/rename.ex
zachallaun, line 1: Since we don't support renaming every type of entity (currently only modules and structs), should we also be using [Prepare Rename](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareRename) for a better UX? Reading through the spec, I *think* that editors will issue `textDocument/prepareRename` when you initiate the rename operation, and we can return `null` immediately to indicate that something doesn't support renaming.

zachallaun (changes requested):

apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
scottming: good catch, thanks.

apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
```diff
@@ -0,0 +1,178 @@
...
+          Enum.filter(results, fn result ->
+            range_text = range_text(result.range)
+            String.contains?(range_text, cursor_entity_string)
+          end)
+
+        filtered
+
+      _ ->
+        []
+    end
```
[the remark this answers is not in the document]
scohen: that's to comport with the backend behaviour. some of the backends we tried could return error tuples. The ETS backend doesn't, so we can remove the tuples from the behavour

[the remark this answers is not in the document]
scohen: Yeah, that's fine.

apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
```diff
@@ -0,0 +1,178 @@
...
+    |> Enum.reject(&(&1 == :error))
+  end
+
+  defp resolve_entity_range(uri, location, entity) do
+    {line, character} = location
+
+    with {:ok, document} <- Document.Store.open_temporary(uri),
+         position = Position.new(document, line, character),
+         analysis = Ast.analyze(document),
+         {:ok, result, range} <- resolve_module(analysis, position) do
```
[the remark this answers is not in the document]
scohen: We need to be sure that the analysis is also purged when a temporarily opened document is closed.

apps/server/lib/lexical/server/iex/helpers.ex
scottming: we can do this in another PR, I removed these changes from this PR.

apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
scottming: haha, Just twice, but I prefer to abstract a function for the item

apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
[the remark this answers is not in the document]
scottming: Does this point relate to this PR? I don't quite understand this comment. Could you elaborate?

apps/remote_control/test/lexical/remote_control/code_intelligence/rename_test.exs
scottming: good suggestion, but we can't just use that function, since here is in  `remote_control` app

apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
scottming: I think we can do this in the future, but for now, I prefer to keep it in this file, as this API is not used elsewhere.

scottming:
### PrepareRename and Rename in *Vscode*


https://github.com/lexical-lsp/lexical/assets/12830256/9f66a703-e9d2-4d9d-95fb-8cdec88a2fa9

### in *Neovim*

https://github.com/lexical-lsp/lexical/assets/12830256/88266e62-0dbf-47b6-8056-9f9a4a09eb2c

@zachallaun I've applied most of your suggestions. Please review them again when you have time. If there are no other issues, then I will add some integration tests for the handlers.

apps/server/lib/lexical/server/provider/handlers/rename.ex
scottming: Good idea, added that.

apps/remote_control/test/lexical/remote_control/code_intelligence/rename_test.exs
```diff
@@ -0,0 +1,338 @@
...
+    uri = subject_uri(project)
+
+    on_exit(fn ->
+      Document.Store.close(uri)
+    end)
+
+    %{uri: uri}
+  end
+
+  describe "rename exact module" do
```
scohen, line 49: nit: can you add the word `is` after the word `cusor` to all the test names?

apps/remote_control/test/lexical/remote_control/code_intelligence/rename_test.exs
```diff
@@ -0,0 +1,338 @@
...
+          |FooBar
+        end
+        ]
+        |> rename("Renamed")
+
+      assert result =~ ~S[alias Foo.Bar, as: Renamed]
+      assert result =~ ~S[  Renamed]
+    end
+
+    test "succeeds when the cursor at the alias_ased child" do
```
scohen, line 125:
```suggestion
    test "succeeds when the cursor is at the aliased child" do
```

apps/remote_control/test/lexical/remote_control/code_intelligence/rename_test.exs
```diff
@@ -0,0 +1,338 @@
...
+          end
+        ]
+        |> rename("Renamed")
+
+      assert result =~ ~S[defmodule TopLevel.Foo.Renamed]
+
+      assert result =~ ~S[  Parent.Renamed]
+    end
+
+    test "only rename aliased when the cursor at the alias_ased" do
```
scohen, line 143:
```suggestion
    test "only rename aliased when the cursor is at the aliased" do
```

apps/remote_control/test/lexical/remote_control/code_intelligence/rename_test.exs
```diff
@@ -0,0 +1,338 @@
...
+      {:ok, result} = ~q[
+        defmodule TopLevel.Module do # ✓
+        end
+
+        defmodule TopLevel.Module.Another do # x
+          alias TopLevel.|Module
+        end
+      ] |> rename("Renamed")
+
+      refute result =~ ~S[defmodule TopLevel.Renamed.Another]
```
scohen, line 185: i would definitely expect the module Toplevel.Module.Another to be renamed to `Toplevel.Renamed.Another`

scohen (commented):
My main concern her is if we could simplify things by making subjects charlists.

apps/remote_control/lib/lexical/remote_control/search/store/backends/ets/schemas/v1.ex
```diff
@@ -25,6 +25,14 @@ defmodule Lexical.RemoteControl.Search.Store.Backends.Ets.Schemas.V1 do
     :path
   ]
 
+  defkey :by_prefix, [
+    :prefix,
+    :type,
+    :subtype,
+    :elixir_version,
+    :erlang_version
+  ]
```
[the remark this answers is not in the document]
scottming: Are you referring to the "subject" in the key or the "subject" in the value? I think if we change the "subject" in the "by_subject" record key to a charlist instead of a binary, it would also achieve this goal. Currently, it is a binary.

https://github.com/lexical-lsp/lexical/blob/2b266ea7ff077cad578c73ee75d79c310a406864/apps/remote_control/lib/lexical/remote_control/search/store/backends/ets/schemas/v1.ex#L59

what we really need is something like this:

```elixir
 {:by_prefix, [70, 111, 111, 46, 66, 97, 114 | :_], :module, :definition,
 "1.15.7", "25.3.2.7"}
```

no matter is `by_prefix` or `by_subject`

apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
```diff
@@ -0,0 +1,176 @@
...
+    |> Store.prefix(subject: [:definition, :reference])
+    |> Enum.filter(&(entry_matching?(&1, cursor_entity_string) and has_dots_in_range?(&1)))
+    |> adjust_range(entity)
+  end
+
+  defp exacts(entity, cursor_entity_string) do
+    entity_string = inspect(entity)
+
+    entity_string
+    |> Store.exact(subject: [:definition, :reference])
```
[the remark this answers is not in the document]
scottming: yes, a dangerous typo.

apps/remote_control/lib/lexical/remote_control/search/store/backends/ets/state.ex
```diff
@@ -148,6 +166,13 @@ defmodule Lexical.RemoteControl.Search.Store.Backends.Ets.State do
   defp to_subject(atom) when is_atom(atom), do: inspect(atom)
   defp to_subject(other), do: to_string(other)
 
+  @dialyzer {:nowarn_function, to_prefix: 1}
+
+  defp to_prefix(prefix) when is_binary(prefix) do
+    [last_char | others] = prefix |> String.to_charlist() |> Enum.reverse()
+    others |> Enum.reverse() |> Enum.concat([last_char | :_])
```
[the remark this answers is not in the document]
scottming: yes.

apps/remote_control/test/lexical/remote_control/code_intelligence/rename_test.exs
```diff
@@ -0,0 +1,338 @@
...
+        |> rename("Renamed")
+
+      assert result =~ ~S[defmodule TopLevel.Renamed do]
+      assert result =~ ~S[defmodule Foo do # skip this]
+      assert result =~ ~S[alias TopLevel.Renamed.Foo]
+    end
+  end
+
+  describe "rename struct" do
+    test "succeeds when the cursor on the definition" do
```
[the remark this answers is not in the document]
scottming: When we hover, before the first letter of the definition, it returns the `module`. So, it is reasonable here. In this case, I think using `"at"` instead of `"before"` would be more appropriate.

apps/remote_control/test/lexical/remote_control/code_intelligence/rename_test.exs
scottming: done.🤗🤗

apps/remote_control/lib/lexical/remote_control/code_intelligence/rename.ex
```diff
@@ -0,0 +1,176 @@
...
+        :error
+    end
+  end
+
+  defp last_part(entity) when is_atom(entity) do
+    entity
+    |> inspect()
+    |> String.split(".")
+    |> List.last()
+  end
```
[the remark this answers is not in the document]
scottming: Currently, the rename operation does not consider Erlang modules. However, in terms of Erlang modules, the `inspect` function is more precise compared to A`tom.to_string`.

apps/remote_control/test/lexical/remote_control/code_intelligence/rename_test.exs
scottming: Here, I would like to preserve the functionality of renaming an exact module.

Consider the following scenario: Let's say we have two modules, `Project.Schema.User` and `Project.Schema`. If we find that `Project.Schema` is not clear in its meaning and rename it to `Project.EctoHelper`, should we also rename `Project.Schema.User` to `Project.EctoHelper.User`?

I believe the answer is no. Clearly, there are times when we want to rename a specific module, and there are times when we want to rename all modules that have that module as a prefix.

scottming:
@scohen I have applied almost all of your suggestions. Please take the time to review this PR again, especially the two commits related to the schema.

scohen:
@scottming I was very wary about this PR, the schema is going to change mightily so we can keep track of blocks and their contents, and I think you're going to need those changes.

scohen:
@scottming Think we should close this? It's likely the index changes will make this somewhat easier.

scottming:
It can only simplify part of things, but I think I can split this PR into multiple PRs.

PR #375, opened 2023-09-17
(Experiment) Make obvious syntax corrections if formatting fails

zachallaun:
This PR should be considered an experiment for now, but I'd love feedback on it because I think something like this is worthwhile.

The gist of it is: The Elixir formatter requires that the code be able to compile, so in the event of an error formatting, try to fix the code and re-format instead of immediately giving up. Right now I've implemented a single correction for `%{foo:bar}` (a space needs to be added after the colon when using keyword syntax).

### DIFF
--- a/apps/remote_control/test/lexical/remote_control/code_mod/format_test.exs
+++ b/apps/remote_control/test/lexical/remote_control/code_mod/format_test.exs
@@ -125,6 +125,28 @@ defmodule Lexical.RemoteControl.CodeMod.FormatTest do
     end
   end
 
+  describe "format/2 syntax correction" do
+    test "adds missing space after keyword", %{project: project} do
+      assert {:ok, result} = ~q[
+        %{foo:bar}
+      ] |> modify(project: project)
+
+      assert result == ~q[
+        %{foo: bar}
+      ]t
+    end
+
+    test "adds missing space after multiple keywords", %{project: project} do
+      assert {:ok, result} = ~q[
+        %{foo:bar, baz:buzz}
+      ] |> modify(project: project)
+
+      assert result == ~q[
+        %{foo: bar, baz: buzz}
+      ]t
+    end
+  end
+
   describe "emitting diagnostics" do
     setup [:with_real_project]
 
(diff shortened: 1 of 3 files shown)


apps/remote_control/test/lexical/remote_control/code_mod/format_test.exs
```diff
@@ -125,6 +125,28 @@ defmodule Lexical.RemoteControl.CodeMod.FormatTest do
...
+      ] |> modify(project: project)
+
+      assert result == ~q[
+        %{foo: bar}
+      ]t
+    end
+
+    test "adds missing space after multiple keywords", %{project: project} do
+      assert {:ok, result} = ~q[
+        %{foo:bar, baz:buzz}
```
scottming, line 141: I'm all for this change, but will its original format function work simultaneously?

zachallaun: I'm not sure I understand the question. What do you mean by "work simultaneously?"

The way this works now is by trying to format, and in the event of an error, checking if there's a "known fix" available for it. It then calls the same format function again recursively, so a second error may lead to a second fix, and so on. If the formatter returns successfully the first time, no fixes are attempted.

zachallaun: No worries! The recursion is a bit non-obvious, since it's re-wrapping the formatting function with itself before formatting the corrected code:

```elixir
wrap_with_syntax_corrections(formatter_fn).(corrected)
```

If there's a better/cleaner way to structure this, I'm all for it!

scohen:
My first take is that I think that this should be done by the formatter, and I wonder why it isn't. There's a bit of chesterton's fence going on here. Why didn't  the core team implement this?

zachallaun:
> My first take is that I think that this should be done by the formatter, and I wonder why it isn't. There's a bit of chesterton's fence going on here. Why didn't the core team implement this?

Been meaning to reply to this, but kept forgetting to.

I don't think this feature belongs in core Elixir because it's changing program behavior, and that's not something (IMO) that core tools should ever be doing -- the Elixir formatter should never change program behavior, because it doesn't know in what context it may be run.

We, however, know the context we're formatting in: an open editor buffer, for a file being actively edited. In all likelihood, if `foo:bar` is being changed by *our* formatter it's because it was just typed.

scohen:
It's not valid though. There is no behavior if it doesn't compile.

PR #377, opened 2023-09-19
Mnesia store

scohen:
There are two commits here, please review them separately.


apps/remote_control/lib/lexical/remote_control.ex
```diff
@@ -90,7 +90,8 @@ defmodule Lexical.RemoteControl do
   end
 
   defp app_globs do
-    Enum.map(@allowed_apps, fn app_name -> "/**/#{app_name}*/ebin" end)
+    app_globs = Enum.map(@allowed_apps, fn app_name -> "/**/#{app_name}*/ebin" end)
+    ["/**/priv" | app_globs]
```
scohen, line 94: This was a fun discovery. When we launch the _second_ vm , it wasn't adding the priv directories to the paths.

apps/remote_control/lib/lexical/remote_control/search/store/mnesia/connection.ex
```diff
@@ -0,0 +1,206 @@
...
+    port_args = [
+      "--name",
+      to_string(mnesia_node_name(project)),
+      "--cookie",
+      Node.get_cookie(),
+      "-e",
+      ~s[Node.connect(#{inspect(Node.self())})],
+      "--no-halt"
+      | path_args
+    ]
```
zachallaun, line 152: I noticed that you're calling `to_string(mnesia_node_name(project))`, but we're not stringifying the result of `Node.get_cookie()`, which returns the cookie as an atom. IMO, let's not worry about stringifying at all here, and let's just map `to_string` over the args in `RemoteControl.Port.open_elixir/2`.

```suggestion
    self = inspect(Node.self())

    port_args = [
      "--name",
      mnesia_node_name(project),
      "--cookie",
      Node.get_cookie(),
      "-e",
      "Node.connect(#{self})",
      "--no-halt"
      | path_args
    ]
```

apps/remote_control/lib/lexical/remote_control/port.ex
```diff
@@ -0,0 +1,81 @@
...
+  @doc """
+  Launches an executable in the project context via a port.
+  """
+  def open(%Project{} = project, executable, opts) do
+    {launcher, opts} = Keyword.pop_lazy(opts, :path, &path/0)
+
+    opts =
+      opts
+      |> Keyword.put_new_lazy(:cd, fn -> Project.root_path(project) end)
+      |> Keyword.update(:args, [executable], fn old_args -> [executable | old_args] end)
```
zachallaun, line 44: Ensure string args (see comment on usage in [`connection.ex`](apps/remote_control/lib/lexical/remote_control/search/store/mnesia/connection.ex))

```suggestion
      |> Keyword.update(:args, [executable], fn old_args ->
        Enum.map([executable | old_args], &to_string/1)
      end)
```

apps/remote_control/lib/lexical/remote_control/port.ex
```diff
@@ -0,0 +1,81 @@
...
+
+  def path({:unix, _}) do
+    with :non_existing <- :code.where_is_file(~c"port_wrapper.sh") do
+      :remote_control
+      |> :code.priv_dir()
+      |> Path.join("port_wrapper.sh")
+      |> Path.expand()
+    end
+    |> to_string()
+  end
```
zachallaun, line 71: Let's additionally raise a descriptive error for non-unix systems for now:

```elixir
def path(unsupported) do
  raise ArgumentError, "OS not currently supported: #{inspect(unsupported)}"
end
```

apps/remote_control/lib/lexical/remote_control/search/store.ex
```diff
@@ -34,12 +37,8 @@ defmodule Lexical.RemoteControl.Search.Store do
     GenServer.stop(__MODULE__)
   end
 
-  def metadata do
-    GenServer.call(__MODULE__, :metadata)
-  end
-
   def all do
-    GenServer.call(__MODULE__, :all)
+    GenServer.call(__MODULE__, :all, 20_000)
```
zachallaun, line 41: I think this timeout is here to match the default 20s for indexing, but that may not be necessary with #371.

apps/remote_control/lib/lexical/remote_control/search/store/backend.ex
```diff
@@ -11,18 +12,29 @@ defmodule Lexical.RemoteControl.Search.Store.Backend do
           subtypes: [Entry.entry_subtype()]
         }
 
-  @callback new(Path.t()) :: {:ok, load_state(), name()} | {:error, :any}
-  @callback new(Path.t(), version()) :: {:ok, load_state(), name()} | {:error, any()}
+  @callback new(Project.t()) :: {:ok, load_state()} | {:error, :any}
```
zachallaun, line 15: Is this `:any` meant to be `any()`?

```suggestion
  @callback new(Project.t()) :: {:ok, load_state()} | {:error, any()}
```

apps/remote_control/lib/lexical/remote_control/search/store/backend.ex
zachallaun, line 1: One thing I've been trying to keep an eye on is global state, as it makes it harder to transition our tests to eventually be async. I wonder if backends would benefit either from having their own state that gets passed into every callback, or maybe just having the project passed in every callback so that, minimally, the project can be used for naming.

This would allow the Mnesia backend to use this pattern that we use in a lot of other places:

```elixir
  def wait_for_start(%Project[} = project) do
    GenServer.call(name(project), :wait_for_start)
  end

  def initial_state(%Project{} = project) do
    GenServer.call(name(project), :initial_state)
  end

  def start_link(%Project{} = project) do
    GenServer.start_link(__MODULE__, [project], name: name(project))
  end

  defp name(%Project{} = project) do
    :"#{Project.name(project)}-search-backend-mnesia"
  end
```

apps/remote_control/lib/lexical/remote_control/search/store/mnesia.ex
```diff
@@ -0,0 +1,104 @@
...
+  def init([%Project{} = project]) do
+    {:ok, State.new(project), {:continue, :connect}}
+  end
+
+  def handle_continue(:connect, %State{} = state) do
+    case State.connect(state) do
+      {:connect_to_node, new_state} ->
+        {:noreply, new_state, {:continue, :connect_to_node}}
+    end
+  end
```
zachallaun, line 70: I think this should be `ensure_node` (mentioned elsewhere) and I also wonder if it should just return the new state instead of a tuple, since it only ever returns one thing.

```suggestion
  def handle_continue(:ensure_node, %State{} = state) do
    new_state = State.ensure_node(state)
    {:noreply, new_state, {:continue, :connect_to_node}}
  end
```

apps/remote_control/lib/lexical/remote_control/search/store/mnesia/schema.ex
```diff
@@ -0,0 +1,171 @@
+defmodule Lexical.RemoteControl.Search.Store.Mnesia.Schema do
+  alias Lexical.Project
+  alias Lexical.RemoteControl.Search.Indexer.Entry
+  alias Lexical.RemoteControl.Search.Store.Mnesia
+  alias Lexical.RemoteControl.Search.Store.Mnesia.State
+  alias Lexical.VM.Versions
+
+  import Record
+
+  @fields [:path, :reference, :value]
```
zachallaun, line 10: What do you think about renaming `:reference` to `:ref` to avoid any potential confusion between this and entry subtypes, which can be `:definition | :reference`?

```suggestion
  @fields [:path, :ref, :value]
```

apps/remote_control/lib/lexical/remote_control/search/store/mnesia/schema.ex
```diff
@@ -0,0 +1,171 @@
...
+    else
+      defaults
+      |> Keyword.put(:disc_copies, [])
+      |> Keyword.put(:ram_copies, [state.mnesia_node, Node.self()])
+    end
+  end
+
+  defp mnesia_dir(%Project{} = project) do
+    project
+    |> Project.workspace_path(Path.join(["indexes", "mnesia"]))
```
zachallaun, line 163:
```suggestion
    |> Project.workspace_path("indexes/mnesia")
```

apps/remote_control/lib/lexical/remote_control/search/store/mnesia/state.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Lexical.RemoteControl.Search.Store.Mnesia.State do
+  alias Lexical.Project
+  alias Lexical.Project
```
zachallaun, line 3:
```suggestion
```

apps/remote_control/lib/lexical/remote_control/search/store/mnesia/connection.ex
```diff
@@ -0,0 +1,206 @@
+defmodule Lexical.RemoteControl.Search.Store.Mnesia.Connection do
```
zachallaun, line 1: At the risk of nesting overload, maybe this should live under `State` to better indicate that it shouldn't be used directly (it's delegated to in `State`).

```suggestion
defmodule Lexical.RemoteControl.Search.Store.Mnesia.State.Connection do
```

apps/server/lib/lexical/server/state.ex
```diff
@@ -234,6 +234,7 @@ defmodule Lexical.Server.State do
         definition_provider: true,
         document_formatting_provider: true,
         hover_provider: true,
+        references_provider: true,
```
zachallaun, line 237: I think the `FindReferences` handler is still commented out, right? Is this line something from testing that got left in on accident?

zachallaun (commented):
I haven't looked closely at the tests yet, but made some comments from initial review. Honestly, I don't think the Mnesia stuff is *so* bad, though it's definitely a lot more of a song and dance than it should be.

apps/remote_control/lib/lexical/remote_control/search/store/mnesia/connection.ex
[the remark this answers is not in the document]
scohen: It's just to make the logic around who does the starting / stopping of the mnesia node simpler and lock free.

apps/remote_control/lib/lexical/remote_control/search/store/mnesia/schema.ex
scohen: I use `Path.join` to be cross-platform.

apps/remote_control/lib/lexical/remote_control/search/store/mnesia/connection.ex
scohen: worse, I want to move `Mnesia` to `backends` along with `Ets`

apps/server/lib/lexical/server/state.ex
scohen: ah yes. I'll remove.

apps/remote_control/lib/lexical/remote_control/port.ex
scohen: I'm not sure how I feel about this. 

Things can get turned into strings in surprising ways. I favor ensuring the caller does what they want to do with the arguments and pass them into this function.

apps/remote_control/lib/lexical/remote_control/search/store.ex
scohen: correct. Removing.

apps/remote_control/lib/lexical/remote_control/search/store/mnesia/schema.ex
zachallaun: Ah, good point! Maybe `Project.workspace_path` could accept a list as its second arg, but that could be discussed/changed later.

scohen:
Abandoning for ETS.

PR #383, opened 2023-09-19
Enabled warnings-as-errors on umbrella

scohen:


zachallaun (approved):

zachallaun:
Is the idea to fix the warnings that are now causing CI to fail separately?

scohen:
So I noticed that I have the habit of calling `Logger.warn`, which is deprecated in elixir 1.15, and that emits warnings in the server. We also have this flag when compiling the projects, just not when compiling the umbrella.

scohen:
...and by "I noticed", I mean it was pointed out to me

PR #385, opened 2023-09-21
Refactor client capability tracking

zachallaun:
While walking through how server and client capabilities were managed, I saw some opportunities to refactor the code that manages configuration and client capabilities. The main changes are:

* Rename `Lexical.Server.Configuration.supports?/1` to `client_supports?/1`
* Refactor `client_supports?/1` to not duplicate the set of keys being tracked
* Refactor `Lexical.Server.Configuration.Support` to generate the struct and `new/1` function based on a mapping.
* Change the capability attributes from the format `:thing?` to `:thing`. This is done because the `%Configuration.Support{}` struct is never intended to be accessed directly, but instead through `Configuration.client_supports?/1`, which reads better as `Configuration.client_supports?(:thing)`.

### DIFF
--- a/apps/server/lib/lexical/server/configuration/support.ex
+++ b/apps/server/lib/lexical/server/configuration/support.ex
@@ -1,65 +1,71 @@
 defmodule Lexical.Server.Configuration.Support do
+  @moduledoc false
+
   alias Lexical.Protocol.Types.ClientCapabilities
 
-  defstruct code_action_dynamic_registration?: false,
-            hierarchical_document_symbols?: false,
-            snippet?: false,
-            deprecated?: false,
-            tags?: false,
-            signature_help?: false,
-            work_done_progress?: false
+  # To track a new client capability, add a new field and the path to the
+  # capability in the `Lexical.Protocol.Types.ClientCapabilities` struct
+  # to this mapping:
+  @client_capability_mapping [
+    code_action_dynamic_registration: [
+      :text_document,
+      :code_action,
+      :dynamic_registration
+    ],
+    hierarchical_symbols: [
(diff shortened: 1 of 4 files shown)


apps/server/lib/lexical/server/configuration/support.ex
```diff
@@ -1,65 +1,65 @@
...
+      :window,
+      :work_done_progress
+    ]
+  ]
 
-    tags? =
-      client_capabilities
-      |> get_in([:text_document, :completion, :completion_item, :tag_support])
-      |> bool()
+  defstruct Enum.map(@client_capability_mapping, fn {field, _} -> {field, false} end)
```
scohen, line 48: i usually like to see the defstruct more spelled out than built.

scohen (approved):
I was going to say that this is refactoring a bit of code that isn't really used much, since we have no configuration, but this is pretty much the direction I wanted to take config in anyways. When we actually have configuration, it needs to be globally accessible, and this does that.

apps/server/lib/lexical/server/configuration/support.ex
zachallaun: I agree in the usual case. In this case, it seemed like a lot of duplication.

The alternative might be to write things out here explicitly, and then make it so that it errors in calls to `new` if there's any mismatch between the `@client_capability_mapping` and the fields in the struct.

zachallaun: I decided to just write it out. It's not that much, and any mismatch is guarded by the `struct(__MODULE__, defaults)` call.

scohen (approved):

PR #386, opened 2023-09-22
Explicitly implement protocol in completion modules

zachallaun:
Looking at the commit that introduced the `Impl` helper (4d1269e2), its inclusion is understandable. It replaced a previously existing `Translator` that was doing more heavy lifting, but it's now such a thin layer of abstraction that it does not seem worth hiding the intent, which is implementing `Lexical.Completion.Translatable`.

This commit also removes the `__deriving__/3` implementation in for the `Translatable` protocol because it was only used by the now-deleted `Impl` and I don't believe it would likely be used by external implementors of the protocol.


scohen (approved):
Agreed, it isn't saving much work any more.

zachallaun:
Whoops. Removed unrelated commit. Will merge once checks pass.

PR #387, opened 2023-09-22
CubDB Backend

scohen:
After the nightmare that was the mnesia backend, I figured I'd try cubdb. It's alright, the schema took some serious thinking (and I'm sure it can be improved), but it's reasonably fast enough for our purposes, and the code doesn't make my eyes bleed.

I'm going to be adding more tests, but it's mostly ready to be integrated.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/store/backends/cub.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/store/backends/cub.ex
@@ -33,7 +33,6 @@ defmodule Lexical.RemoteControl.Search.Store.Backends.Cub do
   end
 
   @impl Backend
-
   def destroy(%Project{} = project) do
     project
     |> cub_directory()
@@ -80,14 +79,11 @@ defmodule Lexical.RemoteControl.Search.Store.Backends.Cub do
     leader_name = leader_name(project)
 
     with :undefined <- :global.whereis_name(leader_name),
-         {:ok, :leader} <- :global.trans(leader_name, fn -> become_leader(project) end, [], 0) do
-      Process.flag(:trap_exit, true)
+         :ok <- become_leader(project) do
       {:noreply, initialize_cub(project)}
     else
       _ ->
-        leader_pid = :global.whereis_name(leader_name(project))
-        Process.monitor(leader_pid)
-        {:noreply, State.new_follower(project, leader_pid)}
+        {:noreply, follow_leader(project)}
     end
   end
 
@@ -165,19 +161,32 @@ defmodule Lexical.RemoteControl.Search.Store.Backends.Cub do
   end
 
   defp become_leader(%Project{} = project) do
-    case :global.register_name(leader_name(project), self(), &:global.random_notify_name/3) do
-      :yes ->
-        {:ok, :leader}
+    leader_name = leader_name(project)
+    :global.trans(leader_name, fn -> do_become_leader(project) end, [], 0)
+  end
 
-      :no ->
-        :aborted
+  defp do_become_leader(%Project{} = project) do
+    leader_name = leader_name(project)
+
+    with :undefined <- :global.whereis_name(leader_name),
+         :yes <- :global.register_name(leader_name, self(), &:global.random_notify_name/3) do
+      Process.flag(:trap_exit, true)
+      :ok
+    else
+      _ ->
+        :error
     end
   end
 
+  defp follow_leader(%Project{} = project) do
+    leader_pid = :global.whereis_name(leader_name(project))
+    Process.monitor(leader_pid)
+    State.new_follower(project, leader_pid)
+  end
+
   defp initialize_cub(%Project{} = project) do
-    with {:ok, cub_pid} <- CubDB.start_link(cub_directory(project), auto_compact: true) do
-      State.new_leader(project, cub_pid)
-    end
+    {:ok, cub_pid} = CubDB.start_link(cub_directory(project), auto_compact: true)
+    State.new_leader(project, cub_pid)
   end
 
   defp cub_directory(%Project{} = project) do
--- a/apps/remote_control/lib/lexical/remote_control/search/store/backends/cub.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/store/backends/cub.ex
@@ -1,0 +1,212 @@
+defmodule Lexical.RemoteControl.Search.Store.Backends.Cub do
+  alias Lexical.Project
+  alias Lexical.RemoteControl
+  alias Lexical.RemoteControl.Search.Store.Backend
+  alias Lexical.RemoteControl.Search.Store.Backends.Cub.State
+
+  use GenServer
+  @behaviour Backend
+
+  @impl Backend
+  def new(%Project{} = project) do
+    start_link(project)
+  end
+
+  @impl Backend
+  def prepare(pid) do
+    GenServer.call(pid, :prepare, :infinity)
+  end
+
+  @impl Backend
+  def sync(%Project{}) do
+    :ok
+  end
+
+  @impl Backend
+  def insert(entries) do
+    GenServer.call(genserver_name(), {:insert, entries}, :infinity)
+  end
+
+  @impl Backend
+  def drop do
+    GenServer.call(genserver_name(), :drop)
+  end
+
+  @impl Backend
+
+  def destroy(%Project{} = project) do
+    project
+    |> cub_directory()
+    |> File.rm_rf()
+  end
+
+  @impl Backend
+  def select_all do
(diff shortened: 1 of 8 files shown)


scohen:
Note: Cub requires exclusive access to the data by a single process. I've used `:global` to make this happen, and I've tested starting two stores in two terminals, then killing one or the other and have watched the ownership transfer back and forth. Pretty cool.

apps/remote_control/lib/lexical/remote_control/search/store/backends/cub/state.ex
```diff
@@ -0,0 +1,242 @@
...
+      CubDB.Tx.clear(transaction)
+
+      transaction =
+        Enum.reduce(keys_and_values, transaction, fn {key, value}, transaction ->
+          CubDB.Tx.put(transaction, key, value)
+        end)
+
+      {:commit, transaction, :ok}
+    end)
+  end
```
zachallaun, line 92: I wonder if a transaction is needed or desired here. Since only the leader can write anyways, the "worst case" is that a follower does a read in between clearing and replacing the entries. And you could argue that no result is better than a stale result anyways.

```suggestion
  def replace_all(%__MODULE__{} = state, entries) do
    :ok = drop(state)
    :ok = insert(state, entries)
    :ok
  end
```

apps/remote_control/lib/lexical/remote_control/search/store/backends/cub/state.ex
```diff
@@ -0,0 +1,242 @@
...
+
+      {subject_keys, ref_keys} =
+        transaction
+        |> CubDB.Tx.get(path_key, [])
+        |> Enum.unzip()
+
+      deleted_entries =
+        ref_keys
+        |> Enum.map(&CubDB.Tx.get(transaction, &1))
+        |> Enum.to_list()
```
zachallaun, line 106: I think the `Enum.to_list` is redundant.

```suggestion
      deleted_entries = Enum.map(ref_keys, &CubDB.Tx.get(transaction, &1))
```

apps/remote_control/lib/lexical/remote_control/search/store/backends/cub/state.ex
```diff
@@ -0,0 +1,242 @@
...
+      |> CubDB.Snapshot.select(min_key: min_subject_key(), max_key: max_subject_key())
+      |> Stream.flat_map(fn
+        {subject_key(subject: db_subject), entries} ->
+          Enum.filter(entries, fn entry ->
+            correct_version_range?(entry, versions) and
+              subject_matches?(db_subject, subject) and
+              type_and_subtype_match?(entry, type, subtype)
+          end)
+      end)
+      |> Enum.to_list()
```
zachallaun, line 133: As is, this could be simplified a bit:

```suggestion
      |> Enum.flat_map(fn {subject_key(subject: db_subject), entries} ->
        Enum.filter(entries, fn entry ->
          correct_version_range?(entry, versions) and
            subject_matches?(db_subject, subject) and
            type_and_subtype_match?(entry, type, subtype)
        end)
      end)
```

However, do we need `type_and_subtype_match?(entry, type, subtype)`? If we have a `type` and `subtype` already, can't we use that in the `select` so that we only have to check versions and subject?

apps/remote_control/lib/lexical/remote_control/search/store/backends/cub/state.ex
```diff
@@ -0,0 +1,242 @@
...
+    subject_key(
+      type: min_for_type(:atom),
+      subtype: min_for_type(:atom),
+      subject: min_for_type(:binary)
+    )
+  end
+
+  defp max_subject_key do
+    subject_key(type: max_for_type(:atom), subtype: max_for_type(:atom))
+  end
```
zachallaun, line 221: Does this work? I thought we needed to wrap binaries in `{binary}` so that the min key could be `{}` and the max `{"", ""}`.

zachallaun (commented):
Some initial comments from reading through the code. Looks a lot more sane than Mnesia though if the perf is acceptable!

apps/remote_control/lib/lexical/remote_control/search/store/backends/cub/state.ex
```diff
@@ -0,0 +1,242 @@
+defmodule Lexical.RemoteControl.Search.Store.Backends.Cub.State do
+  @moduledoc """
+  A cubdb schema and functions for data manipulation
+
+  CubDB is an... interesting database. It relies on erlang term ordering, which is documented below.
+  Furthermore, the positions of the records are significant in picking the ranges of things that we
+  must iterate over in order to limit the amount of stuff each `select` call sees.
```
zachallaun, line 7: Suggested rewrite:

CubDB is a KV store that stores data sorted using Erlang term ordering:

    number < atom < reference < fun < port < pid < tuple < map < nil < list < binary

The main query method is `select` using a min and max key. We use records as keys, which allow us to limit the data we scan by specifying exact values at the "front" of the record and relying on term order to construct wildcard queries at the "back" of the record. 

For instance, if we've stored the following keys:

    {:subject_key, :module, :definition, {"Foo"}}
    {:subject_key, :module, :definition, {"Bar"}}
    {:subject_key, :module, :reference, {"Foo"}}
    
We could select all module definitions using the min and max keys:

    iex> CubDB.select(db,
    ...>   min_key: {:subject_key, :module, :definition, {}},
    ...>   max_key: {:subject_key, :module, :definition, {"", ""}}
    ...> ) |> Enum.map(&elem(&1, 0))
    [{:subject_key, :module, :definition, {"Bar"}},
     {:subject_key, :module, :definition, {"Foo"}}]

(Note that we've wrapped the subject in a tuple. Because strings compare greater than all other terms, wrapping them in a tuple allows us to create max keys greater than all subjects by using a larger tuple, which always compares greater than a smaller tuple regardless of the contained terms.)

apps/remote_control/lib/lexical/remote_control/search/store/backends/cub/state.ex
zachallaun: Oh, right, it works because having a max key early in the key means later ones don't matter. It will matter, though, if we end up doing narrower selects specifying the type/subtype.

apps/remote_control/lib/lexical/remote_control/search/store/backends/cub/state.ex
zachallaun: BTW this is written assuming we need to do the "wrap binaries in a tuple" trick. Can remove/adjust that part if we don't actually need to do that.

apps/remote_control/lib/lexical/remote_control/search/store/backends/cub/state.ex
scohen: The transaction was to guard against no results being emitted, I don't want the lsp to have hiccups where you don't see completions, etc. how is not using a tx here better?

apps/remote_control/lib/lexical/remote_control/search/store/backends/cub/state.ex
scohen: Selects don't guarantee the subtype and types are always what you want. If the type is a wildcard, then you can get any subtype.

apps/remote_control/lib/lexical/remote_control/search/store/backends/cub/state.ex
scohen: We still might need to do that. This is a draft to get the ideas out there.

apps/remote_control/lib/lexical/remote_control/search/store/backends/cub/state.ex
scohen: The unit tests caught this

apps/remote_control/lib/lexical/remote_control/search/store/backends/cub.ex
```diff
@@ -0,0 +1,212 @@
...
+    # This clause means we were randomly notified by the call to
+    # :global.random_notify_name. We've been selected to be the leader!
+    new_state = become_leader(state.project)
+    {:noreply, new_state}
+  end
+
+  # Private
+
+  defp leader_name(%Project{} = project) do
+    {:cub, Project.name(project)}
```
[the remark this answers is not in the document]
scohen: This isn't that kind of name. It needs to be a two tuple

apps/remote_control/lib/lexical/remote_control/search/store/backends/cub/state.ex
scohen: Also, while reading the docs, it seemed to me that keeping filtering operations inside streams would be faster.

PR #388, opened 2023-09-24
Update allowed Elixir/Erlang versions to be consistent with docs

zachallaun:


zachallaun:
After discussing with @scohen, 1.13.0 should be fine, but all testing has historically been done on M-series Macs and current docs reflect versions that folks were able to actually build/run. I'm going to close this for now.

PR #389, opened 2023-09-25
Improve incompatible version errors on boot

zachallaun:
This commit makes a few improvements to the errors shown when Lexical boots with incompatible Elixir/Erlang versions:

* Consistent formatting for all errors
* Prevent an exception that would occur if using an incompatible major version (e.g. Elixir 1.12)
* If both Elixir and Erlang are incompatible, show both errors
* Show full range of compatible versions instead of just the range for the current major version

**Example before:**

image

**Example after:**

image

### DIFF
--- a/apps/common/lib/lexical/vm/versions.ex
+++ b/apps/common/lib/lexical/vm/versions.ex
@@ -64,7 +64,7 @@ defmodule Lexical.VM.Versions do
         |> Path.dirname()
         |> compatible?()
 
-      :error ->
+      _ ->
         false
     end
   end
@@ -196,20 +196,16 @@ defmodule Lexical.VM.Versions do
     Enum.join(normalized, ".")
   end
 
-  require Logger
-
   defp code_find_file(file_name) when is_binary(file_name) do
     file_name
     |> String.to_charlist()
     |> code_find_file()
   end
 
-  defp code_find_file(file_name) do
-    Logger.info("file name is #{file_name}")
-
+  defp code_find_file(file_name) when is_list(file_name) do
     case :code.where_is_file(file_name) do
       :non_existing ->
-        :error
+        {:error, {:file_missing, file_name}}
 
       path ->
         {:ok, List.to_string(path)}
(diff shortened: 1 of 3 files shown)


apps/common/lib/lexical/vm/versions.ex
```diff
@@ -196,20 +196,16 @@ defmodule Lexical.VM.Versions do
...
   end
 
-  defp code_find_file(file_name) do
-    Logger.info("file name is #{file_name}")
-
+  defp code_find_file(file_name) when is_list(file_name) do
     case :code.where_is_file(file_name) do
       :non_existing ->
-        :error
+        {:error, {:file_missing, file_name}}
```
zachallaun, line 208: This is done to be compatible with the typespec for `compiled/0`:

```elixir
@spec compiled() :: {:ok, t} | {:error, atom()}
```

scohen (commented):
Have you tried running this under incompatible erlang versions? I don't think it'll even boot.

zachallaun:
> Have you tried running this under incompatible erlang versions? I don't think it'll even boot.

You're right, it doesn't. That check was there previously, however, so I figured there might be some circumstances where it *does* boot and then fails later.

apps/server/lib/lexical/server/boot.ex
```diff
@@ -16,16 +16,31 @@ defmodule Lexical.Server.Boot do
...
+        |> Enum.join("\n\n")
+        |> halt()
+    end
+
     Application.ensure_all_started(:server)
   end
 
+  @doc false
+  def detect_errors do
+    List.wrap(packaging_errors()) ++ List.wrap(versioning_errors())
```
scohen, line 41: versioning_errors is always a list, i think that the naming indicates that `packaging_errors` should also be a list, but it can be nil. I think you should move `List.wrap` to `packaging_errors` so this is just `Enum.concat(packaging_errors(), versioning_errors())`

scohen (commented):
My main question is if this will even work for erlang version mismatches. I think it'll only work if we were compiled on an older version and are running on a newer version.

scohen (approved):

PR #394, opened 2023-09-29
Fix the `eex` compiled flaky test

scottming:
The previous CI was fine, and I rebased the latest branch, but the merge still fails the test

### DIFF
--- a/apps/remote_control/test/lexical/remote_control/build/document/compilers/eex_test.exs
+++ b/apps/remote_control/test/lexical/remote_control/build/document/compilers/eex_test.exs
@@ -1,6 +1,7 @@
 defmodule Lexical.RemoteControl.Build.Document.Compilers.EExTest do
   alias Lexical.Document
   alias Lexical.Plugin.V1.Diagnostic.Result
+  alias Lexical.RemoteControl.Build
   alias Lexical.RemoteControl.Build.CaptureServer
   alias Lexical.RemoteControl.Build.Document.Compilers
   alias Lexical.RemoteControl.Dispatch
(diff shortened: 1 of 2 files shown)


zachallaun (commented):
This doesn't make sense to me in a `setup` block unless these compiler options are being changed in individual tests.

I feel like this should be in a `setup_all` and we should be resetting whatever the previous compiler options were on exit, since this is shared global state:

```elixir
setup_all do
  prev_compiler_options = Code.compiler_options()
  Code.compiler_options(parser_options: [columns: true, token_metadata: true])

  on_exit(fn ->
    Code.compiler_options(prev_compiler_options)
  end)

  :ok
end
```

apps/remote_control/test/lexical/remote_control/build/document/compilers/eex_test.exs
```diff
@@ -26,6 +26,17 @@ defmodule Lexical.RemoteControl.Build.Document.Compilers.EExTest do
     Document.new("file:///file.eex", content, 0)
   end
 
+  setup_all do
+    prev_compiler_options = Code.compiler_options()
+    Code.compiler_options(parser_options: [columns: true, token_metadata: true])
```
scohen, line 31: should we just call `Build.State.set_compiler_options()` ?

apps/remote_control/test/lexical/remote_control/build/document/compilers/heex_test.exs
```diff
@@ -22,8 +22,15 @@ defmodule Lexical.RemoteControl.Build.Document.Compilers.HeexTest do
     Document.new("file:///file.heex", content, 0)
   end
 
-  setup do
+  setup_all do
+    prev_compiler_options = Code.compiler_options()
```
scohen, line 26: same comment

apps/remote_control/test/lexical/remote_control/build/document/compilers/eex_test.exs
scottming: yeah, that's better.

scohen (approved):

PR #395, opened 2023-09-29
Refactor shell scripts and add Docker-based integration tests

zachallaun:
* Add `bin/activate_version_manager.sh`

This script encapsulates detecting, finding, and activating a version manager. Currently asdf and rtx are supported, though more work likely needs to be done to make sure that they can be found/activated in more cases.

* Refactor `bin/start_lexical.sh`

This script now uses `activate_version_manager.sh` to activate the version manager. It additionally ensures that it exits with an error code if any commands run exit non-zero.

* `bin/boot.exs` halts after booting Lexical if `LX_HALT_AFTER_BOOT` is set

* Begin using `shellcheck` and `shfmt` for script linting and formatting

Both of these tools can be installed using asdf/rtx and have editor support through plugins. CI is not currently enforcing that these be used; we can decide on that later.

* Add `integration/` dir and integration tests for booting Lexical

Currently, these test system installations of elixir, asdf, and rtx, all using bash and running in a Docker container. I'd like to add tests using different shells (zsh/fish/etc.) at some point as well.

These tests are not currently running in CI. That will be added in a later commit.

### DIFF
--- a/apps/remote_control/priv/port_wrapper.sh
+++ b/apps/remote_control/priv/port_wrapper.sh
@@ -1,27 +1,10 @@
 #!/usr/bin/env bash
 
-detect_version_manager() {
-    if command -v asdf > /dev/null && asdf which elixir > /dev/null 2>&1 ; then
-        echo "asdf"
-    elif command -v rtx > /dev/null &&  rtx which elixir > /dev/null 2>&1 ; then
-        echo "rtx"
-    else
-        echo "not_detected"
-    fi
-}
+# Prevents zombie operating system processes
+# https://hexdocs.pm/elixir/Port.html#module-zombie-operating-system-processes
 
 # Start the program in the background
-case "$(detect_version_manager)" in
-    asdf)
-        asdf env erl exec "$@" &
-        ;;
-    rtx)
-        eval "$(rtx env -s bash)"
-        ;;
-    *)
-        exec "$@" &
-        ;;
-esac
+exec "$@" &
 pid1=$!
 
 # Silence warnings from here on
@@ -30,8 +13,8 @@ exec >/dev/null 2>&1
 # Read from stdin in the background and
 # kill running program when stdin closes
 exec 0<&0 $(
-  while read; do :; done
-  kill -KILL $pid1
+    while read; do :; done
+    kill -KILL $pid1
 ) &
 pid2=$!
 
--- a/integration/Dockerfile
+++ b/integration/Dockerfile
@@ -30,7 +30,6 @@ RUN integration/boot/set_up_asdf.sh
 COPY apps apps
 COPY bin bin
 COPY config config
-COPY priv priv
 COPY projects projects
 COPY mix* .
 
--- a/integration/Dockerfile
+++ b/integration/Dockerfile
@@ -1,0 +1,40 @@
+# Used to build a clean container for testing the Lexical boot sequence.
+#
+# Build: docker build -t lx -f integration/Dockerfile .
+# Run:   docker run -it lx
+
+ARG ELIXIR_VERSION=1.15.6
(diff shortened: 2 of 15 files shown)


integration/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+COPY integration/boot/set_up_rtx.sh integration/boot/set_up_rtx.sh
+RUN integration/boot/set_up_rtx.sh
+
+COPY integration/boot/set_up_asdf.sh integration/boot/set_up_asdf.sh
+RUN integration/boot/set_up_asdf.sh
+
+COPY apps apps
+COPY bin bin
+COPY config config
+COPY priv priv
```
Blond11516, line 33:
```suggestion
```
I had to remove this to make build work. I don't have a priv directory at the project root.

integration/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+  openssl \
+  libssl-dev
+
+WORKDIR /lexical
+
+COPY integration/boot/set_up_rtx.sh integration/boot/set_up_rtx.sh
+RUN integration/boot/set_up_rtx.sh
+
+COPY integration/boot/set_up_asdf.sh integration/boot/set_up_asdf.sh
+RUN integration/boot/set_up_asdf.sh
```
Blond11516, line 28: Installing Erlang twice in a row is rough 🙃 It's fine for local work since it's cached, but I think we'll probably want to push the image to Github Container Registry if/when we get to putting this in CI.

integration/test.sh
```diff
@@ -0,0 +1,101 @@
...
+
+    if [[ $1 != "" ]]; then
+        command="$1 && $command"
+    fi
+
+    docker run -i lx bash -c "$command" 2>&1
+    return $?
+}
+
+run_test() {
```
Blond11516, line 25: Should this go in `test_utils.sh`?

Blond11516 (commented):
Tested locally, works great!

This is really good stuff, it will make working on the boot process much easier (and less scary).

integration/Dockerfile
zachallaun: Thanks for pointing this out! I don't think the root `priv` ever actually needs to be copied. It only contains cached PLTs from running Dialyzer.

integration/test.sh
zachallaun: As it's currently written, I don't think so, because it calls `start_lexical` directly. I originally had a more abstracted version of it in `test_utils.sh`, but ended up moving it here because there was just a lot of nested output/exit code capture going on and I found it hard to follow/debug.

My inclination is to leave it here until some of this logic is needed elsewhere, then pull it out.

Blond11516: Sounds like a good plan 👍

apps/remote_control/priv/port_wrapper.sh
```diff
@@ -1,42 +1,21 @@
...
-# Clean up
-wait $pid1
-ret=$?
-kill -KILL $pid2
-exit $ret
+# Clean up and exit with wrapped program's exit code
+wait $wrapped_pid
+wrapped_ret=$?
+kill -KILL $monitor_pid
+exit $wrapped_ret
```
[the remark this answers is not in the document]
zachallaun: Yep, this was just my attempt to rewrite/reword things in a way that I thought would make more sense for someone coming to this for the first time.

zachallaun:
This branch fixes a bug that's currently on main that prevents certain environments using `rtx` from starting the remote node.

apps/remote_control/priv/port_wrapper.sh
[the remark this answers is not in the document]
scohen: Just link to the docs

scohen (approved):

PR #396, opened 2023-09-30
Fix Field parsing error for zed editor

scottming:
Zed already supports custom LS for Elixir. I have found an issue with Lexical during startup, and the log displays the following. 
```elixir
10:21:49.299 [info] Application lx_server started at :nonode@nohost
10:21:49.315 [error] Process #PID<0.140.0> terminating
** (UndefinedFunctionError) function nil.parse/1 is undefined
    nil.parse(nil)
    (lx_proto 0.3.0) lib/lexical/proto/field.ex:23: anonymous fn/4 in LXical.Proto.Field.extract/3
    (elixir 1.15.4) lib/enum.ex:4830: Enumerable.List.reduce/3
    (elixir 1.15.4) lib/enum.ex:2564: Enum.reduce_while/3
    (lx_proto 0.3.0) lib/lexical/proto/field.ex:22: LXical.Proto.Field.extract/3
    (lx_protocol 0.3.0) /Users/scottming/Code/lexical/apps/proto/lib/lexical/proto/macros/parse.ex:41: LXical.Protocol.Requests.Initialize.LSP.parse/1
    (lx_protocol 0.3.0) lib/lexical/protocol/requests.ex:9: LXical.Protocol.Requests.Initialize.parse/1
    (lx_server 0.3.0) lib/lexical/server/transport/std_io.ex:90: LXical.Server.Transport.StdIO.loop/3
Initial Call: LXical.Server.Transport.StdIO.init/1
Ancestors: [LXical.Server.Supervisor, #PID<0.133.0>]
```

However, I am currently unable to access detailed logs from the client. After printing it out, I discovered that the `processId` field is `nil`.

At the same time, I checked the LSP specification and found that `ProcessId` can be nullable: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize. It makes sense to fix this issue on the Lexical side.

### DIFF
--- a/apps/proto/lib/lexical/proto/field.ex
+++ b/apps/proto/lib/lexical/proto/field.ex
@@ -67,6 +67,10 @@ defmodule Lexical.Proto.Field do
     extract(alias_module.definition(), name, orig_value)
   end
 
+  def extract(nil, _name, orig_value) do
+    {:ok, orig_value}
+  end
+
   def extract(module, _name, orig_value)
       when is_atom(module) and module not in [:integer, :string, :boolean, :float] do
     module.parse(orig_value)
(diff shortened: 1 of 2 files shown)


scohen (commented):
can you please add some tests?

scohen (approved):

PR #406, opened 2023-10-12
Move `entity` module to `remote_control` app

scottming:
`resolve` does not depend on the project startup.

After moving it, firstly, it makes our testing more convenient, and secondly, it allows other modules of remote_control to reuse the resolve function of the entity.


zachallaun (approved):
Nice!

scohen (changes requested):
It is strange for something in remote control to call RemoteControl.API (this happens in Entity.definition, for example)
The stuff in there needs to be extracted and moved into the providers in the `server` app.

scottming:
> It is strange for something in remote control to call RemoteControl.API (this happens in Entity.definition, for example)
The stuff in there needs to be extracted and moved into the providers in the server app.

Yes, it does seem a bit strange.

In my new commit, I moved this part to the `definition.ex` file.

Additionally, I kept the logic for parsing ElixirSense results in there because once we can replace `ElixirSense,` this part of the logic will be unnecessary.

scohen (approved):

PR #407, opened 2023-10-12
Reorder the startup order of the children of `Server.Project.Supervisor`

scottming:
After we moved the dispatch, the first progress message appeared very slowly, taking about **10** seconds. 

This PR primarily aims to start the `progress` as the first process in `Lexical.Server.Project.Supervisor`, which will advance the appearance of the first progress message by approximately *4* seconds.

### Before

https://github.com/lexical-lsp/lexical/assets/12830256/2a48b0f4-e44d-417a-965a-588ec26dcb7e

### After

https://github.com/lexical-lsp/lexical/assets/12830256/65836121-6e0b-44e5-8f2e-1502023c9623

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control.ex
+++ b/apps/remote_control/lib/lexical/remote_control.ex
@@ -50,10 +50,10 @@ defmodule Lexical.RemoteControl do
   end
 
   defp start_net_kernel(%Project{} = project) do
-    :net_kernel.start([master_node_name(project)])
+    :net_kernel.start([manager_node_name(project)])
   end
 
-  def master_node_name(%Project{} = project) do
+  def manager_node_name(%Project{} = project) do
     :"manager-#{Project.name(project)}-#{Project.entropy(project)}@127.0.0.1"
   end
 
--- a/apps/remote_control/lib/lexical/remote_control.ex
+++ b/apps/remote_control/lib/lexical/remote_control.ex
@@ -50,7 +50,11 @@ defmodule Lexical.RemoteControl do
   end
 
   defp start_net_kernel(%Project{} = project) do
-    :net_kernel.start([:"manager-#{Project.name(project)}-#{Project.entropy(project)}@127.0.0.1"])
+    :net_kernel.start([master_node_name(project)])
+  end
+
+  def master_node_name(%Project{} = project) do
+    :"manager-#{Project.name(project)}-#{Project.entropy(project)}@127.0.0.1"
   end
 
   def ensure_apps_started(node, app_names) do
--- a/apps/remote_control/lib/lexical/remote_control/dispatch.ex
+++ b/apps/remote_control/lib/lexical/remote_control/dispatch.ex
@@ -44,11 +44,15 @@ defmodule Lexical.RemoteControl.Dispatch do
 
   # GenServer callbacks
 
-  def start_link(_) do
+  def start_link(opts) do
     case :gen_event.start_link(name()) do
       {:ok, pid} = success ->
         Enum.each(@handlers, &:gen_event.add_handler(pid, &1, []))
-        register_progress_listener()
+
+        if opts[:progress] do
+          register_progress_listener()
+        end
+
         success
 
       error ->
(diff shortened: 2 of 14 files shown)


apps/remote_control/lib/lexical/remote_control.ex
```diff
@@ -50,7 +50,11 @@ defmodule Lexical.RemoteControl do
   end
 
   defp start_net_kernel(%Project{} = project) do
-    :net_kernel.start([:"manager-#{Project.name(project)}-#{Project.entropy(project)}@127.0.0.1"])
+    :net_kernel.start([master_node_name(project)])
+  end
+
+  def master_node_name(%Project{} = project) do
```
zachallaun, line 56:
```suggestion
  def manager_node_name(%Project{} = project) do
```

apps/remote_control/test/lexical/remote_control/dispatch/handler_test.exs
```diff
@@ -4,8 +4,10 @@ defmodule Lexical.RemoteControl.Dispatch.HandlerTest do
 
   import Messages
   use ExUnit.Case
+  use Patch
 
   setup do
+    patch(Dispatch, :register_progress_listener, fn -> :ok end)
```
zachallaun, line 10: Looking at the other usage, I think this can be simplified?

```suggestion
    patch(Dispatch, :register_progress_listener, :ok)
```

apps/remote_control/lib/lexical/remote_control/dispatch.ex
```diff
@@ -63,4 +66,15 @@ defmodule Lexical.RemoteControl.Dispatch do
...
+
+  defp register_progress_listener do
+    register_listener(progress_pid(), [project_progress()])
+  end
+
+  defp progress_pid do
+    project = RemoteControl.get_project()
+    master_node_name = RemoteControl.master_node_name(project)
+    progress_name = :rpc.call(master_node_name, Lexical.Server.Project.Progress, :name, [project])
+    :rpc.call(master_node_name, Process, :whereis, [progress_name])
```
zachallaun, line 78: Worth adding a `Lexical.Server.Project.Progress.whereis(project)` to eliminate some of this back-and-forth?

```suggestion
    :rpc.call(master_node_name, Lexical.Server.Project.Progress, :whereis, [project])
```

apps/remote_control/lib/lexical/remote_control/dispatch.ex
```diff
@@ -46,6 +48,7 @@ defmodule Lexical.RemoteControl.Dispatch do
     case :gen_event.start_link(name()) do
       {:ok, pid} = success ->
         Enum.each(@handlers, &:gen_event.add_handler(pid, &1, []))
+        register_progress_listener()
```
zachallaun, line 51: One thing I don't love is having to patch `register_progress_listener` in so many tests. We're not using the `Dispatch.start_link` argument right now -- what if we added a `:progress` opt that defaults to `true` and only call `register_progress_listener` if it's set?

Then, in tests, where we don't want to register for progress, we can do:

```elixir
start_supervised!({Dispatch, progress: false})
```

zachallaun (commented):
One thing I don't love is having to patch `register_progress_listener` in so many tests. We're not using the `Dispatch.start_link` argument right now -- what if we added a `:progress` opt that defaults to

apps/remote_control/lib/lexical/remote_control/dispatch.ex
scottming: Good idea

apps/remote_control/lib/lexical/remote_control.ex
scohen: Thank you ;)

scohen (changes requested):
Basically, what @zachallaun said

apps/remote_control/lib/lexical/remote_control/dispatch.ex
scottming: I think we make `progress` default to `nil`, and set `true` in the `application.ex`, because most tests or processes do not need to care about the progress.

zachallaun (approved):

scohen (approved):

PR #408, opened 2023-10-12
Make the operational behavior of the ancestors of structures and modules more consistent

scottming:
Taking `Mix.Task.Compiler.Diagnostic` as an example, the references are inconsistent when the cursor is on `%Mix.Task.|Compiler.Diagnostic{}` compared to when it is on `Mix.Task.|Compiler.Diagnostic`.

I believe that maintaining consistency between them would provide a better user experience and make it easier to extend for future features, such as renaming.

CleanShot 2023-10-12 at 22 31 41@2x

CleanShot 2023-10-12 at 22 31 52@2x

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
@@ -94,24 +94,8 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
       resolve_struct(charlist, node_range, document, position)
     else
       _ ->
-        if follow_percent?(document, position) do
-          resolve_struct(charlist, node_range, document, position)
-        else
-          resolve_module(charlist, node_range, document, position)
-        end
+        resolve_module(charlist, node_range, document, position)
     end
-  end
-
-  # |%MyStruct{}
-  defp follow_percent?(document, position) do
-    after_cursor =
-      Document.fragment(
-        document,
-        position,
-        Position.new(document, position.line, position.character + 1)
-      )
-
-    after_cursor == "%"
   end
 
   defp resolve_struct(charlist, node_range, document, %Position{} = position) do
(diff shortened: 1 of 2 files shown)


apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -93,13 +93,36 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
          :struct <- kind_of_alias(path) do
       resolve_struct(charlist, node_range, document, position)
     else
-      _ -> resolve_module(charlist, node_range, document, position)
+      _ ->
+        if follow_percent?(document, position) do
```
scottming, line 97: we can't fetch the path for some code like this:

```elixir
      code = ~q[
        defmodule MyModule do
          |%MyStruct{}
        end
      ]
```

not sure if this is a bug of `path_at`

zachallaun: This does look like a bug in `path_at`. It should return the `%` node. I'll look into it.

zachallaun: This should be fixed by:

- #419

zachallaun (changes requested):
I'd like to see if we can get #419 merged first to hopefully simplify/remove the need for `follow_percent?`

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -89,11 +89,15 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
   end
 
   defp resolve_alias(charlist, node_range, document, position) do
-    with {:ok, path} <- Ast.path_at(document, position),
+    {{_line, start_column}, _} = node_range
+
+    with true <- not suffix_contains_module?(charlist, start_column, position),
```
zachallaun, line 94:
```suggestion
    with false <- suffix_contains_module?(charlist, start_column, position),
```

apps/remote_control/test/lexical/remote_control/code_intelligence/entity_test.exs
```diff
@@ -236,13 +236,13 @@ defmodule Lexical.RemoteControl.CodeIntelligence.EntityTest do
              """
     end
 
-    test "includes trailing module segments" do
+    test "shouldn't includes trailing module segments" do
```
zachallaun, line 239:
```suggestion
    test "shouldn't include trailing module segments" do
```

scohen (approved):
LGTM once you implement the suggestions.

PR #413, opened 2023-10-13
update the nix file hash

Leastrio:


zachallaun:
Caveat: I know nothing about nix.

I wonder if this is something that can be automated. When does this hash need to be updated? Should we set up a GitHub action to update it automatically whenever new commits land in `main`?

scohen:
@hauleth ☝️

hauleth:
> When does this hash need to be updated?

When there is change in `mix.lock`.

> Should we set up a GitHub action to update it automatically whenever new commits land in main?

I would need to check if that is possible. Right now I always have done that by hand. My workflow for that was:

- Change the hash in the `nix/lexical.nix` to something I know is wrong
- Try to build
- See failure and see what is the correct hash
- Copy it to the derivation

I would need to see if that process can be automated in any way, so it could be done automatically in the CI.

hauleth:
Maybe @NobbZ has some idea how this can be automated in other way than just running with fake hash and then grepping it out.

NobbZ:
Should definitely be possible to automate that.

I have to experiment with the tools available and this sources over the weekend.

I am not sure which do understand mix.

hauleth:
Ok, I have created hacky script that generate hash for the dependencies:

```sh
nix build --impure --expr '(builtins.getFlake (toString ./.)).packages.x86_64-linux.lexical.mixFodDeps.overrideAttrs (a: b: { outputHash = "sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="; })' 2>&1 | awk '/got:/ { print $2 }'
```

I couldn't (yet) find a way to do that without that pesky `--impure` flag, but so far that seems to do the job pretty well.

PR #416, opened 2023-10-13
chore: update Nix definition

hauleth:


zachallaun:
Closing as duplicate of #413 (we could use your advice there)

PR #417, opened 2023-10-13
chore: update Nix definition

hauleth:
Close #413

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,20 @@
+name: Nix derivation checks
+
+on:
+  push:
+    branches: ["main"]
+    paths: ["mix.lock"]
+
+jobs:
+  auto-update-nix-hash:
+    runs-on: ubuntu-latest
+    name: Auto update Nix hash
+    steps:
+    - uses: actions/checkout@v4
+    - uses: DeterminateSystems/nix-installer-action@main
+    - uses: DeterminateSystems/magic-nix-cache-action@main
+    - run: |
+        nix build --no-link .#__fodHashGen 2>&1 | awk '/gen:/ { print $2 }' > nix/hash
+        git config user.email ""
+        git config user.name "GitHub Action Bot"
+        git commit -m 'Update Nix hash of Mix deps' nix/hash && git push || true
--- a/flake.nix
+++ b/flake.nix
@@ -14,12 +14,18 @@
       let
         pkgs = import nixpkgs { inherit system; };
         erl = pkgs.beam.packages.erlang;
+        lexical = erl.callPackage ./nix/lexical.nix {};
       in
       {
-        packages = rec {
-          lexical = erl.callPackage ./nix/lexical.nix {};
+        packages = {
+          inherit lexical;
 
           default = lexical;
+
+          # Private package used to automatically generate hash for Mix deps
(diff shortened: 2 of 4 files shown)


.github/workflows/nix.yml
```diff
@@ -0,0 +1,20 @@
...
+    name: Auto update Nix hash
+    steps:
+    - uses: actions/checkout@v4
+    - uses: DeterminateSystems/nix-installer-action@main
+    - uses: DeterminateSystems/magic-nix-cache-action@main
+    - run: |
+        nix build --no-link .#__fodHashGen 2>&1 | awk '/gen:/ { print $2 }' > nix/hash
+        git config user.email ""
+        git config user.name "GitHub Action Bot"
+        git commit -m 'chore: update Nix FOD hash' nix/hash
```
zachallaun, line 19: Because we don't enforce conventional commits, I think I'd prefer this without the prefix.

```suggestion
        git commit -m 'Update Nix FOD hash' nix/hash
```

flake.nix
```diff
@@ -14,12 +14,19 @@
...
       {
-        packages = rec {
-          lexical = erl.callPackage ./nix/lexical.nix {};
+        packages = {
+          inherit lexical;
 
           default = lexical;
+
+          # Private package used to automatically generate hash for the Mix
+          # dependencies of the Lexical
```
zachallaun, line 26:
```suggestion
          # Private package used to automatically generate hash for Mix deps
```

.github/workflows/nix.yml
```diff
@@ -0,0 +1,20 @@
+name: Nix derivation checks
+
+on:
+  push:
+    paths: ["mix.lock"]
```
zachallaun, line 5: We probably want to only do this on `main`. I think this would do it?

```suggestion
on:
  push:
    paths: ["mix.lock"]
    branches: ["main"]
```

zachallaun (commented):
Thank you for figuring this out!

zachallaun (approved):
Looks good!

@scohen WDYT?

scohen:
Looks good to me.

.github/workflows/nix.yml
```diff
@@ -0,0 +1,20 @@
+name: Nix derivation checks
+
+on:
+  push:
+    branches: ["main"]
```
scohen, line 5: do you want to update the nix hash on every push to main, or when we do releases?

zachallaun: I think it's only pushes to main where mix.lock changed.

scohen: yes, but do we want all pushes to main that change the mix lock to update the nix flake, or should the flake get changed only when we release?

zachallaun: I think it should be any change on main to mix.lock, so that main is always usable for nix users.

scohen: i'm unfamiliar with how nix works, so ok.

zachallaun: My point earlier was just that I don't think it will be very noisy, since mix.lock doesn't change often.

hauleth: @scohen conceptually it is pretty easy - you point to the Git ref in this repo, it fetches project and build it. The thing is that it checks the dependencies hash against the fetched deps. So this simply makes it in a way that will allow to pin any (updated) commit after `mix.lock` was updated.

PR #420, opened 2023-10-15
i96: send and log messages

### LINKED ISSUE #96: Add ability to send messages to the server
Presently we don't have the ability to notify the user of... stuff. 

To fix this, we need to implement
[`window/showMessage`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessage)
[`window/showMessageRequest`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessageRequest)
[`window/logMessage`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_logMessage)

jollyjerr:
Heyo! I'm opening this up to start some conversation, but I'm definitely looking for pointers and guidance as I learn your codebase. I'm trying to take on a small chunk here to learn about how Lexical works so I can help out in larger ways down the road.

Taking a look at #96 I think showMessage and logMessage are fundamentally different operations than showMessageRequest, so I'm wondering if we could tackle show/log here and follow up with showMessageRequest which will likely need to spin up a process to handle the request/response lifecycle (similar to textDocument/codeAction followed by codeAction/resolve as zachallaun mentioned in Discord).

For my approach here, I'm thinking "log" does not really need to be implemented by each `Transport` because it was just using `Transport.write` at the end of the day. I pulled the existing log implementation out of the transport behavior and implemented `show` next to it.

I also fixed some small typos I noticed in some of the docs as I was learning.

Screenshot 2023-10-15 at 12 43 24 PM

closes #96


scottming:
I actually don't really understand the requirement and PR's solution, I would like to know, what are the difference and connections between the three, `Transport` , `Window` and `JsonRpc.Backend`?

apps/server/lib/lexical/server/window.ex
```diff
@@ -0,0 +1,37 @@
...
+  alias Lexical.Server.Transport
+
+  @type level :: :error | :warning | :info | :log
+
+  @spec log(level, String.t(), [{:label, String.t()}]) :: String.t()
+  def log(level, message, opts \\ [])
+
+  def log(level, message, opts) when level in [:error, :warning, :info, :log] do
+    formatted_message = format_message(message, opts)
+    log_message = apply(LogMessage, level, [formatted_message])
```
scottming, line 13: don't quite understand why `apply` here.

apps/server/lib/lexical/server/transport/std_io.ex
```diff
@@ -19,22 +17,24 @@ defmodule Lexical.Server.Transport.StdIO do
     loop([], device, callback)
   end
 
-  def write(io_device \\ :stdio, payload)
+  def write(payload, opts \\ [io_device: :stdio])
```
scottming, line 20: I'd prefer to keep the previous form, since it's consistent with `IO.write`

apps/server/lib/lexical/server/window.ex
```diff
@@ -0,0 +1,37 @@
+defmodule Lexical.Server.Window do
```
scohen, line 1: This being in the server app means we won't be able to use it in `remote_control`, which is where the vast majority of work happens. This might not be an issue, and I don't want to create a bunch of yak-shaving for this PR, but we might need to use a `RemoteControl.Dispatch` listener or handler to respond to messages coming from the `remote_control` project.

apps/server/lib/lexical/server/json_rpc_backend.ex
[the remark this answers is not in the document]
jollyjerr: I see this module was originally built to send logger logs to the client and is still in our `runtime.exs`. Did this never end up being used? I like the thought behind it!

```elixir
  Code.ensure_loaded?(JsonRpcBackend) ->
    config :logger,
      backends: [JsonRpcBackend]

    config :logger, JsonRpcBackend,
      level: :error,
      format: "$message",
      metadata: []
```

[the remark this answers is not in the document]
jollyjerr: If we do remove this file - it was the only thing ever passing a different io_device to the std_io transport, so I think we can just remove the option to pass a device config as well. I'll push a change so you see what I mean but we can revert if I'm missing something.

apps/server/lib/lexical/server/window.ex
jollyjerr: Ah this is a really helpful comment. I was having trouble understanding what code should live in `remote_control` vs `server`. I don't see another instance of code in `server` subscribing to `RemoteControl.Dispatch` but I think that would be a valid thing to add in this PR if you wish? The ability to invoke logs and messages from remote_control could be needed eventually

scohen: The dependencies in the sub apps go one way. Anyone can use common, proto and protocol are used by server, as is remote_control. `server` can use `remote_control`, but not vice-versa. 

There is a lot of code in `server` that subscribes to certain messages via dispatch, it's just done via `RemoteControl.Api.register_listener`. There are a bunch of things that do that at present. 

I don't think we have to drink the whole ocean with your first PR, we can add the ability of `RemoteControl` to send messages later.

apps/server/lib/lexical/server/json_rpc_backend.ex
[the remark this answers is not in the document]
scohen: I'd like to keep the io device, as it's sometimes useful to use when debugging.

apps/server/lib/lexical/server/window.ex
```diff
@@ -0,0 +1,31 @@
...
+  def show(level, message) do
+    show_message = apply(ShowMessage, level, [message])
+    Transport.write(show_message)
+    message
+  end
+
+  defp format_message(message, opts) do
+    case Keyword.get(opts, :label) do
+      nil -> inspect(message) <> "\n"
+      label -> "#{label}: #{inspect(message, limit: :infinity)}\n"
```
scohen, line 28: Using inspect here will quote strings. Is this intended?
Maybe we require that the message is binary, and let the caller format it.

apps/server/lib/lexical/server/window.ex
```diff
@@ -0,0 +1,31 @@
...
+  alias Lexical.Protocol.Notifications.LogMessage
+  alias Lexical.Protocol.Notifications.ShowMessage
+  alias Lexical.Server.Transport
+
+  @type level :: :error | :warning | :info | :log
+
+  @spec log(level, String.t(), [{:label, String.t()}]) :: String.t()
+  def log(level, message, opts \\ [])
+
+  def log(level, message, opts) when level in [:error, :warning, :info, :log] do
```
scohen, line 11: Thoughts on generating helper functions for each level?

scohen: like...
```elixir
@levels [:error, :warning, :info]

for level <- levels do 
  def unquote(level)(message, opts \\ []) do 
    log(level, message, opts)
   end
end
```

apps/server/lib/lexical/server/window.ex
jollyjerr: Yeah that sounds good 👍🏼  I had just moved this code from the transport layer so there was no reasoning behind the inspect here

apps/server/lib/lexical/server/window.ex
jollyjerr: Love it!
Screenshot 2023-10-21 at 12 22 51 PM

apps/server/lib/lexical/server/window.ex
```diff
@@ -0,0 +1,27 @@
...
+  alias Lexical.Protocol.Notifications.ShowMessage
+  alias Lexical.Server.Transport
+
+  @type level :: :error | :warning | :info | :log
+
+  @spec log(level, String.t()) :: String.t()
+  def log(level, message) when level in [:error, :warning, :info, :log] do
+    log_message = apply(LogMessage, level, [message])
+    Transport.write(log_message)
+    message
```
scohen, line 12: We probably should return the result of Transoort.write, or just ok for both log and show. I think returning the message made more sense when it could be anything rather than just a string

apps/server/lib/lexical/server/window.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Lexical.Server.Window do
+  alias Lexical.Protocol.Notifications.LogMessage
+  alias Lexical.Protocol.Notifications.ShowMessage
+  alias Lexical.Server.Transport
+
+  @type level :: :error | :warning | :info | :log
+
+  @spec log(level, String.t()) :: String.t()
+  def log(level, message) when level in [:error, :warning, :info, :log] do
```
scohen, line 9: We should put a guard ensuring message is binary

scohen (approved):
Fix the last comments and we're good to go.

jollyjerr:
Thank you all for the reviews!

scohen (approved):

PR #421, opened 2023-10-16
Improved memory performance while indexing

scohen:
While indexing a small-ish project like Lexical, memory would spike to over 4 gigs used. This change prevents messages from building up in async tasks and, while it results in slightly slower indexing time (6 seconds vs. 4 seconds), memory only spikes to around 300mb.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/indexer.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/indexer.ex
@@ -60,19 +60,57 @@ defmodule Lexical.RemoteControl.Search.Indexer do
     end
   end
 
-  # Note: I tried doing something very similar using
-  # Task.async_stream, but it was barely faster than just
-  # doing things with Enum (16s vs 20s on my M2 max). This
-  # function takes around 4 seconds to reindex all Lexical's
-  # modules, making it 5x faster than Enum and 4x faster than
-  # Task.async_stream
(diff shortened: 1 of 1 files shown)


apps/remote_control/lib/lexical/remote_control/search/indexer.ex
```diff
@@ -60,19 +60,57 @@ defmodule Lexical.RemoteControl.Search.Indexer do
...
+    |> Enum.shuffle()
+    |> Stream.chunk_while(initial_state, chunk_fn, after_fn)
+    |> Task.async_stream(&Enum.map(&1, processor), timeout: timeout)
+    |> Enum.flat_map(fn
+      {:ok, entry_chunks} -> entry_chunks
+      _ -> []
+    end)
+  end
+
+  defp path_to_sizes(paths) do
```
zachallaun, line 104: This currently silently ignores paths for which `File.stat` doesn't return `{:ok, stat}`. Should we log something?

zachallaun (approved):

apps/remote_control/lib/lexical/remote_control/search/indexer.ex
scohen: i thought about doing that, but this is really specific, to this implementation, which grabs all the changed files in a call literally the line before `async_chunks` is called. If a file is deleted, it was another process that did it, and I don't think logging would make any difference other than being noisy.

PR #422, opened 2023-10-18
Fix invocation with rtx

soundmonster:
Fixes the LSP from crashing when used with a project that also uses `rtx`

Invocation of the LSP when using `rtx` resulted on timeout errors that were ultimately due to the port wrapper script
not invoking the provided launcher binary but instead just evaluating env vars and exiting.


soundmonster:
Thank you! I'll check and will close if already fixed.

PR #431, opened 2023-10-18
Quiescent sync

scohen:
Two commits here, unrelated. Please review separately.


apps/remote_control/lib/lexical/remote_control/search/store.ex
```diff
@@ -28,12 +29,19 @@ defmodule Lexical.RemoteControl.Search.Store do
...
+                     )
+
+  defrecordp :store, state: nil, update_ref: nil
 
   use GenServer
   require Logger
 
   def stop do
-    GenServer.call(__MODULE__, :drop)
+    GenServer.call(__MODULE__, :on_stop)
```
zachallaun, line 44: Would it be better to use the `terminate/2` callback here? According to the docs, `GenServer.stop/1` will already execute the terminate callback, and it also allows us to potentially do this cleanup when the store is killed by other means.

zachallaun (approved):
Couple of comments, but this looks fine to me.

PR #434, opened 2023-10-19
Put module below call signature for hover funs/types

zachallaun:
For especially long module names, the actual function name/signature could be far off to the right. This change puts the module in which the call is defined immediately below the signature, so that the signature itself is always in a consistent location.

### DIFF
--- a/apps/server/lib/lexical/server/provider/handlers/hover.ex
+++ b/apps/server/lib/lexical/server/provider/handlers/hover.ex
@@ -126,7 +126,8 @@ defmodule Lexical.Server.Provider.Handlers.Hover do
 
     header =
       Markdown.code_block("""
-      #{module_name}.#{entry.name}/#{entry.arity}
+      #{entry.name}/#{entry.arity}
+      #{module_name}
 
       #{type_defs(entry)}\
       """)
@@ -145,7 +146,7 @@ defmodule Lexical.Server.Provider.Handlers.Hover do
           ""
         end
 
-      {:ok, "#{macro_prefix}#{module_name}.#{signature}"}
+      {:ok, "#{macro_prefix}#{signature}\n#{module_name}"}
     end
   end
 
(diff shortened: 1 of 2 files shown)


scohen (approved):

scottming:
I think it might be better to keep the last module for excessively long module names because we usually don't care about the specific path of the module. 

However, the last module is typically our context, keep that so we don't need to scan our eyes to the right.

### This PR
CleanShot 2023-10-20 at 15 21 38 2@2x

### Keep last module


CleanShot 2023-10-20 at 15 59 29@2x


CleanShot 2023-10-20 at 16 00 39@2x

 ```elixir
 diff --git a/apps/server/lib/lexical/server/provider/handlers/hover.ex b/apps/server/lib/lexical/server/provider/handlers/hover.ex
index 1b9853d..60efc3d 100644
--- a/apps/server/lib/lexical/server/provider/handlers/hover.ex
+++ b/apps/server/lib/lexical/server/provider/handlers/hover.ex
@@ -146,10 +146,25 @@ defmodule Lexical.Server.Provider.Handlers.Hover do
           ""
         end

-      {:ok, "#{macro_prefix}#{signature}\n#{module_name}"}
+      one_line_header = "#{macro_prefix}#{module_name}.#{signature}"
+
+      formatted_header =
+        "#{macro_prefix}#{last_module_name(module_name)}.#{signature}\n#{module_name}"
+
+      if String.length(one_line_header) >= 50 do
+        {:ok, formatted_header}
+      else
+        {:ok, one_line_header}
+      end
     end
   end

+  defp last_module_name(module_name) do
+    module_name
+    |> String.split(".")
+    |> List.last()
+  end
+
   defp type_defs(%Docs.Entry{metadata: %{opaque: true}} = entry) do
     Enum.map_join(entry.defs, "\n", fn def ->
       def
```

scottming:
I have another idea, which is to gray out the preceding modules. This effect might work better.

The effect would be like this:

CleanShot 2023-10-20 at 16 27 02@2x
CleanShot 2023-10-20 at 16 29 51@2x


By doing this, we could even keep only the last module in all of the code blocks.

scottming (approved):
However, it may not look as good in VS Code. Up to you, I don't have a strong opinion.

CleanShot 2023-10-20 at 21 51 11@2x

scottming (approved):

PR #435, opened 2023-10-19
Removed optional synchronization for ETS backend

scohen:
Since we made the store buffer updates, it doesn't make sense to also have the backkends do the same.


zachallaun (approved):
LGTM

PR #436, opened 2023-10-20
Testing CI (do not merge)

zachallaun:
Trying to diagnose the `:enoent` error we're getting [here](https://github.com/lexical-lsp/lexical/actions/runs/6576602438/job/17880567067).

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -161,14 +161,14 @@ jobs:
       # and running the workflow steps.
       matrix:
         include:
-          - elixir: "1.15.3-otp-26"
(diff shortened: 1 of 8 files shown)


zachallaun:
Current theory:

Since this error seems to be coming from the Elixir parallel compiler trying to read in a non-existing file, I believe this was a bug in Elixir 1.15.3 that was resolved [in 1.15.4](https://hexdocs.pm/elixir/1.15/changelog.html#v1-15-4-2023-07-18):

- [mix compile] Track removed files per local dependency (this addresses a bug where files depending on modules from path dependencies always recompiled)

PR #438, opened 2023-10-20
Bugfix: Document Store needs to be globally unique

### LINKED ISSUE #425: Indexing: Document Store needs a new registration mechanism
Prior to the ETS index, each editor had a server node and a project node kept in distribution. We used `:global` to register the document store in the distribution, and this worked well, since both the server node and the project node could see it. 

Things got a little more complicated with the ets store. Since only one thing can write to an index at a time, we do a leader election to see who gets ownership of the index directory. This means we need to find the leader node for our project, and connect to it, so we can use its index. The knock-on effect of this is that global registration for the project node doesn't work any more, because now the distribution encompasses all of the nodes for all editors due to distribution requiring all nodes to be connected to all other nodes.

What's happening now is that name registration fails for `Document.Store`, and the `Document.Store` process get killed, and chaos ensues. 

What needs to happen is that `Document.Store` is registered for only the set of nodes that comprises an editor's project. This might be as easy as including a project's `entropy` in the store's registered name.

scohen:
Due to all nodes being connected via distribution after the ETS store is enabled, we can no longer register a document store globally with a non-unique name because Document.Store needs to be associated with an editor.
This change creates some entropy when the server starts to make the document store's name unique, and passes that entropy to the project node.

Fixes #425

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/bootstrap.ex
+++ b/apps/remote_control/lib/lexical/remote_control/bootstrap.ex
@@ -10,7 +10,8 @@ defmodule Lexical.RemoteControl.Bootstrap do
   alias Lexical.RemoteControl
   require Logger
 
-  def init(%Project{} = project, remote_control_config) do
+  def init(%Project{} = project, document_store_entropy, remote_control_config) do
+    Lexical.Document.Store.set_entropy(document_store_entropy)
     maybe_append_hex_path()
     Application.put_all_env(remote_control: remote_control_config)
 
--- a/apps/remote_control/lib/lexical/remote_control/project_node.ex
+++ b/apps/remote_control/lib/lexical/remote_control/project_node.ex
@@ -97,20 +97,18 @@ defmodule Lexical.RemoteControl.ProjectNode do
     end
   end
 
+  alias Lexical.Document
   alias Lexical.RemoteControl.ProjectNodeSupervisor
   use GenServer
 
   def start(project, paths) do
     node_name = Project.node_name(project)
     remote_control_config = Application.get_all_env(:remote_control)
+    bootstrap_args = [project, Document.Store.entropy(), remote_control_config]
 
     with {:ok, node_pid} <- ProjectNodeSupervisor.start_project_node(project),
          :ok <- start_node(project, paths),
-         :ok <-
-           :rpc.call(node_name, RemoteControl.Bootstrap, :init, [
-             project,
-             remote_control_config
-           ]) do
+         :ok <- :rpc.call(node_name, RemoteControl.Bootstrap, :init, bootstrap_args) do
       {:ok, node_pid}
     end
   end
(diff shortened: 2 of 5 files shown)


scottming:
I tried this branch and encountered the "not_leader" error. I think we need to further address these areas.


```elixir
2023-10-21T14:54:01.849768+08:00 error: GenServer LXical.RemoteControl.Search.Store terminating, ** (CaseClauseError) no case clause matching: :error, (lx_remote_control 0.3.0) lib/lexical/remote_control/search/store/state.ex:59: LXical.RemoteControl.Search.Store.State.async_load_complete/2, (lx_remote_control 0.3.0) lib/lexical/remote_control/search/store.ex:117: LXical.RemoteControl.Search.Store.handle_info/2, (stdlib 4.3.1.2) gen_server.erl:1123: :gen_server.try_dispatch/4, (stdlib 4.3.1.2) gen_server.erl:1200: :gen_server.handle_msg/6, (stdlib 4.3.1.2) proc_lib.erl:240: :proc_lib.init_p_do_apply/3, Last message: {#Reference<0.1891001025.2318991362.198912>, :error}, State: {nil, %LXical.RemoteControl.Search.Store.State{project: %LXical.Project{root_uri: "file:///Users/scottming/Code/dummy", mix_exs_uri: "file:///Users/scottming/Code/dummy/mix.exs", mix_project?: true, mix_env: nil, mix_target: nil, env_variables: %{}, project_module: Dummy.MixProject, entropy: 45455, project_config: [app: :dummy, version: "0.1.0", elixir: "~> 1.15", start_permanent: false, deps: []]}, backend: LXical.RemoteControl.Search.Store.Backends.Ets, create_index: &LXical.RemoteControl.Search.Indexer.create_index/1, update_index: &LXical.RemoteControl.Search.Indexer.update_index/2, loaded?: false, fuzzy: %LXical.RemoteControl.Search.Fuzzy{subject_to_values: %{}, grouping_key_to_values: %{}, preprocessed_subjects: %{}, mapper: #Function<8.98552950/1 in LXical.RemoteControl.Search.Fuzzy.from_entries/1>, subject_converter: #Function<9.98552950/1 in LXical.RemoteControl.Search.Fuzzy.stringify>}, async_load_ref: #Reference<0.1891001025.2318991362.198912>, update_buffer: %{}}}
2023-10-21T14:54:01.864380+08:00 info: not leader
2023-10-21T14:54:01.864617+08:00 error: Could not initialize index due to :error
```
https://github.com/lexical-lsp/lexical/blob/4ef324143cecce608a162d58ff12af052cf9fbd1/apps/remote_control/lib/lexical/remote_control/search/store/backends/ets/state.ex#L43

https://github.com/lexical-lsp/lexical/blob/4ef324143cecce608a162d58ff12af052cf9fbd1/apps/remote_control/lib/lexical/remote_control/search/store/state.ex#L58

scottming:
Change this function

https://github.com/lexical-lsp/lexical/blob/4ef324143cecce608a162d58ff12af052cf9fbd1/apps/remote_control/lib/lexical/remote_control/search/store/backends/ets/state.ex#L43

to 

```diff
-  def prepare(%__MODULE__{leader?: false}) do
-    {:error, :not_leader}
+  def prepare(%__MODULE__{leader?: false} = state) do
+    {{:error, :not_leader}, state}
   end
   ```

then everything works fine. @scohen

scottming (approved):
It works very well.

Issue #263, opened 2023-07-12
Function ordering needs work


scohen:
I think function ordering needs some help; it should be sorted with lower arity functions first, currently, higher arity functions come first.

Nezteb:
A few qestions. I found some relevant tests:

- https://github.com/lexical-lsp/lexical/blob/185152d56f9a357bd98e704de3bd2c752f3db0ee/apps/server/test/lexical/server/code_intelligence/completion/translations/function_test.exs#L236
- https://github.com/lexical-lsp/lexical/blob/185152d56f9a357bd98e704de3bd2c752f3db0ee/apps/server/test/lexical/server/code_intelligence/completion_test.exs#L162
- https://github.com/lexical-lsp/lexical/blob/185152d56f9a357bd98e704de3bd2c752f3db0ee/apps/server/test/lexical/server/code_intelligence/completion/translations/macro_test.exs#L593

However, if I add `IO.inspect` to those tests to view the actual completions, the lower arity functions are already first in the list. Am I looking in the wrong place?

After a few glances at implementation code, my initial idea was to add a new `arity_boost` variable here: https://github.com/lexical-lsp/lexical/blob/1bd442e72fb82af89ddb9c63279d3dda084ff408/apps/server/lib/lexical/server/code_intelligence/completion/env.ex#L318, but I'm not sure if that's the correct place for a fix.

Thoughts?

scohen:
@Nezteb I think I'd keep the local / global things, but maybe change the boost to have three characters rather than one. 
i.e.
```
000_000_#{item.label}
```
You'll also need to `String.pad_leading` the boosts now.

PR #442, opened 2023-10-23
Consider arity in function ordering

yerguden:
My attempt at the "function ordering by arity" issue. I am not sure if passing a list to boost was the sensible approach here, but I figured any pr would make the discussion easier 😄 

related to #263


scohen (commented):
You're right about a pr being a good jumping off point for discussion, thanks for putting one together. 
Looking at the code, I think an overall simpler approach would to be for builder to allow sort text to already be defined and then prepending boost to it. Then all you have to do is set the sort text for functions to be `<name>:<arity>` where arity is a three digit number, like 001..you wouldn't need to mess with boost at all then.

apps/server/lib/lexical/server/code_intelligence/completion/builder.ex
```diff
@@ -80,15 +80,23 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Builder do
...
 
-  def boost(%Completion.Item{} = item, local_boost, global_boost)
-      when local_boost in 0..9 and global_boost in 0..9 do
-    global_boost = Integer.to_string(9 - global_boost)
-    local_boost = Integer.to_string(9 - local_boost)
+  def boost(%Completion.Item{} = item, local_boost , global_boost)
+    when is_list(local_boost) and is_list(global_boost) and length(local_boost) <= 3 and length(global_boost) <= 3 do
+    sort_text = "0#{boost_to_sort_string(global_boost)}#{boost_to_sort_string(local_boost)}_#{item.label}"
 
-    sort_text = "0#{global_boost}#{local_boost}_#{item.label}"
```
scohen, line 88: This line would be 
```elixir
sort_text = "0#{global_boost}#{local_boost}_#{item.sort_text || item.label}"

File: apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex [hunk 2]
```diff
@@ -99,11 +99,12 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Callable do
 
   @default_functions ["module_info", "behaviour_info"]
 
-  defp maybe_boost(item, %_{name: name}, default_boost \\ 5) do
+  defp maybe_boost(item, %_{name: name, arity: arity}, default_boost \\ 5) do
     if String.starts_with?(name, "__") or name in @default_functions do
       item
     else
-      Builder.boost(item, default_boost)
+      arity_boost = if arity < 9, do: 9 - arity, else: 9
```
scohen, line 106: This would just set the sort text

yerguden:
Thanks for the guidance, it did end up becoming a lot simpler 😄

scohen (commented):
One more change, then we're good.

apps/server/lib/lexical/server/code_intelligence/completion/builder.ex
```diff
@@ -85,7 +85,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Builder do
     global_boost = Integer.to_string(9 - global_boost)
     local_boost = Integer.to_string(9 - local_boost)
 
-    sort_text = "0#{global_boost}#{local_boost}_#{item.label}"
+    sort_text = "0#{global_boost}#{local_boost}_#{item.sort_text || item.label}"
```
scohen, line 88: heh, super, super nit (and this is my fault)
```suggestion
    sort_text = "0#{global_boost}#{local_boost}_#{fallback(item.sort_text, item.label)}"
```

scohen:
FWIW, when this lands, i intend to release it in 0.4.1

scohen:
@yerguden were you going to make the last two changes, or do you want me to merge it and make them?

yerguden:
Hi, I was planning on making them today. Sorry for the delay; I had a busier-than-usual week.

yerguden:
I added some changes, though I am unsure if they are in the right direction. I'll jot down my thoughts on the relevant bits.

apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
```diff
@@ -101,7 +104,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Callable do
 
   defp maybe_boost(item, %_{name: name}, default_boost \\ 5) do
     if String.starts_with?(name, "__") or name in @default_functions do
-      item
+      Builder.boost(item, 0)
```
yerguden, line 107: One of the changes I am not proud of is having to do this here; I needed to add the default boost back here because overriding sort_text essentially stripped the one that came from Builder.snippet. 

I would love to hear any suggestions you might have

apps/server/test/lexical/server/code_intelligence/completion_test.exs
```diff
@@ -102,11 +102,24 @@ defmodule Lexical.Server.CodeIntelligence.CompletionTest do
...
+        metadata: %{},
+        arity: 0
+      },
+      %Candidate.Macro{
+        name: "my_macro",
+        origin: full_name,
+        argument_names: [],
+        metadata: %{},
+        arity: 0
+      },
```
yerguden, line 122: I added an arity of 0 to the mocks in this part (instead of adjusting the sort_text function to handle the arity value of nil) because I thought that was the realistic case. But I'm not super on that 😕 .

apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
scohen: the `set_sort_text` should call `boost(0)`

yerguden: Thanks, that makes sense 👍

yerguden: My bad; I remembered my issue with that approach. If I do a boost(0) there, later maybe_boost function will prepend the already boosted sort_text, and it ends up with something like "094_099_function_name:arity" because boost also updates the sort_text of the item.

yerguden:
Again, thanks for the excellent advice :) I made the changes we discussed.

apps/server/lib/lexical/server/code_intelligence/completion/builder.ex
```diff
@@ -84,8 +84,9 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Builder do
       when local_boost in 0..9 and global_boost in 0..9 do
     global_boost = Integer.to_string(9 - global_boost)
     local_boost = Integer.to_string(9 - local_boost)
+    stripped_sort_text = strip_boost(item.sort_text)
```
scohen, line 87: wouldn't this be more clear?
```suggestion
    stripped_sort_text =
     item.label
     |> fallback(item.sort_text)
     |> strip_sort_text()
```

apps/server/test/lexical/server/code_intelligence/completion_test.exs
scohen: seems reasonable

apps/server/lib/lexical/server/code_intelligence/completion/builder.ex
yerguden: Looks good, updated the part

scohen (approved):

PR #449, opened 2023-10-25
Omit name and body during updates

scohen:
I re-ran the release to regenerate the artifacts, and it clobbered my release notes. I think this will prevent that from happening in the future.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,3 +53,5 @@ jobs:
           generateReleaseNotes: true
           allowUpdates: true
           artifactErrorsFailBuild: true
+          omitBodyDuringUpdate: true
+          omitNameDuringUpdate: true


.github/workflows/release.yml
```diff
@@ -53,3 +53,5 @@ jobs:
           generateReleaseNotes: true
```
Blond11516, line 53:
```suggestion
```
I think removing that should be enough? Though I don't see the harm in using `omitBodyDuringUpdate` as well to be extra safe.

.github/workflows/release.yml
```diff
@@ -53,3 +53,5 @@ jobs:
...
           allowUpdates: true
           artifactErrorsFailBuild: true
+          omitBodyDuringUpdate: true
+          omitNameDuringUpdate: true
```
Blond11516, line 57: That's an interesting option. How is it even supposed to know which release to update if the name doesn't match? I'm not really expecting an answer here, just thinking outloud.

Blond11516 (approved):
👍

.github/workflows/release.yml
scohen: I think it uses the tag name to figure out which release to use.

.github/workflows/release.yml
scohen: Yeah, I saw this, it's probably the safer option, I'll pick that instead.

PR #453, opened 2023-10-26
Code Actions refactor

### LINKED ISSUE #446: Usability: Improve Code Actions
I think code actions' usability can be improved a lot. Right now, there's a router in `server` that delegates to calls via `RemoteControl.Api`. I think we can push the router into `remote_control`, pass in the diagnostic information and have the router emit code actions for the diagnostic. This means people would only need to do the work inside `remote_control`.

scohen:
After observing a PR for code actions, it seemed there was much to much busywork in creating them. This PR seeks to remedy that. Now, in order to create a code action, you just need to create a new handler in the `remote_control` app.

The server provider now encompasses all code actions, and handles marshalling from and to the protocol. AST transformations now reside in remote_control, which is where they belong.

Fixes #446


scottming:
I acknowledge that it takes 2.4 seconds for amortization when used. However, when we let Elixir automatically load a small number of modules during usage, users won't perceive this impact. For example, the module `Jason.Encode` only takes 0.003 seconds.

So, I am more inclined to do this at runtime.

**Manually loading**

```elixir
iex(1)> Code.loaded?(Jason.Encode)
false
iex(2)> :timer.tc(fn -> Code.ensure_loaded(Jason.Encode) end)
{3766, {:module, Jason.Encode}}
iex(3)>
```


**Automatically Landing**

```elixir
iex(1)> Code.loaded?(Jason.Encode)
false
iex(2)> Jason.encode(%{})
{:ok, "{}"}
iex(3)> Code.loaded?(Jason.Encode)
true
```

apps/server/lib/lexical/server/provider/handlers/code_action.ex
```diff
@@ -1,18 +1,42 @@
 defmodule Lexical.Server.Provider.Handlers.CodeAction do
   alias Lexical.Protocol.Requests
   alias Lexical.Protocol.Responses
-  alias Lexical.Server.Provider.CodeAction.ReplaceWithUnderscore
-  alias Lexical.Server.Provider.CodeAction.ReplaceWithUnderscore
+  alias Lexical.Protocol.Types
+  alias Lexical.Protocol.Types.CodeAction, as: CodeActionResult
+  alias Lexical.Protocol.Types.Workspace
```
zachallaun, line 6: I've thought about this a few times, and I'm wondering if we should always start using the `Types` prefix when dealing with protocol types. You're already doing it below with `%Types.Diagnostic{}`, and I think just using `Types.CodeAction` is more clear than aliasing as `CodeActionResult`.

```suggestion
```

apps/remote_control/lib/lexical/remote_control/code_action.ex
```diff
@@ -0,0 +1,55 @@
...
+  @type t :: %__MODULE__{title: String.t(), kind: code_action_kind, changes: Changes.t()}
+
+  @handlers [Handlers.ReplaceWithUnderscore]
+
+  @spec new(Lexical.uri(), String.t(), code_action_kind(), Changes.t()) :: t()
+  def new(uri, title, kind, changes) do
+    %__MODULE__{uri: uri, title: title, changes: changes, kind: kind}
+  end
+
+  @spec for_range(Document.t(), Range.t(), [Diagnostic.t()], [code_action_kind | :all]) :: [t()]
```
zachallaun, line 30:
```suggestion
  @spec for_range(Document.t(), Range.t(), [Diagnostic.t()], [code_action_kind] | :all) :: [t()]
```

apps/remote_control/lib/lexical/remote_control/code_action.ex
```diff
@@ -0,0 +1,55 @@
...
+          | :quick_fix
+          | :refactor
+          | :refactor_extract
+          | :refactor_inline
+          | :refactor_rewrite
+          | :source
+          | :source_organize_imports
+          | :source_fix_all
+
+  @type t :: %__MODULE__{title: String.t(), kind: code_action_kind, changes: Changes.t()}
```
zachallaun, line 21: Missing `uri`:

```suggestion
  @type t :: %__MODULE__{
          title: String.t(),
          kind: code_action_kind,
          changes: Changes.t(),
          uri: Lexical.uri()
        }
```

apps/remote_control/lib/lexical/remote_control/code_action.ex
```diff
@@ -0,0 +1,55 @@
...
+  def for_range(%Document{} = doc, %Range{} = range, diagnostics, kinds) do
+    results =
+      Enum.reduce(@handlers, [], fn handler, acc ->
+        if applies?(kinds, handler) do
+          actions = handler.actions(doc, range, diagnostics)
+          actions ++ acc
+        else
+          acc
+        end
+      end)
```
zachallaun, line 40:
```suggestion
      Enum.flat_map(@handlers, fn handler ->
        if applies?(kinds, handler) do
          handler.actions(doc, range, diagnostics)
        else
          []
        end
      end)
```

apps/remote_control/lib/lexical/remote_control/code_action.ex
```diff
@@ -0,0 +1,55 @@
...
+
+  defp applies?(:all, _handler_module) do
+    true
+  end
+
+  defp applies?(kinds, handler_module) do
+    requested_kinds = MapSet.new(kinds)
+    handler_kinds = MapSet.new(handler_module.kinds())
+    applicable_kinds = MapSet.intersection(requested_kinds, handler_kinds)
+    not Enum.empty?(applicable_kinds)
```
zachallaun, line 53: Slight simplication:

```suggestion
    not MapSet.disjoint?(requested_kinds, handler_kinds)
```

Code golf isn't at all the goal here, but this goal function could be a one-liner just using lists:

```elixir
(kinds -- handler_module.kinds()) != kinds
```

apps/remote_control/lib/lexical/remote_control/code_action/handler.ex
```diff
@@ -0,0 +1,10 @@
+defmodule Lexical.RemoteControl.CodeAction.Handler do
+  alias Lexical.Document
+  alias Lexical.Document.Changes
+  alias Lexical.Document.Range
+  alias Lexical.RemoteControl.CodeAction
+  alias Lexical.RemoteControl.CodeAction.Diagnostic
+
+  @callback actions(Document.t(), Range.t(), [Diagnostic.t()]) :: [Changes.t()]
```
zachallaun, line 8:
```suggestion
  @callback actions(Document.t(), Range.t(), [Diagnostic.t()]) :: [CodeAction.t()]
```

zachallaun (approved):
Looks like a good direction to me. I think there will be at least one more change to how these work when plugins are able to provide code actions -- it might be that we want all the "built-in" handlers to just use the plugin system to register themselves. Plugin discovery happens in the remote control app though, so it's already better for these to be here.

scohen:
>  it might be that we want all the "built-in" handlers to just use the plugin system to register themselves. 

100% agree. We need to eat our own dogfood.

apps/remote_control/lib/lexical/remote_control/code_action.ex
scohen: good catch!

PR #455, opened 2023-10-28
Optimise the manual loading of dependent apps and modules.

scottming:
I have always felt that the startup of 'lexical' is very slow. After some investigation, I found that this is mainly because we manually load all dependent modules, which takes about `2.4` seconds.

After removing this line: `Enum.each(modules, &Code.ensure_loaded!/1)`, I found that 'lexical' works perfectly fine. Additionally, when we start the `:server` app,it will automatically load and start the dependent apps, so we don't need to manually load the dependent apps either.

### before

https://github.com/lexical-lsp/lexical/assets/12830256/1abf21e3-5cb7-4ad4-b4a8-f39e38b9da99

### After

https://github.com/lexical-lsp/lexical/assets/12830256/c9029063-44c6-4434-8230-884680d22f65

### DIFF
--- a/apps/server/lib/lexical/server/boot.ex
+++ b/apps/server/lib/lexical/server/boot.ex
@@ -4,6 +4,7 @@ defmodule Lexical.Server.Boot do
 
   Packaging will ensure that config.exs and runtime.exs will be visible to the `:code` module
   """
+  alias Future.Code
   alias Lexical.VM.Versions
   require Logger
 
@@ -78,8 +79,8 @@ defmodule Lexical.Server.Boot do
   defp load_app_modules(app_name) do
     Application.ensure_loaded(app_name)
 
-    with {:ok, modules} <- :application.get_key(app_name, :modules) do
-      Enum.each(modules, &Code.ensure_loaded!/1)
+    with {:ok, modules} <- Application.spec(app_name, :modules) do
+      Code.ensure_all_loaded!(modules)
     end
   end
 
--- a/apps/server/lib/lexical/server/boot.ex
+++ b/apps/server/lib/lexical/server/boot.ex
@@ -78,10 +78,8 @@ defmodule Lexical.Server.Boot do
 
   defp load_app_modules(app_name) do
     Application.ensure_loaded(app_name)
-
-    with {:ok, modules} <- Application.spec(app_name, :modules) do
-      Code.ensure_all_loaded!(modules)
-    end
+    modules = Application.spec(app_name, :modules)
+    Code.ensure_all_loaded!(modules)
   end
 
   @allowed_elixir %{
--- a/apps/server/lib/lexical/server/boot.ex
+++ b/apps/server/lib/lexical/server/boot.ex
@@ -22,8 +22,6 @@ defmodule Lexical.Server.Boot do
     load_config()
     Application.ensure_all_started(:logger)
 
-    Enum.each(@dep_apps, &load_app_modules/1)
-
     case detect_errors() do
       [] ->
         :ok
@@ -35,6 +33,7 @@ defmodule Lexical.Server.Boot do
     end
 
     Application.ensure_all_started(:server)
+    Enum.each(@dep_apps, &load_app_modules/1)
   end
 
   @doc false
--- a/apps/server/lib/lexical/server/boot.ex
+++ b/apps/server/lib/lexical/server/boot.ex
@@ -22,6 +22,8 @@ defmodule Lexical.Server.Boot do
     load_config()
     Application.ensure_all_started(:logger)
 
+    Enum.each(@dep_apps, &load_app_modules/1)
+
     case detect_errors() do
       [] ->
         :ok
@@ -33,7 +35,6 @@ defmodule Lexical.Server.Boot do
     end
 
     Application.ensure_all_started(:server)
-    Enum.each(@dep_apps, &load_app_modules/1)
   end
 
   @doc false
--- a/apps/proto/lib/lexical/proto/field.ex
+++ b/apps/proto/lib/lexical/proto/field.ex
@@ -274,7 +274,7 @@ defmodule Lexical.Proto.Field do
   end
 
   def encode(module, field_value) when is_atom(module) do
-    if Code.ensure_loaded?(module) and function_exported?(module, :encode, 1) do
+    if function_exported?(module, :encode, 1) do
       module.encode(field_value)
     else
       {:ok, field_value}
--- a/apps/proto/lib/lexical/proto/field.ex
+++ b/apps/proto/lib/lexical/proto/field.ex
@@ -274,7 +274,7 @@ defmodule Lexical.Proto.Field do
   end
 
   def encode(module, field_value) when is_atom(module) do
-    if function_exported?(module, :encode, 1) do
+    if Code.ensure_loaded?(module) and function_exported?(module, :encode, 1) do
       module.encode(field_value)
     else
       {:ok, field_value}
--- a/apps/protocol/test/lexical/proto_test.exs
+++ b/apps/protocol/test/lexical/proto_test.exs
@@ -1,11 +1,9 @@
 defmodule Lexical.ProtoTest do
-  alias Future.Code, as: Code
   alias Lexical.Document
   alias Lexical.Proto
   alias Lexical.Proto.Convert
   alias Lexical.Proto.LspTypes
   alias Lexical.Protocol.Types
-  alias Lexical.Test.Protocol.Fixtures
   alias Lexical.Test.Protocol.Fixtures.LspProtocol
 
   import LspProtocol
@@ -284,16 +282,6 @@ defmodule Lexical.ProtoTest do
       assert ConstantTest.good() == 1
       assert ConstantTest.bad() == 2
       assert ConstantTest.ugly() == 3
-    end
-
-    test "constants should be encoded even if the module isn't loaded" do
-      proto = Fixtures.UsesConstantsForUnloading.new(name: "Clint", state: :good)
-
-      :code.purge(Fixtures.ConstantsForUnloading)
-      :code.delete(Fixtures.ConstantsForUnloading)
-
-      refute Code.loaded?(Fixtures.ConstantsForUnloading)
-      assert Jason.encode(proto) == {:ok, ~s|{"name":"Clint","state":1}|}
     end
 
     test "constants should parse" do
--- a/apps/protocol/test/lexical/proto_test.exs
+++ b/apps/protocol/test/lexical/proto_test.exs
@@ -1,9 +1,11 @@
 defmodule Lexical.ProtoTest do
+  alias Future.Code, as: Code
   alias Lexical.Document
   alias Lexical.Proto
   alias Lexical.Proto.Convert
   alias Lexical.Proto.LspTypes
   alias Lexical.Protocol.Types
+  alias Lexical.Test.Protocol.Fixtures
   alias Lexical.Test.Protocol.Fixtures.LspProtocol
 
   import LspProtocol
@@ -282,6 +284,16 @@ defmodule Lexical.ProtoTest do
       assert ConstantTest.good() == 1
       assert ConstantTest.bad() == 2
       assert ConstantTest.ugly() == 3
+    end
+
+    test "constants should be encoded even if the module isn't loaded" do
+      proto = Fixtures.UsesConstantsForUnloading.new(name: "Clint", state: :good)
+
+      :code.purge(Fixtures.ConstantsForUnloading)
+      :code.delete(Fixtures.ConstantsForUnloading)
+
+      refute Code.loaded?(Fixtures.ConstantsForUnloading)
+      assert Jason.encode(proto) == {:ok, ~s|{"name":"Clint","state":1}|}
     end
 
     test "constants should parse" do
--- a/apps/protocol/test/support/lexical/test/protocol/fixtures/constants_for_unloading.ex
+++ b/apps/protocol/test/support/lexical/test/protocol/fixtures/constants_for_unloading.ex
@@ -1,0 +1,6 @@
+defmodule Lexical.Test.Protocol.Fixtures.ConstantsForUnloading do
+  alias Lexical.Proto
+  use Proto
+
+  defenum good: 1, bad: 2, ugly: 3
+end
--- a/apps/protocol/test/support/lexical/test/protocol/fixtures/constants_for_unloading.ex
+++ b/apps/protocol/test/support/lexical/test/protocol/fixtures/constants_for_unloading.ex
@@ -1,6 +1,0 @@
-defmodule Lexical.Test.Protocol.Fixtures.ConstantsForUnloading do
-  alias Lexical.Proto
-  use Proto
-
-  defenum good: 1, bad: 2, ugly: 3
-end
--- a/apps/protocol/test/support/lexical/test/protocol/fixtures/uses_constants_for_unloading.ex
+++ b/apps/protocol/test/support/lexical/test/protocol/fixtures/uses_constants_for_unloading.ex
@@ -1,0 +1,7 @@
+defmodule Lexical.Test.Protocol.Fixtures.UsesConstantsForUnloading do
+  alias Lexical.Proto
+  alias Lexical.Test.Protocol.Fixtures.ConstantsForUnloading
+  use Proto
+
+  deftype name: string(), state: ConstantsForUnloading
+end
--- a/apps/protocol/test/support/lexical/test/protocol/fixtures/uses_constants_for_unloading.ex
+++ b/apps/protocol/test/support/lexical/test/protocol/fixtures/uses_constants_for_unloading.ex
@@ -1,7 +1,0 @@
-defmodule Lexical.Test.Protocol.Fixtures.UsesConstantsForUnloading do
-  alias Lexical.Proto
-  alias Lexical.Test.Protocol.Fixtures.ConstantsForUnloading
-  use Proto
-
-  deftype name: string(), state: ConstantsForUnloading
-end


zachallaun:
I'll have to wait for @scohen for a more definitive answer, but I thought this is there for struct discovery, which happens in the compilation tracer. Or is this redundant and all that compilation happens anyways?

scohen:
I’m pretty sure that it’s there so elixir sense can see the modules. You can’t see the modules unless they’re loaded.

scottming:
@zachallaun 

but I thought this is there for struct discovery, which happens in the compilation tracer. 

yes, happens in the compilation tracer.

> Or is this redundant and all that compilation happens anyways?

I think so.

scottming:
@scohen 

> I’m pretty sure that it’s there so elixir sense can see the modules. 

All of our Elixirsense calls occur in Project node, so it doesn't make sense to load the lx_xx module in the server node.

CleanShot 2023-10-29 at 09 33 54@2x

and I don't quite understand why ElixirSense needs to know about these modules that have already been namespace.

> You can’t see the modules unless they’re loaded.

It seems that it doesn't affect the discovery of modules, and auto-completion works well in this branch.

CleanShot 2023-10-29 at 09 32 42@2x

scohen:
@scottming there's definitely a reason that this is here, I didn't just plop that code in there for fun.

scohen:
This might have been carried over from the old packaging, I believe releases are started in a mode where all modules need to be explicitly loaded at app start, while the current mode doesn't need that. 
The problem is, any speed boost you perceive will be amortized over the life of the VM, the modules will still need to be loaded at some time, so that 2.4 seconds will happen while the user is editing code. It's a judgement call to say that doing this at startup is better or worse than doing this at runtime. I'm on the fence.

scottming:
I acknowledge that the loading time will be amortized when used. However, when we let Elixir automatically load a small number of modules during usage, users won't perceive this impact. For example, the module `Jason.Encode` only takes 0.003 seconds.

So, I am more inclined to do this at runtime.

**Manually loading**

```elixir
iex(1)> Code.loaded?(Jason.Encode)
false
iex(2)> :timer.tc(fn -> Code.ensure_loaded(Jason.Encode) end)
{3766, {:module, Jason.Encode}}
iex(3)>
```


**Automatically Landing**

```elixir
iex(1)> Code.loaded?(Jason.Encode)
false
iex(2)> Jason.encode(%{})
{:ok, "{}"}
iex(3)> Code.loaded?(Jason.Encode)
true
```

scottming:
Okay, today when I was using the commit that no longer performs loading, I did encounter some strange behaviors with autocompletion. For example, in VS Code, the autocompletion was not visible, and in Neovim, snippets were not being autocompleted.

image

```elixir
10:54:29.581 [info] Local completions are [%LXical.RemoteControl.Completion.Candidate.Module{full_name: "Dummy.Application", metadata: %{}, name: "Application", summary: ""}, %LXical.RemoteControl.Completion.Candidate.Module{full_name: "Dummy.MixProject", metadata: %{}, name: "MixProject", summary: ""}, %LXical.RemoteControl.Completion.Candidate.Function{argument_names: ["atom"], arity: 1, name: "__info__", origin: "Dummy", type: :function, visibility: :public, spec: "@spec __info__(:attributes) :: keyword()\n@spec __info__(:compile) :: [term()]\n@spec __info__(:functions) :: [{atom, non_neg_integer}]\n@spec __info__(:macros) :: [{atom, non_neg_integer}]\n@spec __info__(:md5) :: binary()\n@spec __info__(:module) :: module()", metadata: %{builtin: true}}, %LXical.RemoteControl.Completion.Candidate.Function{argument_names: [], arity: 0, name: "hello", origin: "Dummy", type: :function, visibility: :public, spec: "", metadata: %{}}, %LXical.RemoteControl.Completion.Candidate.Function{argument_names: [], arity: 0, name: "module_info", origin: "Dummy", type: :function, visibility: :public, spec: "@spec module_info :: [{:module | :attributes | :compile | :exports | :md5 | :native, term}]", metadata: %{builtin: true}}, %LXical.RemoteControl.Completion.Candidate.Function{argument_names: ["key"], arity: 1, name: "module_info", origin: "Dummy", type: :function, visibility: :public, spec: "@spec module_info(:module) :: atom\n@spec module_info(:attributes | :compile) :: [{atom, term}]\n@spec module_info(:md5) :: binary\n@spec module_info(:exports | :functions | :nifs) :: [{atom, non_neg_integer}]\n@spec module_info(:native) :: boolean", metadata: %{builtin: true}}, %LXical.RemoteControl.Completion.Candidate.Function{argument_names: ["s"], arity: 1, name: "print_hello", origin: "Dummy", type: :function, visibility: :public, spec: "", metadata: %{}}]
10:54:29.607 [info] Emitting completions: [#Protocol.Types.Completion.Item<[detail: "Dummy.Application", kind: :module, label: "Application", sort_text: "079_Application", text_edit: %LXical.Document.Changes{document: #LXical.Document<path: "/Users/scottming/Code/dummy/test/dummy_test.exs", version: 16, dirty?: true, lines: %Lines<"defmodule DummyTest do..."(9 lines)>, ...>, edits: %LXical.Document.Edit{text: "Application", range: %LXical.Document.Range{start: %LXical.Document.Position{line: 6, character: 11, valid?: true, context_line: {:line, "    Dummy.", "\n", 6, true}, document_line_count: 9, starting_index: 1}, end: %LXical.Document.Position{line: 6, character: 11, valid?: true, context_line: {:line, "    Dummy.", "\n", 6, true}, document_line_count: 9, starting_index: 1}}}}]>, #Protocol.Types.Completion.Item<[detail: "Dummy.MixProject", kind: :module, label: "MixProject", sort_text: "079_MixProject", text_edit: %LXical.Document.Changes{document: #LXical.Document<path: "/Users/scottming/Code/dummy/test/dummy_test.exs", version: 16, dirty?: true, lines: %Lines<"defmodule DummyTest do..."(9 lines)>, ...>, edits: %LXical.Document.Edit{text: "MixProject", range: %LXical.Document.Range{start: %LXical.Document.Position{line: 6, character: 11, valid?: true, context_line: {:line, "    Dummy.", "\n", 6, true}, document_line_count: 9, starting_index: 1}, end: %LXical.Document.Position{line: 6, character: 11, valid?: true, context_line: {:line, "    Dummy.", "\n", 6, true}, document_line_count: 9, starting_index: 1}}}}]>, #Protocol.Types.Completion.Item<[insert_text_format: :snippet, kind: :function, label: "__info__(atom)", sort_text: "099___info__(atom)", text_edit: %LXical.Document.Changes{document: #LXical.Document<path: "/Users/scottming/Code/dummy/test/dummy_test.exs", version: 16, dirty?: true, lines: %Lines<"defmodule DummyTest do..."(9 lines)>, ...>, edits: %LXical.Document.Edit{text: "__info__(${1:atom})", range: %LXical.Document.Range{start: %LXical.Document.Position{line: 6, character: 11, valid?: true, context_line: {:line, "    Dummy.", "\n", 6, true}, document_line_count: 9, starting_index: 1}, end: %LXical.Document.Position{line: 6, character: 11, valid?: true, context_line: {:line, "    Dummy.", "\n", 6, true}, document_line_count: 9, starting_index: 1}}}}]>, #Protocol.Types.Completion.Item<[insert_text_format: :snippet, kind: :function, label: "hello()", sort_text: "094_hello()", text_edit: %LXical.Document.Changes{document: #LXical.Document<path: "/Users/scottming/Code/dummy/test/dummy_test.exs", version: 16, dirty?: true, lines: %Lines<"defmodule DummyTest do..."(9 lines)>, ...>, edits: %LXical.Document.Edit{text: "hello()", range: %LXical.Document.Range{start: %LXical.Document.Position{line: 6, character: 11, valid?: true, context_line: {:line, "    Dummy.", "\n", 6, true}, document_line_count: 9, starting_index: 1}, end: %LXical.Document.Position{line: 6, character: 11, valid?: true, context_line: {:line, "    Dummy.", "\n", 6, true}, document_line_count: 9, starting_index: 1}}}}]>, #Protocol.Types.Completion.Item<[insert_text_format: :snippet, kind: :function, label: "module_info()", sort_text: "099_module_info()", text_edit: %LXical.Document.Changes{document: #LXical.Document<path: "/Users/scottming/Code/dummy/test/dummy_test.exs", version: 16, dirty?: true, lines: %Lines<"defmodule DummyTest do..."(9 lines)>, ...>, edits: %LXical.Document.Edit{text: "module_info()", range: %LXical.Document.Range{start: %LXical.Document.Position{line: 6, character: 11, valid?: true, context_line: {:line, "    Dummy.", "\n", 6, true}, document_line_count: 9, starting_index: 1}, end: %LXical.Document.Position{line: 6, character: 11, valid?: true, context_line: {:line, "    Dummy.", "\n", 6, true}, document_line_count: 9, starting_index: 1}}}}]>, #Protocol.Types.Completion.Item<[insert_text_format: :snippet, kind: :function, label: "module_info(key)", sort_text: "099_module_info(key)", text_edit: %LXical.Document.Changes{document: #LXical.Document<path: "/Users/scottming/Code/dummy/test/dummy_test.exs", version: 16, dirty?: true, lines: %Lines<"defmodule DummyTest do..."(9 lines)>, ...>, edits: %LXical.Document.Edit{text: "module_info(${1:key})", range: %LXical.Document.Range{start: %LXical.Document.Position{line: 6, character: 11, valid?: true, context_line: {:line, "    Dummy.", "\n", 6, true}, document_line_count: 9, starting_index: 1}, end: %LXical.Document.Position{line: 6, character: 11, valid?: true, context_line: {:line, "    Dummy.", "\n", 6, true}, document_line_count: 9, starting_index: 1}}}}]>, #Protocol.Types.Completion.Item<[insert_text_format: :snippet, kind: :function, label: "print_hello(s)", sort_text: "094_print_hello(s)", text_edit: %LXical.Document.Changes{document: #LXical.Document<path: "/Users/scottming/Code/dummy/test/dummy_test.exs", version: 16, dirty?: true, lines: %Lines<"defmodule DummyTest do..."(9 lines)>, ...>, edits: %LXical.Document.Edit{text: "print_hello(${1:s})", range: %LXical.Document.Range{start: %LXical.Document.Position{line: 6, character: 11, valid?: true, context_line: {:line, "    Dummy.", "\n", 6, true}, document_line_count: 9, starting_index: 1}, end: %LXical.Document.Position{line: 6, character: 11, valid?: true, context_line: {:line, "    Dummy.", "\n", 6, true}, document_line_count: 9, starting_index: 1}}}}]>]
```

It's really weird, When I have time, I will further investigate.

scohen:
part of the contract of the call to boot is that when it's done, the app is ready to use. Making things async breaks this contract.

scottming:
Today I tried to pinpoint the real cause of the autocompletion failure. I discovered that `ElixirSense` does not cause it. Simply loading modules that start with `LXical.Protocol.Types` can resolve this issue.

But do you know why? @scohen

scohen:
It's likely because it can't find the protocol modules can't be found because they're not loaded. I'm not sure why this is happening, but it worries me that it is, and I wonder what other issues we're not finding. 

This change seems very risky to me.

zachallaun:
My 2 cents: It's better to amortize load time during usage instead of up-front. Everyone will have their own preference, but I think that, as a general rule, time-to-things-working should be kept as low as possible.

I don't love the idea of special-casing `LXical.Prototype.Types`. What if we just eagerly load all of Lexical? It will still be much less than loading *everything*, but ensures that no Lexical plumbing is missing.

scottming:
@zachallaun 

I mostly agree with your point. However, loading all the modules starting with `LXical` is still too many.

I have been using this branch for two days now, and I haven't encountered any other issues so far.

I speculate that manually loading all the modules starting with `LXical.Prototype.Types` is successful because there might be some issue with `deftype` preventing Elixir from loading them when they are used, or there might be a problem with the automatic loading process. By the way, these modules all contain `deftype`.

scohen:
> I mostly agree with your point. However, loading all the modules starting with LXical is still too many.

Zach was suggesting loading modules starting with `LXical.Protocol.Types`, not `LXical`.  How does that perform?

scohen:
> I explored this further, and it may be a problem with defenum, which doesn't encode the atom into a number

The problem must be elsewhere, this seems to work correctly:
```
iex(2)> Lexical.Protocol.Types.Completion.Item.new(insert_text_format: :snippet, label: "yep") |> Jason.encode
{:ok, "{\"insertTextFormat\":2,\"label\":\"yep\"}"}
```

scohen:
@scottming I think this points to a bigger problem with this PR. 
I suspect what's happening is that the `Jason.Encoder` implementation for the type isn't being found, and it's reverting back to the map implementation, thus atoms are encoded as strings.

Protocols are another thing that i believe need to be loaded before we have a functional app. 

...this PR continues to worry me.

scottming:
@scohen @zachallaun 

I think I have identified the root cause of the problem. In these lines, when the module is not loaded, it falls back to the previous value.

A module will be automatically loaded when used, but these lines 
https://github.com/lexical-lsp/lexical/blob/ba21277427045ebb2a5f67070fc6941ae78537fd/apps/proto/lib/lexical/proto/field.ex#L277-L280
prevent the module from being used, resulting in this strange issue.

However, for this kind of situation, I'm not sure how to write unit tests. How can I define a module within a test that is initially not loaded, but can still be `ensure_loaded?` later on?

scohen:
I'm R- on this PR. It's too dangerous. 

I looked for `function_exported?` and it occurs in modules we don't control (like `Future.Code` and `Code`).

scottming:
> I'm R- on this PR. It's too dangerous.

> I looked for function_exported? and it occurs in modules we don't control (like Future.Code and Code).

@scohen I understand your concern. If we really must use `load_all_modules` of all the dep apps, could we place `load_all_modules` after the `:server` app starts, similar to [--preload-modules](https://github.com/elixir-lang/elixir/commit/4ea903b8cfc99a86fbf4084b25de1f289cd5e287) in `mix app.start`?

Since `load_all_modules` is not a necessary condition for starting the server, this could also allow us to see the *progress message* more quickly(For people who launch nvim *dozens* of times a day, like me, this is a much better experience).

scohen:
> could we place load_all_modules after the :server app starts

The app can start, but the modules need to be loaded before it's initialized

scohen:
> Since load_all_modules is not a necessary condition for starting the server

You have no idea if this is true or not. It might be true today (and i'm not sure it is), but it's possible that it might not be true tomorrow. This is why this change is so risky.

scottming:
I still prefer to initialize as fast as possible.

The changes in the last two commits should have very low risk. However, if you still perceive some risk, let's keep only the commit with `Code.ensure_all_loaded!(modules)`, as that alone can speed up by approximately *1.5* seconds.

apps/server/lib/lexical/server/boot.ex
```diff
@@ -78,8 +79,8 @@ defmodule Lexical.Server.Boot do
   defp load_app_modules(app_name) do
     Application.ensure_loaded(app_name)
 
-    with {:ok, modules} <- :application.get_key(app_name, :modules) do
-      Enum.each(modules, &Code.ensure_loaded!/1)
+    with {:ok, modules} <- Application.spec(app_name, :modules) do
```
scohen, line 82: I think the reason this is faster is that this block never executes. `Application.spec` returns either the list of modules or nil, which causes this arm of the with to not match, and `Code.ensure_all_loaded!` won't run.

scohen (changes requested):

apps/server/lib/lexical/server/boot.ex
```diff
@@ -78,8 +79,8 @@ defmodule Lexical.Server.Boot do
...
+      Code.ensure_all_loaded!(modules)
```
scohen, line 83: This function was introduced in elixir 1.15

scohen:
My suggestion is to switch back to the erlang functions and use [ensure_modules_loaded](https://www.erlang.org/doc/man/code#ensure_modules_loaded-1) and see if that's faster. It might be, since each call to Code in the previous each function requires a trip in and out of the code server. 

It might also be faster to collect all the modules that need to be loaded from each application and load them all once

apps/server/lib/lexical/server/boot.ex
scottming: good catch.

apps/server/lib/lexical/server/boot.ex
scottming: yes, but the we already `alias Future.Code`

and the logic inside is simple, so I think it is fine.

```elixir
  @spec ensure_all_loaded!([module]) :: :ok
  def ensure_all_loaded!(modules) do
    case ensure_all_loaded(modules) do
      :ok ->
        :ok

      {:error, errors} ->
        formatted_errors =
          errors
          |> Enum.sort()
          |> Enum.map_join("\n", fn {module, reason} ->
            "  * #{inspect(module)} due to reason #{inspect(reason)}"
          end)

        raise ArgumentError, "could not load the following modules:\n\n" <> formatted_errors
    end
  end
  ```

**Comment by scottming** [comment]
> It might also be faster to collect all the modules that need to be loaded from each application and load them all once

The test results indicate that individually loading all modules for each app is faster compared to loading all modules of all apps at once.

**each app loading**

```elixir
    Application.ensure_loaded(app_name)
    modules = Application.spec(app_name, :modules)
    Code.ensure_all_loaded!(modules)
```
**results:**
```
~/Code/lexical @3a2c5e68 !1 ❯ ./_build/dev/package/lexical/bin/start_lexical.sh                  5s   25.3.2.7  1.15.7-otp-25
Detected Elixir through asdf: /Users/scottming/.asdf/installs/elixir/1.15.7-otp-25/bin/elixir
[(server 0.3.0) lib/lexical/server/boot.ex:26: Lexical.Server.Boot.start/0]
time #=> 613491

^C
BREAK: (a)bort (A)bort with dump (c)ontinue (p)roc info (i)nfo
       (l)oaded (v)ersion (k)ill (D)b-tables (d)istribution
^C%                                                                                                                             
~/Code/lexical @3a2c5e68 !1 ❯ ./_build/dev/package/lexical/bin/start_lexical.sh                       25.3.2.7  1.15.7-otp-25
Detected Elixir through asdf: /Users/scottming/.asdf/installs/elixir/1.15.7-otp-25/bin/elixir
[(server 0.3.0) lib/lexical/server/boot.ex:26: Lexical.Server.Boot.start/0]
time #=> 630133

^C
BREAK: (a)bort (A)bort with dump (c)ontinue (p)roc info (i)nfo
       (l)oaded (v)ersion (k)ill (D)b-tables (d)istribution
^C%                                                                                                                             
~/Code/lexical @3a2c5e68 !1 ❯ ./_build/dev/package/lexical/bin/start_lexical.sh                       25.3.2.7  1.15.7-otp-25
Detected Elixir through asdf: /Users/scottming/.asdf/installs/elixir/1.15.7-otp-25/bin/elixir
[(server 0.3.0) lib/lexical/server/boot.ex:26: Lexical.Server.Boot.start/0]
time #=> 613521
```

**load all at once**

```elixir
  modules =
    Enum.flat_map(app_names, fn app_name ->
      Application.ensure_loaded(app_name)
      Application.spec(app_name, :modules)
    end)

  Code.ensure_all_loaded!(modules)
```
  
**results**
  
  
 ```elixir
 ~/Code/lexical cancel-manually-load-deps ⇡1 !1 ❯ ./_build/dev/package/lexical/bin/start_lexical.sh
Detected Elixir through asdf: /Users/scottming/.asdf/installs/elixir/1.15.7-otp-25/bin/elixir
[(server 0.3.0) lib/lexical/server/boot.ex:26: Lexical.Server.Boot.start/0]
time #=> 658426

^C
BREAK: (a)bort (A)bort with dump (c)ontinue (p)roc info (i)nfo
       (l)oaded (v)ersion (k)ill (D)b-tables (d)istribution
^C%                                                                                                                             
~/Code/lexical cancel-manually-load-deps ⇡1 !1 ❯ ./_build/dev/package/lexical/bin/start_lexical.sh    25.3.2.7  1.15.7-otp-25
Detected Elixir through asdf: /Users/scottming/.asdf/installs/elixir/1.15.7-otp-25/bin/elixir
[(server 0.3.0) lib/lexical/server/boot.ex:26: Lexical.Server.Boot.start/0]
time #=> 655624

^C
BREAK: (a)bort (A)bort with dump (c)ontinue (p)roc info (i)nfo
       (l)oaded (v)ersion (k)ill (D)b-tables (d)istribution
^C%                                                                                                                             
~/Code/lexical cancel-manually-load-deps ⇡1 !1 ❯ ./_build/dev/package/lexical/bin/start_lexical.sh    25.3.2.7  1.15.7-otp-25
Detected Elixir through asdf: /Users/scottming/.asdf/installs/elixir/1.15.7-otp-25/bin/elixir
[(server 0.3.0) lib/lexical/server/boot.ex:26: Lexical.Server.Boot.start/0]
time #=> 656506

^C
BREAK: (a)bort (A)bort with dump (c)ontinue (p)roc info (i)nfo
       (l)oaded (v)ersion (k)ill (D)b-tables (d)istribution
^C%  
 ```

apps/server/lib/lexical/server/boot.ex
```diff
@@ -77,10 +78,8 @@ defmodule Lexical.Server.Boot do
 
   defp load_app_modules(app_name) do
     Application.ensure_loaded(app_name)
-
-    with {:ok, modules} <- :application.get_key(app_name, :modules) do
-      Enum.each(modules, &Code.ensure_loaded!/1)
-    end
+    modules = Application.spec(app_name, :modules)
+    Code.ensure_all_loaded!(modules)
```
scohen, line 82: This function is not available in Elixir < 1.15. 
The dream of using the elixir stdlib dies here.

```suggestion
  with {:ok, modules} <- :application.get_key(app_name, :modules),
          :ok <- :code.ensure_modules_loaded(modules) do 
         :ok 
  else _ -> 
    raise "Could not load modules in application #{app_name}" 
 end
 
```

scohen (changes requested):

apps/server/lib/lexical/server/boot.ex
scohen: The erlang call that it uses was introduced in erlang 19, so it's compatible.

scohen (approved):

PR #457, opened 2023-10-31
Analysis

### LINKED ISSUE #426: Indexing: Improve the performance of `Aliases`
According to @zachallaun, we spend a _lot_ of time in index generation figuring out where aliases come from. Aliases was implemented simply, and it investigates the AST from scratch each time it's called. A better approach would be to generate some metadata document the first time that it's called, and subsequent calls would be able to use that metadata to figure out what they need to know.

zachallaun:
Fixes #426 

This PR introduces the concept of *analysis*, which is encapsulated by the `%Lexical.Ast.Analysis{}` struct. The idea is that, in the happy case, we can analyze the AST up front to capture important information and then re-use it for subsequent operations.

The most important new APIs/changes introduced:

* `Lexical.Ast.analyze(document)` - returns `%Analysis{}` which may or may not be valid (`analysis.valid?`). Used for whole-document analysis, e.g. indexing.
* `Lexical.Ast.reanalyze_to(analysis, position)` - returns `%Analysis{}` using the fragment of the original document up to the given position. Useful for things that may not require the whole document, e.g. resolving aliases. If the given analysis is already valid, this is a no-op.
* `Lexical.Ast.Aliases` has been removed in favor of analysis.
* Many `Lexical.Ast` functions have been refactored to also (or only) accept an analysis in place of a document.
* Added concept of *derived data* to `Lexical.Document.Store`. This allows for a derived analysis to be saved alongside the document and fetched using `{:ok, %Document{}, %Analysis{}} = Document.Store.fetch(uri, :analysis)`.
* Refactored a number of existing LS features to use analysis, including entity resolution, completions, references.

---

Running the following extremely simple "benchmark" in `iex -S mix`:

```elixir
# on main
iex> time fn -> Lexical.RemoteControl.Search.Indexer.create_index(project()); nil end
Time: 9.7 seconds

# on this branch
iex> time fn -> Lexical.RemoteControl.Search.Indexer.create_index(project()); nil end
Time: 1.8 seconds
```


apps/common/lib/lexical/ast.ex
```diff
@@ -88,6 +88,44 @@ defmodule Lexical.Ast do
...
+  end
+
+  @doc """
+  Attempts to analyze a document at least until `position`.
+
+  This can be used to analyze an invalid AST up until, for instance, the
+  cursor position. The full analysis will be completed if the AST is
+  valid.
+
+  This function is not guaranteed to return a valid analysis.
```
scohen, line 108: then why is it called safe? 🤔 

(i get it, because it can't crash)

apps/common/lib/lexical/ast.ex
```diff
@@ -436,34 +474,15 @@ defmodule Lexical.Ast do
...
-  def expand_aliases(module, %Document{} = document, quoted_document, %Position{} = position)
-      when is_atom(module) and not is_nil(module) do
-    module
-    |> Module.split()
-    |> Enum.map(&String.to_atom/1)
-    |> expand_aliases(document, quoted_document, position)
-  end
-
-  def expand_aliases(segments, %Document{} = document, quoted_document, %Position{} = position)
+  def expand_aliases(segments, analysis_or_document, %Position{} = position)
```
scohen, line 481: this function still has a pretty significant performance footgun if you pass in a document. I think we should be putting the AST (and possibly the analysis) into the document store, and retrieve it from there. This will, I think, require moving AST into `lexical_shared`

apps/common/lib/lexical/ast/analysis/analyzer.ex
```diff
@@ -0,0 +1,424 @@
...
+      %Scope{id: id, range: range, module: module, parent_aliases: parent_aliases}
+    end
+
+    def global do
+      %Scope{id: :global}
+    end
+
+    def alias_map(%Scope{} = scope) do
+      scope.aliases
+      |> Enum.sort_by(& &1.line)
```
scohen, line 36: we should write a comment about this sort before mapping. it looks wrong, but I think that's just to ensure that later aliases in a given scope win.

apps/common/lib/lexical/ast/analysis/analyzer.ex
```diff
@@ -0,0 +1,424 @@
...
+      |> Map.new(&{&1.as, &1})
+      |> Enum.into(scope.parent_aliases)
+    end
+
+    def empty?(%Scope{aliases: []}), do: true
```
scohen, line 41: nit: scopes are about more than just aliases. Imports, function definitions, and variables will all live here soon.

apps/common/lib/lexical/ast/analysis/analyzer.ex
```diff
@@ -0,0 +1,424 @@
...
+    |> Map.reject(fn {_id, scope} -> Scope.empty?(scope) end)
+    |> correct_ranges(quoted, document)
+    |> Map.values()
+  end
+
+  defp preprocess(quoted) do
+    Macro.prewalk(quoted, fn quoted ->
+      quoted
+      |> with_scope_id()
+    end)
```
scohen, line 173:
```suggestion
    Macro.prewalk(quoted, &with_scope_id/1)
```

apps/common/lib/lexical/ast/analysis/analyzer.ex
```diff
@@ -0,0 +1,424 @@
...
+    with %Zipper{node: sibling} <- Zipper.right(zipper),
+         %{start: sibling_start} <- Sourceror.get_range(sibling) do
+      new_end = Position.new(document, sibling_start[:line], sibling_start[:column])
+      put_in(scope.range.end, new_end)
+    else
+      _ ->
+        # we go up twice to get to the real parent because ast pairs
+        # are always in a list
+        %Zipper{node: parent} = zipper |> Zipper.up() |> Zipper.up()
+        %{end: parent_end} = Sourceror.get_range(parent)
```
scohen, line 205:
```suggestion
        parent_end = Sourceror.get_range(parent).end
```

apps/common/lib/lexical/ast/analysis/analyzer.ex
```diff
@@ -0,0 +1,424 @@
...
+  # alias Foo
+  # alias Foo.Bar
+  # alias __MODULE__.Foo
+  defp pre({:alias, meta, [aliases]}, state) do
+    case expand_alias(aliases, state) do
+      [_ | _] = segments ->
+        alias = Alias.new(segments, List.last(segments), meta[:line])
+
+        state
+        |> State.push_alias(alias)
```
scohen, line 280: i really, really, really  don't like single arg pipes
```suggestion
       State.push_alias(state, alias)
```

apps/common/lib/lexical/ast/analysis/analyzer.ex
```diff
@@ -0,0 +1,424 @@
...
+            [{{:__block__, _, [:as]}, {:__aliases__, _, [alias_as]}}]
+          ]},
+         state
+       ) do
+    case expand_alias(aliases, state) do
+      [_ | _] = segments ->
+        alias = Alias.new(segments, alias_as, meta[:line])
+
+        state
+        |> State.push_alias(alias)
```
scohen, line 301:
```suggestion
        State.push_alias(state, alias)
```

apps/common/lib/lexical/ast/analysis/analyzer.ex
```diff
@@ -0,0 +1,424 @@
...
+
+      [] ->
+        state
+    end
+  end
+
+  # clauses: ->
+  defp pre({clause, _, _} = quoted, state) when clause in @clauses do
+    state
+    |> State.maybe_push_scope_for(quoted)
```
scohen, line 311:
```suggestion
    State.maybe_push_scope_for(state, quoted)
```

apps/common/lib/lexical/ast/analysis/analyzer.ex
```diff
@@ -0,0 +1,424 @@
...
+  end
+
+  # blocks: do, else, etc.
+  defp pre({{:__block__, _, [block]}, _} = quoted, state) when block in @block_keywords do
+    state
+    |> State.maybe_push_scope_for(quoted)
```
scohen, line 317:
```suggestion
    State.maybe_push_scope_for(state, quoted)
```

apps/common/lib/lexical/ast/analysis.ex
```diff
@@ -0,0 +1,69 @@
...
+      [] ->
+        %{}
+    end
+  end
+
+  defp scopes_at(%__MODULE__{scopes: scopes}, %Position{} = position) do
+    scopes
+    |> Stream.filter(fn %Analyzer.Scope{range: range} = scope ->
+      scope.id == :global or Range.contains?(range, position)
+    end)
```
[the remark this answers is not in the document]
zachallaun: Not enough; I'll switch to Enum.

apps/common/lib/lexical/ast.ex
zachallaun: I don't love the name, but I wanted *something* that indicated that it's *trying some stuff* -- I don't think "safe" is the right word.

apps/common/lib/lexical/ast/analysis/analyzer.ex
zachallaun: You're right, but that's why this is a function in the `Scope` module instead of an ad-hoc check for empty `scope.aliases` somewhere else. We'll indeed have to update this as scopes start to store more information. (I want to start aggregating imports as soon as possible so that we can start resolving unqualified calls in `Entity.resolve`.)

apps/common/lib/lexical/ast/analysis/analyzer.ex
zachallaun: Yep, a decent amount of this is somewhat leftover from an earlier refactoring where there were a couple of steps here instead of just one.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
```diff
@@ -105,8 +105,7 @@ defmodule Lexical.RemoteControl.Search.Indexer.Extractors.Module do
     {line, column} = reducer.position
     position = Position.new(reducer.document, line, column)
 
-    {:ok, expanded} =
-      Ast.expand_aliases(unresolved_alias, reducer.document, reducer.quoted_document, position)
+    {:ok, expanded} = Ast.expand_aliases(unresolved_alias, reducer.analysis, position)
```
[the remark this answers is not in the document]
zachallaun: I had done a more "thorough integration" with the indexer, but there were enough bits that I was unsure of that I chose for now to have a minimal footprint. At the moment, this is basically just me changing as few indexer LOC as possible in this PR.

apps/common/lib/lexical/ast/analysis/analyzer.ex
zachallaun: Will add a comment and this also made me realize that there was a bug as I was using this function incorrectly in my implementation of `Analysis.aliases_at`.

zachallaun:
Reminder to self: the pattern for detecting `alias Foo, as: ...` is incorrect, as other options can be passed as well, like `warn: false`.

apps/server/lib/lexical/server/application.ex
```diff
@@ -11,15 +14,15 @@ defmodule Lexical.Server.Application do
   @impl true
   def start(_type, _args) do
     children = [
-      Lexical.Document.Store,
-      Lexical.Server,
-      {DynamicSupervisor, Lexical.Server.Project.Supervisor.options()},
+      {Document.Store, derive: [analysis: &Lexical.Ast.analyze/1]},
```
scohen, line 17: This is very cool.

apps/remote_control/lib/lexical/remote_control/dispatch/handlers/indexing.ex
```diff
@@ -24,9 +25,10 @@ defmodule Lexical.RemoteControl.Dispatch.Handlers.Indexing do
     {:ok, state}
   end
 
-  defp reindex(%Document{} = document, quoted_ast) do
-    with :ok <- ensure_latest_version(document),
-         {:ok, entries} <- Indexer.Quoted.index(document, quoted_ast) do
+  defp reindex(uri) do
+    with {:ok, {%Document{} = document, %Analysis{} = analysis}} <-
```
scohen, line 29: I think we should drop the nested tuples here and return `{:ok, %Document{}, %Analysis{}}` Thoughts?

zachallaun:
@scohen @scottming I'm going to mark this as ready for review, as I think it's very close to a merge-able state. I'll update the OP with some additional information.

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -21,15 +22,17 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
 
   Returns `{:ok, resolved, range}` if successful, `{:error, error}` otherwise.
   """
-  @spec resolve(Document.t(), Position.t()) :: {:ok, resolved, Range.t()} | {:error, term()}
-  def resolve(%Document{} = document, %Position{} = position) do
-    with {:ok, surround_context} <- Ast.surround_context(document, position),
-         {:ok, resolved, {begin_pos, end_pos}} <- resolve(surround_context, document, position) do
+  @spec resolve(Analysis.t(), Position.t()) :: {:ok, resolved, Range.t()} | {:error, term()}
+  def resolve(%Analysis{} = analysis, %Position{} = position) do
+    analysis = Ast.analyze_to(analysis, position)
```
zachallaun, line 27: I'm not sure this should be done here. I think this should probably be done before the call to `resolve` is made so that we can ensure `analyze_to` is only called once per request.

projects/lexical_shared/lib/lexical/document/store.ex
```diff
@@ -1,97 +1,148 @@
...
 
-    @table_name Document.Store
+    defrecord :open_doc, document: nil, derived: %{}
 
-    def new do
-      :ets.new(@table_name, [:named_table, :set, :protected, read_concurrency: true])
+    def new(opts \\ []) do
+      [derive: funs] =
+        opts
+        |> Keyword.validate!(derive: [])
```
scohen, line 41: TIL

projects/lexical_shared/lib/lexical/document/store.ex
```diff
@@ -1,97 +1,148 @@
...
+        |> Keyword.take([:derive])
```
scohen, line 42: alternate take: 
```suggestion
      derivation_funs =
        opts
        |> Keyword.validate!(derive: [])
        |> Keyword.get(:derive)
```

projects/lexical_shared/lib/lexical/document/store.ex
```diff
@@ -1,97 +1,148 @@
...
 
     @spec open(t, Lexical.uri(), String.t(), pos_integer()) :: {:ok, t} | {:error, :already_open}
+    def open(%__MODULE__{temporary_open_refs: refs} = store, uri, text, version)
+        when is_map_key(refs, uri) do
+      {_, store} =
+        store
+        |> maybe_cancel_ref(uri)
+        |> pop_open_doc(uri)
+
+      open(store, uri, text, version)
```
scohen, line 94: this seems like it's doing a bit more work that is necessary, why not just cancel the ref?

projects/lexical_shared/lib/lexical/document/store.ex
```diff
@@ -100,123 +151,101 @@ defmodule Lexical.Document.Store do
...
-
-  alias Lexical.Document
-  alias Lexical.ProcessCache
+    defp normalize_error(:error), do: {:error, :not_open}
+    defp normalize_error(e), do: e
 
-  @type t :: %State{}
+    defp derive(%__MODULE__{} = store, key, document) do
+      case store.derivation_funs do
+        %{^key => fun} ->
```
scohen, line 218: maybe: 
```suggestion
        %{^key => fun} when is_function(fun, 1) ->
```

Ideally, this check should be done in the constructor

projects/lexical_shared/lib/lexical/document/store.ex
```diff
@@ -240,24 +269,27 @@ defmodule Lexical.Document.Store do
...
+  @spec start_link(start_opts) :: GenServer.on_start()
+  def start_link(opts) do
+    GenServer.start_link(__MODULE__, opts, name: name())
   end
 
-  def init(_) do
-    {:ok, State.new()}
+  @impl GenServer
+  def init(opts) do
+    {:ok, State.new(opts)}
```
scohen, line 289: consider ensuring the derivation functions are single arity functions and returning an error if they aren't.

projects/lexical_shared/test/lexical/document/store_test.exs
```diff
@@ -179,4 +181,61 @@ defmodule Lexical.Document.StoreTest do
...
+    test "update when the document changes" do
+      assert :ok =
+               Document.Store.update(uri(), fn document ->
+                 Document.apply_content_changes(document, 2, [
+                   build_change(text: "dog")
+                 ])
+               end)
+
+      assert {:ok, doc, 3} = Document.Store.fetch(uri(), :length)
+      assert Document.to_string(doc) == "dog"
```
scohen, line 215: ...and I bet she's a good girl.

scohen (commented):
This change is excellent. 

I have a couple of small comments, but I think we should move forward on this.

Makefile
```diff
@@ -13,10 +13,10 @@ dialyzer.umbrella:
 	mix dialyzer
 
 dialyzer.projects:
-	$(foreach dir, $(dialyzer_dirs), cd projects/$(dir) && mix dialyzer && cd ../..;)
+	$(foreach dir, $(dialyzer_dirs), cd projects/$(dir) && mix dialyzer; cd ../..;)
```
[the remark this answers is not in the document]
zachallaun: It was a minor issue I ran into and figured I'd just fix. The problem was that the `cd ../..` wouldn't run if one of the mix commands failed/errored, which would then lead to a Make error along the lines of "no such directory projects/lexical_shared" because the next loop would still be in the first project directory.

I'm not sure this fix is correct though; it might hide errors if the first mix command fails but the last one succeeds. I'll just revert.

apps/common/test/lexical/ast/analysis_test.exs
```diff
@@ -0,0 +1,51 @@
...
+               Analysis.aliases_at(analysis, position(6, 1))
+
+      assert %{Inner1: _, Foo: _, Outer: _, __MODULE__: _} =
+               Analysis.aliases_at(analysis, position(7, 1))
+
+      assert %{Baz: _, Inner2: _, Inner1: _, Foo: _, Outer: _, __MODULE__: _} =
+               Analysis.aliases_at(analysis, position(10, 1))
+
+      assert %{Inner2: _, Inner1: _, Foo: _, Outer: _, __MODULE__: _} =
+               Analysis.aliases_at(analysis, position(10, 6))
```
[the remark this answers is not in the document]
zachallaun: Heh, whoops -- these were actually sort of "dev tests" that weren't intended to be committed as-is. Will factor them out.

projects/lexical_shared/lib/lexical/document/store.ex
zachallaun: My thought was that this means you can open a temporarily open doc as permanent using different text/version. Perhaps I misunderstood what the desired behavior is here, though?

scohen: hmm... opening things temporarily is done by the server itself, I didn't consider differing (possibly lower) versions. What you have is correct.

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
zachallaun: Nevermind; it's fine for this to happen here because this is just a no-op if `analyze_to` was called earlier.

scohen (approved):

PR #458, opened 2023-10-31
Skip build directories _build and deps in indexing

### LINKED ISSUE #456: Skip build directory in indexing
Presently, we index both the deps and _build directory, so source files in both directories will show up with when you call find references. This raises two questions:
 
  1. Should we index the `_build` directory?
  2. Should we index the `deps` directory

I think the answer to the first question is definitely no.
I'm not so sure about if we should skip the deps directory. I've found it sometimes useful to see usages of a particular module in dependencies, but I can also understand that results from dependencies clutter the results. 

This is a relatively easy fix; we just need to remove the build (and maybe the deps) directories from the [list of indexable files generated](https://github.com/lexical-lsp/lexical/blob/a6dd1011daa19c310b767d89d74126ac2e18881c/apps/remote_control/lib/lexical/remote_control/search/indexer.ex#L126).

milinddethe15:
fixes #456

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/indexer.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/indexer.ex
@@ -128,7 +128,7 @@ defmodule Lexical.RemoteControl.Search.Indexer do
 
     [root_dir, "**", @indexable_extensions]
     |> Path.join()
-    |> Path.wildcard()
+    |> Path.wildcard(~r"^(?!.*(build|deps)).*$")
   end
 
   # stat(path) is here for testing so it can be mocked
--- a/apps/remote_control/lib/lexical/remote_control/search/indexer.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/indexer.ex
@@ -128,7 +128,7 @@ defmodule Lexical.RemoteControl.Search.Indexer do
 
     [root_dir, "**", @indexable_extensions]
     |> Path.join()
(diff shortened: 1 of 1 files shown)


zachallaun (commented):
What happens here if you have a module called `MyProject.Deps.Something` that lives in `lib/my_project/deps/something.ex`? It looks like this will exclude indexing anything in any `deps` folder regardless of nesting. 

What we may need to do -- and might as well do it for `_build` too? -- is to only exclude these folders in the project root, or, if the project is an umbrella project, in the folder of each umbrella app.

scohen:
@milinddethe15 There are safe ways to get the build and deps directories, provided by @[Mix.Project](https://hexdocs.pm/mix/1.12.3/Mix.Project.html#deps_path/1). In order to use it, you'll need to wrap the usage like this:
```elixir
alias Lexical.RemoteControl
...
RemoteControl.Mix.in_project fn -> 
  Mix.Project.deps_path
end
```

zachallaun:
> @zachallaun I can't figure out why we'd want to index _build. Do you have ideas that I'm not having?

No, it's probably safe to always exclude `_build` regardless of location -- it was more of a principled stance to only ignore paths we expect to ignore, i.e. those returned by Mix.Project, as you suggested.

scohen:
@milinddethe15 any chance you can work on this? Otherwise, I'm going to close it.

PR #463, opened 2023-11-03
Ignore `undefined variables` error for HEEx files

### LINKED ISSUE #460: Undefined variable in heex template
Lexical is emitting an error saying that the `requested_ranges_form` variable is undefined, even though it is defined a little bit earlier in a `:let` attribute.

This is my day job code so I cannot post the entire code snippet but here is a screenshot:

image

The error reported by Lexical is `undefined variable "requested_ranges_form"`. When running `mix compile --force` it does not report any warning on error as expected.

scottming:
The error comes from EEx's `eval_quoted`. However, the EEx engine cannot directly recognize variables defined with `:let`. This forces me to ignore diagnostics that start with `undefined variable` for both EEx and HEEx files.

Additionally, as I mentioned in the initial PR, HEEx files heavily rely on definitions from other Elixir files. Therefore, when we compile a single file using the HEEx engine, there are bound to be some limitations.

Fixes #460

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/build/document/compilers/eex.ex
+++ b/apps/remote_control/lib/lexical/remote_control/build/document/compilers/eex.ex
@@ -45,12 +45,13 @@ defmodule Lexical.RemoteControl.Build.Document.Compilers.EEx do
     end
   end
 
-  def eval_quoted(%Document{} = document, quoted_ast) do
+  @spec eval_quoted(Document.t(), Macro.t(), String.t()) :: :ok | {:error, [Result.t()]}
+  def eval_quoted(%Document{} = document, quoted_ast, eval_for \\ "eex") do
     result =
       if Elixir.Features.with_diagnostics?() do
-        eval_quoted_with_diagnostics(quoted_ast, document.path)
+        eval_quoted_with_diagnostics(quoted_ast, document.path, eval_for)
       else
-        do_eval_quoted(quoted_ast, document.path)
+        do_eval_quoted(quoted_ast, document.path, eval_for)
       end
 
     case result do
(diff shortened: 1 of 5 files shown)


apps/remote_control/lib/lexical/remote_control/build/document/compilers/eex.ex
```diff
@@ -111,6 +112,20 @@ defmodule Lexical.RemoteControl.Build.Document.Compilers.EEx do
...
+  defp reject_undefined_variables(errors) do
+    # NOTE: Ignoring error for undefined variables
+    # for example:
+    # ```heex
+    # <.inputs_for :let={finner} field={f[:inner]}>
+    #  <%= finner %>
+    # </.inputs_for>
+    # ```
+    # the `finner` variable is defined by `:let` of `inputs_for`
+    Enum.reject(errors, fn %Result{message: message} ->
```
scohen, line 124: where do the functions come from? Is there a way we can either include them into the heex template or ignore the functions that are default imported?

scottming: yeah, I reverted this, and tried another approach.

crbelaus:
I've tested this branch in my computer and I can confirm that it fixes the issue that I reported on #460

apps/remote_control/lib/lexical/remote_control/build/document/compilers/eex.ex
scohen: awesome

scohen (approved):

scohen:
Thank you, @crbelaus

PR #478, opened 2023-11-08
Added completions for typespecs

### LINKED ISSUE #473: Completion for typespecs
Currently, we don't complete typespecs, looks like an oversight.

scohen:
Prior, we were ignoring typespecs as completion candidates because they are only relevant in a couple of places. This adds detection for those places and only emits completion candidates if they're in a type, typep, opaque or spec.

Additionally, while in one of those contexts, it suppresses most other completions, because they are not relevant.

Fixes #473

### DIFF
--- a/apps/common/lib/lexical/ast/env.ex
+++ b/apps/common/lib/lexical/ast/env.ex
@@ -236,6 +236,14 @@ defmodule Lexical.Ast.Env do
     end)
   end
 
+  defp do_in_context?(env, :type) do
+    ancestor_is_type?(env)
+  end
+
+  defp do_in_context?(env, :spec) do
+    ancestor_is_spec?(env)
+  end
+
   defp do_in_context?(env, :import) do
     in_directive?(env, ~c"import")
   end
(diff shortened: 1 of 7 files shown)


apps/common/lib/lexical/ast/env.ex
```diff
@@ -415,12 +423,52 @@ defmodule Lexical.Ast.Env do
...
+      {:__block__, _, [{:@, metadata, [{:spec, _, _}]}, _]} ->
+        # multi-line spec
+        cursor_in_range?(env, metadata)
+
+      _ ->
+        false
     end)
   end
+
+  defp cursor_in_range?(env, metadata) do
```
scottming, line 462: I would like to know the difference between this function and `Lexical.Ast .contains_position`?

scottming (approved):

apps/common/lib/lexical/ast/env.ex
scohen: I was unaware of that function. I'll try it out and switch.

scohen: actually scratch that. The difference is that `Ast.contains_position?` uses `Sourceror.get_range`, which examines a block of ast and tells you the extent of that.  This function takes a single expression and uses the extent of it in the metadata to figure out the range.

PR #481, opened 2023-11-09
Fix: Don't raise an exception if the build directory doesn't exist

scohen:
If the build directory didn't exist (as it didn't in CI), then an exception was raised when we tried to delete it. This would cause remote control to not boot and the whole system would crash around us. 

Simple solution, don't use bang functions.


scohen:
Note: two commits here, the first is a revert of an already approved PR.

Issue #474, opened 2023-11-07
Unhelpful :: operator completion


scohen:
The operator `::` has a very unhelpful completion, completing to `::(left, right)`. It should be suppressed.

This is easy, it's a macro, and should be skipped by [adding a pattern match](https://github.com/lexical-lsp/lexical/blob/main/apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex#L550).

mdshamoon:
Hi @scohen can I pick up this issue. The project has been really helpful for me. Wanted to contribute back.

scohen:
Fixed via #485

PR #485, opened 2023-11-12
Removed unhelpful completion for :: symbol

mdshamoon:
### Solution description

As mentioned by scohen added this symbol :: in the list of @snippet_macros to not show any completions. 
Target issue #474 

This PR is ready for review. Feedbacks would be really helpful

### DIFF
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
@@ -8,7 +8,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
   alias Lexical.Server.CodeIntelligence.Completion.Translations.Struct
 
   @snippet_macros ~w(def defp defmacro defmacrop defimpl defmodule defprotocol defguard defguardp defexception test use)
-  @unuseful_macros ~w(:: alias! in and or destructure)
+  @unhelpful_macros ~w(:: alias! in and or destructure)
 
   defimpl Translatable, for: Candidate.Macro do
     def translate(macro, builder, %Env{} = env) do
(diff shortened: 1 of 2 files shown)


apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
```diff
@@ -7,7 +7,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
   alias Lexical.Server.CodeIntelligence.Completion.Translations.Callable
   alias Lexical.Server.CodeIntelligence.Completion.Translations.Struct
 
-  @snippet_macros ~w(def defp defmacro defmacrop defimpl defmodule defprotocol defguard defguardp defexception test use)
+  @snippet_macros ~w(def defp defmacro defmacrop defimpl defmodule defprotocol defguard defguardp defexception test use ::)
```
scottming, line 10: I don't think `::` should be considered a snippet_macros. I would prefer to create another attribute, something like:

```elixir
@unuseful_macros ~w(::)
```

By the way, I believe the completions for alias!, in, and and are not useful either. We can include them in the @By the way, I believe the completions for alias!, in, and and are not useful either. We can include them in the @unuseful_macros list as well. list as well, and at the caller side, we can do something like: `@unuseful_macros ++ @snippet_macros`

scohen: I'd just make a function head that skips it, I think it's the only one

mdshamoon: Updated with the feedbacks

scohen (approved):

scottming (approved):
Lgtm

PR #489, opened 2023-11-15
Heavy Refactor: Ast.Env

scohen:
The Ast.Env module was accruing a lot of functionality related to detecting the "part of speech" of code at a given point. It was over 600 lines, and its test was approaching 1000.

Additionally, while the tests for context detection did an admirable job for positive cases, their negative cases (ensure that this code is _not_ a given context) were haphadard, inconsistent or missing entirely.

This PR pulls out context detection into separate modules, each with their own test cases. What's more, the test cases now have a large test suite, that generates both positive _and_ negative tests for each module.

It does this by marking each case with a path of what it's testing, and marking the code with ranges where that context should occur. The test case then performs *exhaustive* testing on the code, going token by token and ensuring that the given token is either in or not in the requested context. This PR also adds over 900 automated unit tests to enforce detection is accurate.

The end result is a lot of edge cases were found and fixed in this PR, and some of the code can be further improved with more confidence that the result is correct, and there won't be regressions.


apps/common/lib/lexical/ast/detection/ancestor.ex
zachallaun, line 1: I've been confused by modules like these before (helper modules in a directory that otherwise contains behaviour implementations). Two alternatives that I can think of:

1. Use a `Helpers` suffix for modules like this, so this becomes `*.Detection.AncestorHelpers`
2. If we don't anticipate other, separate helpers, we could put these in `Detection`, define a `__using__`, and import them:

   ```elixir
   defmacro __using__(_) do
     quote do
       @behaviour unquote(__MODULE__)
       import unquote(__MODULE__)
     end
   end
   ```

   and then replace the various instances of `@behaviour Detection` with `use Detection`.

scohen: i think i favor the second approach, agreed with the confusion of the behaviour implementations mixed with helpers.

apps/common/lib/lexical/ast/detection/pipe.ex
```diff
@@ -0,0 +1,38 @@
...
+  alias Lexical.Ast.Tokens
+  alias Lexical.Document
+  alias Lexical.Document.Position
+
+  use Detection
+
+  @impl Detection
+  def detected?(%Document{} = document, %Position{} = position) do
+    # Document.fragment(document, Position.new(document, position.line, 1), position)
+    # |> IO.inspect()
```
scottming, line 12:
```suggestion
```

apps/common/lib/lexical/ast/detection/pipe.ex
```diff
@@ -0,0 +1,38 @@
...
+
+    document
+    |> Tokens.prefix_stream(position)
+    |> Enum.to_list()
+    #    |> IO.inspect()
```
scottming, line 17:
```suggestion
```

apps/common/lib/lexical/ast/detection/struct_reference.ex
```diff
@@ -0,0 +1,43 @@
...
+        starts_with_percent? =
+          document
+          |> Tokens.prefix_stream(position)
+          |> Enum.take(2)
+          |> Enum.any?(fn
+            {:percent, :%, _} -> true
+            _ -> false
+          end)
+
+        starts_with_percent? and possible_dunder_module(possible_module_struct) and
```
scottming, line 32: what is `dunder`?

scottming (changes requested):

apps/common/test/lexical/ast/detection/struct_field_value_test.exs
```diff
@@ -0,0 +1,25 @@
+defmodule Lexical.Ast.Detection.StructFieldValueTest do
+  alias Lexical.Ast.Detection
+
+  use Lexical.Test.DetectionCase,
+    for: Detection.StructFieldValue,
+    assertions: [[:struct_field_value, :*]],
+    skip: [
+      [:struct_fields, :*],
+      [:struct_reference, :*],
+      [:struct_field_key, :*]
```
scottming, line 10: Aren't key and value mutually exclusive? I don't understand why `field_key` is skipped here.

apps/common/lib/lexical/ast/detection/struct_reference.ex
scohen: no, dunder means `double under`, and refers to `__`

apps/common/test/lexical/ast/detection/struct_field_value_test.exs
scohen: because values can appear inside the tests for keys, and the tests will fail because of this. 
Keep in mind, those tests show where keys can appear, and if anything else appears in the test, they will fail.

scohen:
@scottming  please re-review.

apps/common/test/lexical/ast/detection/struct_reference_test.exs
```diff
@@ -0,0 +1,45 @@
...
+
+  test "is detected if the reference is for %__} " do
+    assert_detected ~q[%__]
+  end
+
+  test "is not detected if the reference is for %__MOC in a function definition" do
+    refute_detected ~q[def my_fn(%__MOC)]
+  end
+
+  test "is not detected if module reference doesn't start with %" do
```
scottming, line 38: assertions and description do not match

apps/common/test/lexical/ast/detection/struct_fields_test.exs
```diff
@@ -0,0 +1,32 @@
...
+  test "is true even if the value of a struct key is a tuple" do
+    assert_detected ~q[%User{«favorite_numbers: {3}»}]
+  end
+
+  test "is true even if the cursor is at a nested struct" do
+    assert_detected ~q[%User{«address: %Address{}»]
+  end
+
+  test "is detected if it spans multiple lines" do
+    assert_detected ~q[
```
scottming, line 26: what's the difference between this test and 

https://github.com/lexical-lsp/lexical/blob/89041565265ddbc06c6adb7429a8524ecc7aad6d/apps/common/test/support/lexical/test/detection_case/suite.ex#L118-L122

scottming (approved):

apps/common/test/lexical/ast/detection/struct_fields_test.exs
scohen: these tests were just copied from what existed before, I didn't audit them to see if they overlapped, though in this example, the distinction is that the struct fields doesn't include the values.

PR #492, opened 2023-11-17
Index variable definitions and usages

scottming:
This PR adopts a combination of operators to extract variable definitions and usages. There are a total of 4 commits, which should address most cases. However, I have not yet handled the `else` block. I would like to gather some feedback in advance.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/indexer/metadata.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/indexer/metadata.ex
@@ -2,6 +2,31 @@ defmodule Lexical.RemoteControl.Search.Indexer.Metadata do
   @moduledoc """
   Utilities for extracting location information from AST metadata nodes.
   """
+  alias Sourceror.Zipper
+
+  # fn -> things end
+  def location({:->, metadata, [[], _]}) do
+    {:expression, position(metadata)}
+  end
+
+  def location({:->, meta, [left, {:__block__, right_meta, right_blocks}]}) do
+    block_start = arrow_start_position(left)
+
+    if pos = position(right_meta, :closing) do
+      block_end = pos
+      {:block, position(meta), block_start, block_end}
+    else
+      block_end = arrow_last_position(right_meta, right_blocks)
+      {:block, position(meta), block_start, block_end}
+    end
+  end
+
+  # fn x -> x end
+  def location({:->, meta, [left, {_, right_meta, nil}]}) do
+    block_start = arrow_start_position(left)
+    {:block, position(meta), block_start, position(right_meta)}
+  end
+
   def location({_, metadata, _}) do
     if Keyword.has_key?(metadata, :do) do
       position = position(metadata)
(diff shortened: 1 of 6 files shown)


apps/remote_control/lib/lexical/remote_control/search/indexer/metadata.ex
```diff
@@ -35,4 +60,35 @@ defmodule Lexical.RemoteControl.Search.Indexer.Metadata do
     |> Keyword.get(key, [])
     |> position()
   end
+
+  defp arrow_start_position([{:when, _, blocks} | _]) do
+    arrow_start_position(blocks)
+  end
+
+  defp arrow_start_position(node) do
+    start_node = %Zipper{node: node} |> Zipper.leftmost() |> Zipper.next() |> Zipper.node()
```
zachallaun, line 69:
```suggestion
    start_node = node |> Zipper.zip() |> Zipper.leftmost() |> Zipper.next() |> Zipper.node()
```

zachallaun (commented):
I still need to think this through some more, but I've been thinking about the relationship between analysis and indexing/extractors:

* analysis is all about pre-analyzing scopes and tying definitions to a position. It should replace the notion of blocks in extractors.
* extractors should possibly have two callbacks: 1) that takes an analysis and returns any definitions that should be indexed, 2) the existing extract callback that now focuses on finding *usages* within the ast
* I started working on tracking imports during analysis last night and made some good progress on a small refactor that will make it easier to extract different features and associate them with the correct scope/definition (e.g. an argument in a fun definition, or a variable binding in a block)
* this would change what extractors need to do, when they encounter a usage, they'd use the analysis to "look up" the associated definition so that they can get its ref for indexing

I think the important bit is that analysis should contain the canonical definition of "scope" -- the indexer shouldn't have to do any scope tracking, it should just walk the ast and collect any usages we care about, and be able to refer to the analysis for definitions. Basically, extractors ask analysis, "I found this usage of `foo` at this position, what does it refer to?"

scottming:
@zachallaun 

> I think the important bit is that analysis should contain the canonical definition of "scope" -- the indexer shouldn't have to do any scope tracking, it should just walk the ast and collect any usages we care about, and be able to refer to the analysis for definitions. Basically, extractors ask analysis, "I found this usage of foo at this position, what does it refer to?"

I mostly agree with your point. It makes sense to separate the logic of defining scope and the extractor.

While implementing this PR, I noticed that our reducer does not have the concept of ancestors. So, I implemented a simple version of it. However, constructing it every time we search for definitions leads to poor performance.

If the analysis could provide answers to these questions you mentioned, our work would undoubtedly be much simpler.

zachallaun:
> While implementing this PR, I noticed that our reducer does not have the concept of ancestors. So, I implemented a simple version of it. However, constructing it every time we search for definitions leads to poor performance.
> 
> If the analysis could provide answers to these questions you mentioned, our work would undoubtedly be much simpler.

Yep! I certainly believe analysis could/should be answering these going forward. I'll look into what changes need to be made to analysis to support this.

PR #493, opened 2023-11-17
Fix find-reference

scottming:
find-reference is broken on the main branch.


zachallaun (approved):

zachallaun:
Whoops!

I'm surprised there isn't a test that caught this. Worth adding one?

scottming:
> I'm surprised there isn't a test that caught this. Worth adding one?

I think matching the `Analysis` struct here: https://github.com/lexical-lsp/lexical/pull/493/files#diff-cee6d5885f568da52d9067503a71fa4f9f0551e2fd1a49e9a8a0ebf40170e22dR52

Then it should be fine.

PR #496, opened 2023-11-20
Fix module completion error after a dot

zachallaun:
I'm having trouble replicating this in a test, but I was getting an error when I tried completing something like:

```elixir
defmodule Something do
  alias Lexical.|
end
```

The error was caused by an exception thrown in analysis due to an incorrect case in its `reify_alias` implementation. The analyzer deals with modules as lists of segments, e.g. `[:Foo, :Bar, :Baz]`, but the macro env requires that they be combined to `Foo.Bar.Baz`.


scohen:
crazy, this works great for me.

scohen:
The gift that keeps on giving.

scohen:
I'm a little worried about the lack of a unit test. Would it be possible to either 
  1. Make a very specific unit test targeting env
  2. Use patch to expose `refiy_alias` and test that way

scohen (approved):

PR #497, opened 2023-11-21
Added indexing for function definitions using `def`

### LINKED ISSUE #466: Indexing: Function definitions
The indexer should create entries for function definitions using `def` and `defp`.

scohen:
Added indexing support for functions defined with def and defp. Both are stored in the index with different types, :public_function and :private_function.

I also added support for recognizing a local function call at a position, and added the ability to extract a range from a document to RangeSupport.

Fixes #466


apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/function_definition.ex
```diff
@@ -0,0 +1,74 @@
...
+  alias Lexical.RemoteControl.Search.Indexer.Entry
+  alias Lexical.RemoteControl.Search.Indexer.Metadata
+  alias Lexical.RemoteControl.Search.Indexer.Source.Block
+  alias Lexical.RemoteControl.Search.Indexer.Source.Reducer
+
+  @function_definitions [:def, :defp]
+
+  def extract({definition, metadata, [{fn_name, _, args}, _body]} = ast, %Reducer{} = reducer)
+      when is_atom(fn_name) and definition in @function_definitions do
+    range = get_extent(reducer.analysis, metadata)
```
scottming, line 16: I'm curious about the benefits of having `get_extent` instead of `get_range` here.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/function_definition.ex
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  defp get_extent(%Analysis{} = analysis, metadata) do
+    {line, column} = Metadata.position(metadata)
+
+    {end_line, end_column} =
+      with nil <- Metadata.position(metadata, :do),
+           nil <- Metadata.position(metadata, :end_of_expression),
+           {:ok, line_text} <- Document.fetch_text_at(analysis.document, line) do
+        end_column = String.length(line_text)
```
scottming, line 66: I believe the range of function definitions should not encompass an entire line. On one hand, many functions span multiple lines, so this range is not accurate. On the other hand, `defp`, `do`, and similar elements are redundant information. Even including parameters is unnecessary. Including only the function name would be more friendly for future implementations such as `document symbol` and `rename`.

The highlighting will also be more consistent.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/function_definition.ex
scohen: I'm not following. Is this a naming concern?

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/function_definition.ex
scohen: this is really one of those "what's the essential nature of X" arguments that programmers like to have, and have no answer. 
What is the definition of a function? I would argue that it is the name and parameter list. Elixir agrees with this definition, as functions are defined by their name and arity. 

> so this range is not accurate.

I'm a little puzzled how you've reached this conclusion. The first two heads in the `with` attempt to figure out where the function definition ends. Pulling the entire line only happens when those heuristics have failed, which only happens on single line functions (and there are tests for those).

>  Including only the function name would be more friendly for future implementations such as document symbol and rename

This is definitely not true, those features will be given a range that is actually valid syntax (a `def` to a `do`) that can be turned into AST with `Code.Fragment.container_cursor_to_quoted` and modified with our AST tools.

scottming: I said it was inaccurate because I didn't see any tests related to *multi-lines*. However, it was a bit late last night, and I didn't look at it very carefully.

Today, I tried adding tests:

```elixir
    test "finds multi line private functions" do
      code = ~q[
        defmodule Privates do
          defp multi_lines(
            a,
            b,
            c,
            d) do
          end
        end
        ]

      {:ok, [multi_line], _} = index_functions(code)

      assert multi_line.type == :private_function
      assert multi_line.subtype == :definition
      assert multi_line.subject == "Privates.multi_lines/4"
      assert "defp multi_lines    a,\n    b,\n    c,\n    d) do" == extract(code, multi_line.range)
    end
```

and if we follow the approach you mentioned, in this particular test, the end_pos should be `{6, 10}` instead of `{6, 8}`.

```elixir
  1) test indexing private function definitions finds multi line private functions (Lexical.RemoteControl.Se
arch.Indexer.SourceTest)
     test/lexical/remote_control/search/indexer/source_test.exs:649
     Assertion with == failed
     code:  assert "defp multi_lines    a,\n    b,\n    c,\n    d) do" == extract(code, multi_line.range)
     left:  "defp multi_lines    a,\n    b,\n    c,\n    d) do"
     right: "defp multi_line    a,\n    b,\n    c,\n    d) d"
     stacktrace:
       test/lexical/remote_control/search/indexer/source_test.exs:666: (test)


Finished in 0.01 seconds (0.00s async, 0.01s sync)
38 tests, 1 failure, 37 excluded
```

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/function_definition.ex
scottming: It's not a concern, my English isn't good. From what I understand, these two words are often interchangeable. I just want to know under what circumstances they would be interchangeable. Here both words appear, the variable is 'range', and the function name is 'get_extent'.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/function_definition.ex
scohen: this is a bug with extract, not the ranges.

scottming: However, I don't think returning 6, 8 is correct here because based on the screenshot, it should be 6, 10.

CleanShot 2023-11-25 at 10 29 33@2x

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/function_definition.ex
scohen: this sounds like a naming concern. `get_range` sounded pretty generic and non-descriptive to me, so I picked `get_extent` though perhaps. `get_definition_range` will be nicer? I'm open to other names.

scottming: > get_definition_range will be nicer?

👍🏻

zachallaun:
Couple of suggestions but looks good.

scohen:
I'll make the changes on the other PR.

PR #498, opened 2023-11-21
Move from flake-utils to flake-parts

hauleth:
Ref #490

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,20 +1,20 @@
 {
   "nodes": {
-    "flake-utils": {
+    "flake-parts": {
       "inputs": {
-        "systems": "systems"
+        "nixpkgs-lib": "nixpkgs-lib"
       },
       "locked": {
-        "lastModified": 1692799911,
-        "narHash": "sha256-3eihraek4qL744EvQXsK1Ha6C3CR7nnT8X2qWap4RNk=",
-        "owner": "numtide",
-        "repo": "flake-utils",
-        "rev": "f9e7cf818399d17d347f847525c5a5a8032e4e44",
+        "lastModified": 1698882062,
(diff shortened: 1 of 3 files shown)


zoedsoupe:
Yes, it would work on an intel CPU, the problem is on M1 CPU family as referenced on the linked pull request.

Then, as it impossible to override normally the derivation, maybe a 2 arity function like

```nix
mkLexical = {erlang, elixir ? erlang.elixir}: erlang.callPackage ./nix/lexical.nix {inherit elixir;};
```

hauleth:
With my code to change Elixir version you need to do:
```nix
my-lexical = mkLexical {
  erlang = pkgs.beam.packages.erlang.extend (efinal: eprev: {
    elixir = eprev.elixir_1_12;
  });
};
```
The thing with having `erlang` and `elixir` as a separate values is that it may end that there will be difference in the OTP and Elixir implementation used in the Lexical derivation. If anything I would try to introduce fixes into the Nixpkgs to make it work without these workarounds.

zoedsoupe:
Ok, that's fine!Zoey de Souza PessanhaDesenvolvedora de Softwarehttps://github.com/zoedsoupeOn 21 Nov 2023, at 13:02, Łukasz Jan Niemier ***@***.***> wrote:﻿
With my code to change Elixir version you need to do:
my-lexical = mkLexical {
  erlang = pkgs.beam.packages.erlang.extend (efinal: eprev: {
    elixir = eprev.elixir_1_12;
  });
};
The thing with having erlang and elixir as a separate values is that it may end that there will be difference in the OTP and Elixir implementation used in the Lexical derivation. If anything I would try to introduce fixes into the Nixpkgs to make it work without these workarounds.

zoedsoupe (approved):

scohen (approved):

PR #500, opened 2023-11-22
Make the warning: prefix optional in rename function regex

### LINKED ISSUE #499: Code action for replacing remote function doesn't work as expected
```elixir
Enum.counts([1, 2, 3])
```
yields a compilation warning as expected:
```
warning: Enum.counts/1 is undefined or private. Did you mean:

      * concat/1
      * concat/2
      * count/1
      * count/2
```

I would expect a code action to be generated for the line that would propose to rename the function to either `count` or `concat`, but it doesn't happen.

After a closer look, it seems that the error message in the field `message` in `%Diagnostic{}` is expected by the code action to start with `warning: ....` but it doesn't; instead it starts from `Enum.counts/1 is undefined ...`. I assume the warning gets parsed into a separate field of the struct, `severity`, which is correctly set to `:warning`. 

I have compiled Lexical with Erlang 24.3.4.12 and Elixir 1.13.4-otp-24, and the project is using OTP 26.1.2 and Elixir 1.15.7-otp-26.

I assume the fix would entail making the `warning: ` part of the regex optional.

soundmonster:
Potentially fixes #499

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/code_action/handlers/replace_remote_function.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_action/handlers/replace_remote_function.ex
@@ -86,10 +86,10 @@ defmodule Lexical.RemoteControl.CodeAction.Handlers.ReplaceRemoteFunction do
     end
   end
 
-  @function_re ~r/warning: ([^\/]+)\/(.*) is undefined or private. Did you mean:.*/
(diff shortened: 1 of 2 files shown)


scohen (approved):

scohen:
LGTM, thanks for the fix.

PR #503, opened 2023-11-27
Refactor indexer to better leverage analysis

zachallaun:
Work in progress

In my extremely informal "benchmark" on my machine, this does not affect performance. Both this branch and main report within ~0.1 second of each other when running:

```elixir
iex> time fn -> Lexical.RemoteControl.Search.Indexer.create_index(project()); nil end
Time: 1.7 seconds
```


scottming:
> Both this branch and main report within ~0.1 second of each other when running:

On my laptop, the main branch takes approximately 0.8 s, and this branch takes 1.5 s.

zachallaun:
> > Both this branch and main report within ~0.1 second of each other when running:
> 
> On my laptop, the main branch takes approximately 0.8 s, and this branch takes 1.5 s, However, performance is not currently the focus of my attention.

Thank you for checking! I'll do some more targeted benchmarking and see if the difference can be minimized. I don't want performance to regress 2x for some.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
```diff
@@ -4,137 +4,111 @@ defmodule Lexical.RemoteControl.Search.Indexer.Extractors.Module do
...
-    {line, column} = reducer.position
-    position = Position.new(reducer.analysis.document, line, column)
-
-    {:ok, expanded} = Ast.expand_alias(unresolved_alias, reducer.analysis, position)
-
-    expanded
+  defp extract_module(%Extractor{} = extractor, module, %Zipper{} = zipper) do
+    {zipper, subtype, parent_kind} =
+      case Zipper.up(zipper) do
+        %Zipper{node: {:defmodule, _, [_, _]}} = defmodule_zipper ->
```
scottming, line 52: Using a zipper means I can easily determine whether the current position, when traversing the simplest target AST, is a definition or a reference (because it allows easy access to parent or sibling nodes, right?

If that's the case, it's simply amazing.

scohen:
Regarding performance. I think we should have a suite of benchmarks that run against the source code to `Enum`, which i've heard is a particularly challenging and long module. Then we won't have to guess.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
scohen: Zippers are cool, yes, but they incur a fairly steep performance penalty over the native methods for AST traversal.

zachallaun: I was thinking about this as well -- we may be able to get away with just having access to ancestors. This still has a lot of benefits and would work, I think, in place of all the zipper usage in this PR, and also would be helpful for @scottming's PR.

apps/common/test/lexical/ast/analysis_test.exs
```diff
@@ -0,0 +1,65 @@
...
+                          ]}
+                       ]},
+                      {{:module, [:Outer, :Inner2]},
+                       [
+                         {{:block, [:Outer, :Inner2]}, []}
+                       ]}
+                    ]}
+                 ]}
+              ]} = code |> analyze() |> Map.fetch!(:tree) |> simplify_tree()
+    end
```
scohen, line 63: there must be a better way to test this

apps/remote_control/lib/lexical/remote_control/search/indexer/extractor.ex
```diff
@@ -0,0 +1,106 @@
...
+  def index(%Analysis{valid?: false}) do
+    {:ok, []}
+  end
+
+  @doc """
+  Records an indexed entry.
+  """
+  def record_entry(
+        %__MODULE__{} = extractor,
+        %Zipper{node: node},
```
scohen, line 51: pulling out the node here seems supurfluous.

scohen (commented):
I'm really concerned about moving over to zippers. I talked to Dorgan about them and even he said they were between 2 and 4x slower than using the plain AST traversal functions.

scohen:
Mine went from 1.7 seconds to 2.7 seconds

scohen:
2.6 seconds

zachallaun:
> 2.6 seconds

Out of curiosity, are you running it twice? The JIT (I think) makes a pretty significant difference for me and the second run in the same IEx session is usually about a second faster both on main and this branch. I want to make sure we're comparing apples to apples.

scottming:
1.4s

zachallaun:
> I wanted to ask if, since we traverse the same AST in the extractor stage, whether expensive operations like `get_range` could be cached?

A quick-and-dirty test of this using the process cache unfortunately made things a bit slower. I'm going to wait on Steve's PR to land before I come back to this, but I think I need to figure out where the extra calls are coming from and rework the strategy.

zachallaun:
`main` has diverged significantly since this branch was created; at this point, any further refactoring will be better done from scratch, potentially using ideas from this PR as a reference.

PR #508, opened 2023-11-30
Added extractor for function references, improved test cases

scohen:
Added an extractor that scans the code for any function references, and indexes them.

Also created an extractor case that has functions that help make testing extraction of various things easier.


scohen:
Note: this really slows down indexing again. Running an index against lexical takes around 12 seconds and produces 200,000 index entries.

@scottming I fixed a couple of the issues you raised in the other PR in this branch.

apps/common/lib/lexical/ast/analysis/analyzer.ex
```diff
@@ -26,26 +26,211 @@ defmodule Lexical.Ast.Analysis.Analyzer do
...
+
+    defp function_and_arity_to_mfa(current_module, fa_list) when is_list(fa_list) do
+      for {function, arity} <- fa_list do
+        {current_module, function, arity}
+      end
+      |> MapSet.new()
+    end
+
+    defp mfas_for(current_module, type) do
+      if Code.ensure_loaded?(current_module) do
```
scottming, line 155: This function seems to have a high cost. I tried simplifying it by removing half of the `Code.ensure_loaded?` calls, and it reduced the indexing time from around 9 seconds to 6 seconds.

https://github.com/lexical-lsp/lexical/compare/function-references...scottming:lexical:optimize-code-load?expand=1

scottming: If we change Code.ensure_loaded? in my branch to Code.load?, the result might surprise you as the time could be reduced to 2 seconds.
```elixir
Time: 1.9 seconds
```

However, there are not as many modules that cannot be loaded by this piece of code:
https://github.com/scottming/lexical/commit/b5f5564d671c9d1a224e0461f364b894c997a6b9#diff-3745d0615d31f6563c8a33a36e1f58bea8539aa6e594fad72eb0d7813cae3fc0R9
 as imagined. Only these modules:
 
 ```elixir
 [(common 0.3.0) lib/lexical/ast/analysis/analyzer.ex:57: Lexical.Ast.Analysis.Analyzer.Import.apply_to_scope/3]
put_unimported_module(import_module) #=> MapSet.new([Bitwise, ExUnit.Assertions, ExUnit.CaptureIO, ExUnit.CaptureLog,
 Kernel.SpecialForms, Lexical.Test.Entry.Builder, Lexical.Test.Fixtures,
 Lexical.Test.Protocol.Fixtures.LspProtocol, Mix.Generator, MyImports,
 Phoenix.Component, Phoenix.Component.Declarative, Phoenix.Controller,
 Phoenix.Controller.Pipeline, Phoenix.HTML, Phoenix.HTML.Tag, Phoenix.LiveView,
 Phoenix.Logger, Phoenix.Socket, Plug.Builder, Plug.Conn, Statistex,
 Supervisor.Spec, TestHelper])
 ```

So, I think we need to find a way to pre-load these modules instead of doing it inside `push_scope`.

scohen: I've pushed a change that caches the results of `ensure_loaded` and reduces the indexing time to 4.4 seconds.

apps/common/lib/lexical/ast/analysis/analyzer.ex
```diff
@@ -26,26 +26,211 @@ defmodule Lexical.Ast.Analysis.Analyzer do
...
+    end
 
-    @type t :: %Scope{}
+    def new(module, selector, line) do
+      %__MODULE__{module: module, selector: expand_selector(selector), line: line}
+    end
 
-    def new(id, %Range{} = range, parent_aliases \\ %{}, module \\ []) do
-      %Scope{id: id, range: range, module: module, parent_aliases: parent_aliases}
+    def apply_to_scope(%__MODULE__{} = import, current_scope, %MapSet{} = current_imports) do
```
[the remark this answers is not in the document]
scohen: And how! Adding a cache around `Code.ensure_loaded` reduces indexing time from 13 seconds to 4.5 seconds.

zachallaun:
I have some architectural concerns related to the module loader and querying module MFAs during analysis, as well as some suggestions that might address it.

My two main concerns are:

1. `Lexical.Module.Loader` will be started twice, both on the server node and in the remote node. However, *it is only the results from the remote node that are ever valid.* I think this means that any analysis generated in the server node will be invalid because user code will not be available, unless I'm missing something. Minimally, it seems really undesirable to have a named process on the server node that should never be used. (Other than during analysis, the loader is only ever used explicitly in `remote_control`, which lends some weight to this.)
3. Prior to this, analysis was essentially a pure function of input -> output. Now it relies on implicit, global state (the runtime code server).

I'd propose the following changes:

1. Analysis of the *AST* should be decoupled from analysis of the *runtime environment*, and `Lexical.Ast.Analysis` should only be concerned with information that can be extracted from the AST without reference to any global state.
2. Analysis should no longer store MFAs, but instead just the imported modules and options like `only: :functions`, `except: [foo: 1]`, etc. A separate API would then be used to resolve an unqualified call at a given position: it would get the imported modules/options from the analysis at position, and then iterate through the modules, using the module loader to get MFAs until it finds the right one (or doesn't).
3. The module loader should move to `remote_control` and, likely, the above described API should be a part of `RemoteControl.Entity`.

scottming:
> Analysis of the AST should be decoupled from analysis of the runtime environment, and Lexical.Ast.Analysis should only be concerned with information that can be extracted from the AST without reference to any global state.

I somewhat agree with Zach's viewpoint. In the current architecture, it is indeed challenging for me to distinguish what tasks should be done in the analysis phase and what tasks should be done in the extractor phase.

If we maintain the purity of the analysis phase, it is highly likely to make each stage clearer for me.

scohen:
I talked with Zack a bit about this PR, and we had a grand strategy that I tried to implement but failed. 

We would _like_ to separate AST work from analysis, but the current code in AST is far too intertwined to pull out cleanly in this PR. AST calls analysis, and the analyzer calls AST. The analysis struct module has functions in it that resolve aliases and imports, and the analysis struct is created by the document store. I don't think there's a way to clean this up without a serious refactor of AST and analysis. It's going to be especially tricky, because a bunch of the fields in the analysis struct directly affect performance. 

I can fix the problem with the loader starting on several nodes easily, but the other stuff is going to be much, much harder. 

> it is indeed challenging for me to distinguish what tasks should be done in the analysis phase and what tasks should be done in the extractor phase

This shouldn't be a problem. Extractors read source code and pull out Entry structs. That is all they do. If they need something else to be done, that needs to happen in code analysis.

scottming:
> I can fix the problem with the loader starting on several nodes easily

In that case, I think we should first address this issue and then move forward, allowing Zach to also advance with the ancestor logic PR. I believe that part is crucial, as without it, all the extraction(like variables) work dependent on the ancestor scope would be challenging to proceed.

scohen:
I did a refactor that gets us part of the way there:
✅  Analysis of the AST should be decoupled from analysis of the runtime environment
✅  Analysis should no longer store MFAs, but instead just the imported modules and options 
✅  The module loader should move to remote_control
❌  above described API should be a part of RemoteControl.Entity

The last part is not done, I want to move the analyzer to remote control entirely, and stick the stuff that's in `analyzer.ex` into `analysis.ex`.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/function_reference.ex
```diff
@@ -0,0 +1,189 @@
...
+                          |> Kernel.__info__()
+                          |> Enum.reduce([], fn {name, _arity}, acc ->
+                            string_name = Atom.to_string(name)
+
+                            if String.starts_with?(string_name, "def") do
+                              [name | acc]
+                            else
+                              acc
+                            end
+                          end)
```
zachallaun, line 20:
```suggestion
                          |> Enum.map(fn {name, _arity} -> Atom.to_string(name) end)
                          |> Enum.filter(&String.starts_with?(&1, "def"))
```

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/function_reference.ex
zachallaun, line 1: Since the module attributes are going to run at compile-time, it's possible that the set of functions does not match the runtime Elixir version. Perhaps we should resolve at runtime and cache with `:persistent_term` instead?

zachallaun (approved):

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/function_reference.ex
scohen: yeah, good point.

apps/remote_control/lib/lexical/remote_control/analyzer/aliases.ex
```diff
@@ -0,0 +1,58 @@
...
+          %{^prefix => _} ->
+            current_module =
+              aliases
+              |> Map.get(prefix)
+              |> Alias.to_module()
+
+            Module.concat([current_module | suffix])
+
+          _ ->
+            Module.concat(module)
```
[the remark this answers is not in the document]
scohen: I'm not sure how to do what we want then.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/function_reference.ex
scohen: I liked this a little more 
```elixir
      for macro_name <- Kernel.__info__(:macros),
          string_name = Atom.to_string(macro_name),
          not String.starts_with?(string_name, "def") do
        macro_name
      end
```

PR #512, opened 2023-12-10
Fix the issue of project.log not displaying logs when developing Lexical with version 1.15.

scottming:
In `test.exs`, `Logger.configure` is called when the project node starts, which causes log information not to be displayed after Elixir version 1.15. If we only want to ignore excessive logs during test execution, it would be more appropriate to handle this in `test_helpers`.

### DIFF
--- a/apps/remote_control/test/test_helper.exs
+++ b/apps/remote_control/test/test_helper.exs
@@ -6,3 +6,9 @@ with :nonode@nohost <- Node.self() do
 end
 
 ExUnit.start(exclude: [:skip])
+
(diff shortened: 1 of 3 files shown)


scohen:
Man i wish there was another way to solve this issue. This is an verbose and hidden fix.

scohen (approved):

scottming:
> Man i wish there was another way to solve this issue. This is an verbose and hidden fix.

Yes, I have also tried making some modifications in the Bootstrap module, but it didn't work.

PR #514, opened 2023-12-12
Handle compilation exiting

### LINKED ISSUE #504: Lexical Build crashes if the compiler exits
steps: 

1. create a file in a project with the below content:
```elixir
defmodule Module do
  @after_verify {__MODULE__, :verify}
  # what happens if I do this
  def verify(_) do
    raise "whoops"
  end
end
```
Note: the above should give you a compile error similar to the below:
```shell
Compiling 21 files (.ex)
warning: redefining module Module (current version loaded from /Users/zimakki/.asdf/installs/elixir/1.15.4-otp-26/bin/../lib/elixir/ebin/Elixir.Module.beam)
  lib/testing.ex:1: Module (module)

** (UndefinedFunctionError) function Module.get_attribute/2 is undefined or private
    (elixir 1.15.4) Module.get_attribute(Module, :__impl__)
    (mix 1.15.4) lib/mix/compilers/elixir.ex:1172: Mix.Compilers.Elixir.detect_kind/1
    (mix 1.15.4) lib/mix/compilers/elixir.ex:1014: Mix.Compilers.Elixir.compiler_loop/4
    (mix 1.15.4) lib/mix/compilers/elixir.ex:173: Mix.Compilers.Elixir.compile/7
    (mix 1.15.4) lib/mix/state.ex:15: Mix.State.lock/2
    (mix 1.15.4) lib/mix/task.ex:447: anonymous fn/3 in Mix.Task.run_task/5
    (mix 1.15.4) lib/mix/tasks/compile.all.ex:124: Mix.Tasks.Compile.All.run_compiler/2
    (mix 1.15.4) lib/mix/tasks/compile.all.ex:104: Mix.Tasks.Compile.All.compile/4
```

2. Close your editor and open it again... 

Note: lexical will crash and become unresponsive

scohen:
It is possible for compilation to exit, and we were not handling it properly. This was complicated by the fact that builds happen inside a GenServer, so wrapping things in a try/rescue with an exit handler wasn't enough.

Instead, i launched another process that's monitored and awaited by the build genserver. Then any DOWN messages are turned into diagnostics, and reported.

Fixes #504

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/build/document.ex
+++ b/apps/remote_control/lib/lexical/remote_control/build/document.ex
@@ -2,35 +2,22 @@ defmodule Lexical.RemoteControl.Build.Document do
   alias Lexical.Document
   alias Lexical.RemoteControl.Build
   alias Lexical.RemoteControl.Build.Document.Compilers
+  alias Lexical.RemoteControl.Build.Isolation
 
   @compilers [Compilers.Config, Compilers.Elixir, Compilers.EEx, Compilers.HEEx, Compilers.NoOp]
 
   def compile(%Document{} = document) do
     compiler = Enum.find(@compilers, & &1.recognizes?(document))
-    me = self()
+    compile_fun = fn -> compiler.compile(document) end
 
-    {pid, ref} =
-      spawn_monitor(fn ->
-        result = compiler.compile(document)
-        send(me, {:result, result})
-      end)
-
-    receive do
-      {:result, result} ->
-        flush_normal_down(ref, pid)
+    case Isolation.invoke(compile_fun) do
+      {:ok, result} ->
         result
 
-      {:DOWN, ^ref, :process, ^pid, {exception, stack}} ->
+      {:error, {exception, stack}} ->
         diagnostic = Build.Error.error_to_diagnostic(document, exception, stack, nil)
         diagnostics = Build.Error.refine_diagnostics([diagnostic])
         {:error, diagnostics}
-    end
-  end
-
-  defp flush_normal_down(ref, pid) do
-    receive do
-      {:DOWN, ^ref, :process, ^pid, :normal} ->
-        :ok
     end
   end
 end
--- a/apps/remote_control/lib/lexical/remote_control/build/document.ex
+++ b/apps/remote_control/lib/lexical/remote_control/build/document.ex
@@ -1,11 +1,36 @@
 defmodule Lexical.RemoteControl.Build.Document do
   alias Lexical.Document
+  alias Lexical.RemoteControl.Build
   alias Lexical.RemoteControl.Build.Document.Compilers
 
   @compilers [Compilers.Config, Compilers.Elixir, Compilers.EEx, Compilers.HEEx, Compilers.NoOp]
 
   def compile(%Document{} = document) do
     compiler = Enum.find(@compilers, & &1.recognizes?(document))
-    compiler.compile(document)
+    me = self()
+
+    {pid, ref} =
+      spawn_monitor(fn ->
+        result = compiler.compile(document)
+        send(me, {:result, result})
+      end)
+
+    receive do
+      {:result, result} ->
+        flush_normal_down(ref, pid)
+        result
+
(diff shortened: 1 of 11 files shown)


apps/remote_control/lib/lexical/remote_control/build/document.ex
```diff
@@ -1,11 +1,36 @@
...
+        diagnostics = Build.Error.refine_diagnostics([diagnostic])
+        {:error, diagnostics}
+    end
+  end
+
+  defp flush_normal_down(ref, pid) do
+    receive do
+      {:DOWN, ^ref, :process, ^pid, :normal} ->
+        :ok
+    end
```
zachallaun, line 34: We should add an `after` block with a small timeout (50-100ms?) that throws a runtime error if the normal down isn't received. (I think it should always be received, but if it's not, this will block forever.)

apps/remote_control/lib/lexical/remote_control/build/document.ex
```diff
@@ -1,11 +1,36 @@
...
+        result = compiler.compile(document)
+        send(me, {:result, result})
+      end)
+
+    receive do
+      {:result, result} ->
+        flush_normal_down(ref, pid)
+        result
+
+      {:DOWN, ^ref, :process, ^pid, {exception, stack}} ->
```
zachallaun, line 23: Are we sure the reason will always be of the shape `{exception, stack}`?

apps/remote_control/lib/lexical/remote_control/build/project.ex
zachallaun, line 1: Same comments as in the document builder apply to the receive blocks here.

zachallaun: I wonder if they should share a common function that yields `{:ok, result}` or `{:error, reason}`, so both files would become something like:

```elixir
compile_fun = fn -> Mix.Task.run(:compile, mix_compile_opts(force?)) end

case Build.run_monitored(compile_fun) do
  {:ok, result} -> ...
  {:error, reason} -> ...
end
```

zachallaun (approved):

zachallaun:
Obviously look into the test errors but the strategy looks good to me.

apps/remote_control/lib/lexical/remote_control/build/document.ex
scohen: For the condition that I'm trying to solve, yes.

apps/remote_control/lib/lexical/remote_control/build/document.ex
scohen: I went back and forth on this. On one hand, I appreciate somewhat defensive coding, but on the other, if we don't receive a message, a large part of the set of guarantees the erlang runtime system provides are false.

apps/remote_control/lib/lexical/remote_control/build/project.ex
scohen: Already done. Moved to `Build.Isolation.invoke(function)`

apps/remote_control/lib/lexical/remote_control/build/document.ex
scohen: ...but I can't be sure that in the future we won't have some other shape. I was hoping to let it crash and have people report a bug if it does, rather than silently swallowing errors.

zachallaun: It's not going to crash unless we crash it though, right? It'll just receive forever?

scohen: ah, good point, I'll make it fail a pattern match

scohen: Now that I introduced the Isolation module, it'll crash in the callers with a case clause error.

apps/remote_control/lib/lexical/remote_control/build/document.ex
scohen: Note, we moved to `Process.demonitor(ref, [:flush])`.

PR #516, opened 2023-12-12
Find function references

scohen:
Find references: Functions

Implemented find references for functions.
Initially, very few references were found, and this was because we
were deduping the index by id. Removing this call allows us to see the
entire project's references, with the notable exception of imported
macros. I believe we'll need to augment the extractor with a
compilation tracer if we wish to see those. The bad news is lexical
takes around 12 seconds to compile, but around 3 seconds to index.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/code_intelligence/references.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_intelligence/references.ex
@@ -3,6 +3,7 @@ defmodule Lexical.RemoteControl.CodeIntelligence.References do
   alias Lexical.Document
   alias Lexical.Document.Location
   alias Lexical.Document.Position
+  alias Lexical.Formats
   alias Lexical.RemoteControl.CodeIntelligence.Entity
   alias Lexical.RemoteControl.Search.Indexer.Entry
   alias Lexical.RemoteControl.Search.Store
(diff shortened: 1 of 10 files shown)


scohen:
Fixed by #508

scottming:
"range" is incorrect, and we shouldn't highlight the `)` character,  Additionally, I think we need a few simple integration tests.

CleanShot 2023-12-13 at 21 13 43@2x

scohen:
Do you mean we should highlight the ) ? Presently it's not highlighted.

scottming:
Additionally, do you think that besides the definition, the places where it’s referenced also need to include the parameters within? Wouldn’t this make it seem like there’s a bit too much noise?

scohen:
I think you brought this up before, and I resisted, I see your point about it being noisy, especially when passing an anonymous function as a parameter. I'd like to use the feature more before we just highlight the name.

scohen:
@scottming I did tone down the definitions to only include the definition up to the `do`.

scottming:
> I think you brought this up before, and I resisted, I see your point about it being noisy, 

At that time, you mentioned that after the function definition includes do, it's better handled as a whole using Fragment-related AST functions. Additionally, considering that each navigation has different requirements for the range, I have no objections to having more information at the indexer stage.

For example, in `DocumentSymbol`, there's a `detail` field; if it includes information about parameters, then the effect is quite good. This is the result in Lua:

CleanShot 2023-12-14 at 21 26 14@2x

> especially when passing an anonymous function as a parameter. 

Yes, it's not just with anonymous functions, the issue also arises when the argument is a list or contains a structure.

> I'd like to use the feature more before I decide if we should just highlight the name.

Sure, I respect your opinion.

scottming (approved):

PR #519, opened 2023-12-14
Added percentage based progress reporters

scohen:
I thought it would be nicer to have the indexer report its progress in terms of percentage completion rather than jamming a bunch of file names in your face.

This change introduces a different funtion for percentage based progress reporting in both remote_control and server. You initialize it with the maximum value for th progress, and then update progress via a updater function that's passed into the callback you give to `with_percent_progress`.


scohen:
@zachallaun @scottming  ping!

scottming (approved):
LGTM

image

PR #522, opened 2023-12-15
Added reindex command

### LINKED ISSUE #517: Reindex command
Lexical should have a lsp command to reindex a project

scohen:
Added a reindex code lens so we can rebuild indexes without resorting to manual operations.

As we build the indexing infrastructure out, we'll likely need to rebuild the index a lot. Presently, this means restarting the server, which can take some time, so I thought i'd add a command to do it instead. I tried using code actions, but this was a bit fraught, so instead, I used a code lens on the project definition in your mix.exs file.

Fixes #517

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/commands/reindex.ex
+++ b/apps/remote_control/lib/lexical/remote_control/commands/reindex.ex
@@ -66,11 +66,15 @@ defmodule Lexical.RemoteControl.Commands.Reindex do
   @moduledoc """
   A simple genserver that prevents more than one reindexing job from running at the same time
   """
+
   alias Lexical.Document
   alias Lexical.Project
+  alias Lexical.RemoteControl.Api
+  alias Lexical.RemoteControl.Dispatch
   alias Lexical.RemoteControl.Search
 
   use GenServer
+  import Api.Messages
 
   def start_link(reindex_fun) when is_function(reindex_fun, 1) do
     GenServer.start_link(__MODULE__, reindex_fun, name: __MODULE__)
@@ -127,8 +131,19 @@ defmodule Lexical.RemoteControl.Commands.Reindex do
   end
 
   defp do_reindex(%Project{} = project) do
-    with {:ok, entries} <- Search.Indexer.create_index(project) do
-      Search.Store.replace(entries)
-    end
+    Dispatch.broadcast(project_reindex_requested(project: project))
+
+    {elapsed_us, result} =
+      :timer.tc(fn ->
+        with {:ok, entries} <- Search.Indexer.create_index(project) do
+          Search.Store.replace(entries)
+        end
+      end)
+
+    Dispatch.broadcast(
+      project_reindexed(project: project, elapsed_ms: round(elapsed_us / 1000), status: :success)
+    )
+
+    result
   end
 end
--- a/apps/remote_control/lib/lexical/remote_control/commands/reindex.ex
+++ b/apps/remote_control/lib/lexical/remote_control/commands/reindex.ex
@@ -35,9 +35,7 @@ defmodule Lexical.RemoteControl.Commands.Reindex do
     def reindex_uri(%__MODULE__{} = state, uri) do
       case entries_for_uri(uri) do
         {:ok, path, entries} ->
-          new_pending_updates = Map.put(state.pending_updates, path, entries)
-
-          %__MODULE__{state | pending_updates: new_pending_updates}
+          put_in(state.pending_updates[path], entries)
 
         _ ->
           state
--- a/apps/remote_control/lib/lexical/remote_control/commands/reindex.ex
+++ b/apps/remote_control/lib/lexical/remote_control/commands/reindex.ex
@@ -1,0 +1,136 @@
+defmodule Lexical.RemoteControl.Commands.Reindex do
+  defmodule State do
+    alias Lexical.Ast.Analysis
+    alias Lexical.Document
+    alias Lexical.RemoteControl.Search
+    alias Lexical.RemoteControl.Search.Indexer
+
+    require Logger
+    defstruct reindex_fun: nil, index_task: nil, pending_updates: %{}
+
+    def new(reindex_fun) do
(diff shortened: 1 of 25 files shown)


apps/server/lib/lexical/server/provider/handlers/commands.ex
```diff
@@ -0,0 +1,65 @@
...
+      {elapsed, :ok} ->
+        message = "Reindexed #{Project.name(project)} in #{Formats.time(elapsed)}"
+        Window.show(:info, message)
+        Responses.ExecuteCommand.new(request_id, "ok")
+
+      {_elapsed, error} ->
+        Window.error("Indexing #{Project.name(project)} failed")
+        Logger.error("Indexing command failed due to #{inspect(error)}")
+
+        internal_error(request_id, "Indexing Failed")
```
scottming, line 54: I've noticed that when I click too quickly, it fails here. I believe that `{:error, "Already Running"}` should be set to info: `"Already running"` instead of the current error failed message. It would result in a better user experience.

CleanShot 2023-12-16 at 15 49 31@2x

zachallaun:
Is there a reason not to use the LSP's Execute Command functionality? We can share known commands with server capabilities on initialize. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_executeCommand

zachallaun:
My apologies: I see we are doing this. The code lens is just a convenient interface to trigger it.

apps/remote_control/lib/lexical/remote_control/commands/reindex.ex
```diff
@@ -0,0 +1,136 @@
...
+
+      state
+    end
+
+    def reindex_uri(%__MODULE__{} = state, uri) do
+      case entries_for_uri(uri) do
+        {:ok, path, entries} ->
+          new_pending_updates = Map.put(state.pending_updates, path, entries)
+
+          %__MODULE__{state | pending_updates: new_pending_updates}
```
zachallaun, line 40:
```suggestion
          put_in(state.pending_updates[path], entries)
```

apps/server/lib/lexical/server/provider/handlers/commands.ex
zachallaun: While I agree with this, I'd probably go even further and say that another reindex command should just cancel the current one and restart it.

apps/server/lib/lexical/server/provider/handlers/commands.ex
```diff
@@ -0,0 +1,65 @@
...
+    end
+  end
+
+  defp internal_error(request_id, message) do
+    Responses.ExecuteCommand.error(request_id, :internal_error, message)
+  end
+
+  defp reindex_name do
+    "Reindex"
```
zachallaun, line 63: This is hard-coded elsewhere as well. IMO we should just use `@reindex_name "Reindex"` at the top and then it can also be used in pattern-matching, etc. No need for this function.

apps/server/lib/lexical/server/provider/handlers/code_lens.ex
```diff
@@ -0,0 +1,50 @@
...
+        nil -> []
+        lens -> List.wrap(lens)
+      end
+
+    response = Responses.CodeLens.new(request.id, lenses)
+    {:reply, response}
+  end
+
+  defp reindex_lens(%Project{} = project, %Document{} = document) do
+    if Path.basename(document.path) == "mix.exs" do
```
zachallaun, line 27: Should this also check that indexing is enabled?

zachallaun (commented):
I can't test this until later, but made some comments for now.

scohen:
> My apologies: I see we are doing this. The code lens is just a convenient interface to trigger it.

my understanding is that the only two ways to use that functionality are code actions and code lenses.

apps/server/lib/lexical/server/provider/handlers/commands.ex
scohen: Why not let the first one finish?

apps/server/lib/lexical/server/provider/handlers/code_lens.ex
scohen: Good point

apps/server/lib/lexical/server/provider/handlers/commands.ex
scohen: That’s a mistake

apps/server/lib/lexical/server/provider/handlers/commands.ex
scohen: Cancelling the task will mean that the index could be in an inconsistent state. Letting the first task finish ensures that the index will always exist, and that it is correct.

scohen: I was considering triggering a recompute of code lenses when indexing starts and stops and not emitting this lens if there's an index job running.

zachallaun: > Cancelling the task will mean that the index could be in an inconsistent state. Letting the first task finish ensures that the index will always exist, and that it is correct.

Oh, I thought all entries were written together at the same time.

My assumption is that, since this is a manually-triggered command, if you're triggering it again while an index is in-process, you're doing it on purpose and the result of the in-process one isn't useful to you (e.g. you changed something and want it reflected in the index).

scohen: > Oh, I thought all entries were written together at the same time.

That's true, but if you exit the process while the write is happening, you will lose data (this is surprisingly easy to do).

scohen: > e.g. you changed something and want it reflected in the index

this is not a use case. Changing something updates the index automatically. The use case for this button is "I've upgraded lexical, and I need to reindex my source code to get some new feature". Mainly, this is for us.

scottming (approved):

zachallaun (approved):

PR #526, opened 2023-12-23
Workaround for indexing performance

scohen:
Indexing slows down project compilation by 4x, but I've been completely unable to figure out why. This PR works around the issue by waiting for the project to compile before indexing is enabled. I'll explain what I know next.

All I had to go on was the following: The code server had a message queue and the dirty schedulers had elevated levels. The odd thing is that these two things happened _after_ the indexer had completed. Whatever the indexer was doing continued to affect the compiler, even after the indexer was done.

I investigated the following hypothesis

 * We were overwhelming the file server with traffic x I disabled any reads and writes, still had the problem
 * We were contending on the `:global` locks x I removed the global registrations, the problem persisted
 * Indexing was exhausting the CPU, so compilation couldn't complete x I set the max concurrnecy of the indexer to 1, it still persisted

I then went and disabled more and more of the indexer and found the following, which was quite surprizing: Loading the ETS table slowed down compilation. To reiterate, the act of setting up the ETS table affected compilation, even after the ets table was set up.

Once I realized this, and that we were having contention in the code server, I tried the following:

 * Eliminating any module references in the index. This had a nice effect on the index size, but didn't afect compilation performance
 * I also avoided any calls to the code server in the indexing path, this also had no effect.

So this is a bit of a stumper for me. I have another branch with some changes to the iex helpers that make debugging this easier if anyone else wants to take a crack at it.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/store.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/store.ex
@@ -4,6 +4,8 @@ defmodule Lexical.RemoteControl.Search.Store do
   """
 
   alias Lexical.Project
+  alias Lexical.RemoteControl.Api
+  alias Lexical.RemoteControl.Dispatch
   alias Lexical.RemoteControl.Search.Indexer.Entry
   alias Lexical.RemoteControl.Search.Store
   alias Lexical.RemoteControl.Search.Store.State
@@ -34,11 +36,11 @@ defmodule Lexical.RemoteControl.Search.Store do
                        2500
                      )
 
+  import Api.Messages
   use GenServer
   require Logger
 
   def stop do
-    #    GenServer.call(__MODULE__, :on_stop)
     GenServer.stop(__MODULE__)
   end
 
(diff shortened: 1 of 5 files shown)


apps/remote_control/lib/lexical/remote_control/search/store.ex
```diff
@@ -175,6 +202,11 @@ defmodule Lexical.RemoteControl.Search.Store do
     {:reply, :ok, {ref, new_state}}
   end
 
+  def handle_call(message, _from, %State{} = state) do
+    Logger.emergency("Received #{inspect(message)}, but the search store isn't enabled yet.")
```
zachallaun, line 206: Semantically, I think warning is sufficient here. I think of emergency as something that requires immediate intervention, but there's nothing really to be done but to wait.

zachallaun (approved):
This seems fine as a stop-gap. I'm definitely curious to figure out what's really going on, though!

apps/remote_control/lib/lexical/remote_control/search/store.ex
scohen: That's fair, I actually can think of a couple of ways to trigger this, so emergency is a bit much.

scohen:
I'm going to write a summary of what I know, what I tried, and what triggers the problem. I think we're going to need to get pretty deep in the compiler or in erlang VM internals to figure out what's going on.

PR #530, opened 2023-12-28
Properly handle expand_alias failures

scohen:
I initially assumed that the cases where `expand_alias` would fail would be rare, so I did a hard match on `{:ok, alias}`. I was wrong, the cases are much more common, and this is causing irritation for people trying to use indexing. This change logs the cases where expand_alias fails, and skips the entries, which should give better developer experience and should also allow us to see what's causing the failures.


scohen:
@zachallaun @scottming I'm going to merge this, it's causing people issues.

PR #534, opened 2023-12-29
Improved lexical's index efficiency

scohen:
For lexical, the index was using 450MB of disk and consumed about a gig in memory. This commit includes the following changes:

  * We stopped duplicating entries in a couple of the records, and now only store them once in the 'by_ref' key. This decreased the index size by 50%, and had no noticable effect on performance.
  * Several of the processes had cache entries lying around that caused them to consume quite a bit (hundreds of megs) of memory. I introduced a cache cleaning block and added it to the call sites, and this reduced the memory usage of those processes back to extremely low levels.
  * We were storing the erlang and elixir versions on each entry. Now we create separate indexes for each elixir/erlang version pair

In total, lexical was taking around 1gb of memory before these changes, and now it takes around 400mb. Disk space dropped from 450 megs to 250.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/store/backends/ets/state.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/store/backends/ets/state.ex
@@ -8,7 +8,6 @@ defmodule Lexical.RemoteControl.Search.Store.Backends.Ets.State do
   alias Lexical.Project
   alias Lexical.RemoteControl.Search.Store.Backends.Ets.Schema
   alias Lexical.RemoteControl.Search.Store.Backends.Ets.Schemas
-  alias Lexical.VM.Versions
 
   @schema_order [
     Schemas.LegacyV0,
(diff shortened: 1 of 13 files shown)


apps/remote_control/lib/lexical/remote_control/search/store/backends/ets/state.ex
```diff
@@ -60,28 +59,26 @@ defmodule Lexical.RemoteControl.Search.Store.Backends.Ets.State do
...
       )
 
     state.table_name
     |> :ets.match_object({match_pattern, :_})
-    |> Enum.flat_map(fn {_, match} -> match end)
+    |> Enum.flat_map(fn {_, id_keys} ->
+      id_keys
+    end)
+    |> MapSet.new()
+    |> Enum.flat_map(&:ets.lookup_element(state.table_name, &1, 2))
```
scottming, line 75: LGTM

scottming (approved):

PR #535, opened 2024-01-01
Support Elixir 1.16

### LINKED ISSUE #513: Support Elixir 1.16
Elixir 1.16 has two main changes: 

1. the addition of the `span` field, which means diagnostics can be located more precisely; We need to change the diagnostics construction way that is generated by `with_diagnostics.`
2. secondly, in the event of an error in parsing the AST, we need to use certain fields from 'meta' to construct two diagnostics, rather than parsing the message info directly.

scottming:
In version 1.16, both parseError and diagnostics provided by the Elixir compiler have become more precise in terms of their locations.

Therefore, I believe it is a good time to separate the `parseError` and `diagnostic error` modules. For `parseError`, we can always provide an accurate range for the starting part(Because the `context` field contains more detailed information). For certain diagnostic errors, we can also use the `span` field to make them more precise.

### ParseError

For example, in this case, on the same line, we can indicate two specific locations with underlines. I believe this approach will result in a better user experience, allowing users to quickly identify the specific issues.

CleanShot 2024-01-01 at 12 36 40@2x

CleanShot 2024-01-01 at 12 38 36@2x

CleanShot 2024-01-01 at 12 38 30@2x

### Span

CleanShot 2024-01-01 at 17 40 04@2x
CleanShot 2024-01-01 at 17 41 32@2x
CleanShot 2024-01-01 at 17 41 59@2x

Fixes #513


apps/remote_control/lib/lexical/remote_control/build/parse_error.ex
```diff
@@ -0,0 +1,176 @@
+defmodule Lexical.RemoteControl.Build.ParseError do
+  alias Lexical.Document
+  alias Lexical.Plugin.V1.Diagnostic.Result
+
+  import Lexical.RemoteControl.Build.ErrorSupport,
```
scohen, line 5: Not a fan of importing normal functions

apps/remote_control/lib/lexical/remote_control/build/error_support.ex
```diff
@@ -0,0 +1,30 @@
+defmodule Lexical.RemoteControl.Build.ErrorSupport do
```
scohen, line 1:
```suggestion
defmodule Lexical.RemoteControl.Build.Error.Location do 
```

apps/remote_control/lib/lexical/remote_control/build/error_support.ex
```diff
@@ -0,0 +1,30 @@
...
+  def context_to_position(context) do
```
scohen, line 2: Is this coming from metadata? Context is such an unhelpful word

apps/remote_control/lib/lexical/remote_control/build/error_support.ex
```diff
@@ -0,0 +1,30 @@
...
+    cond do
+      Keyword.has_key?(context, :line) and Keyword.has_key?(context, :column) ->
```
scohen, line 4: Pull out line and column as a tuple, then pattern match in the function heads. I think that's be cleaner.

apps/remote_control/lib/lexical/remote_control/build/parse_error.ex
```diff
@@ -0,0 +1,176 @@
...
+  end
+
+  def to_diagnostics(%Document{} = source, context, message_info, token)
+      when is_exception(message_info) do
+    to_diagnostics(source, context, Exception.message(message_info), token)
+  end
+
+  def to_diagnostics(%Document{} = source, context, message_info, token) do
+    end_line_fn =
+      if Features.details_in_context?() do
```
scohen, line 33: These two ifs might be better as a cond

apps/remote_control/lib/lexical/remote_control/build/parse_error.ex
```diff
@@ -0,0 +1,176 @@
...
+         token
+       ) do
+    message =
+      if String.starts_with?(message_info, "unexpected") do
+        ~s/#{message_info}#{token}, expected "#{context[:expected_delimiter]}"/
+      else
+        "#{message_info}#{token}"
+      end
+
+    if context[:end_line] do
```
scohen, line 68: I feel like this should be happening in the location code

apps/remote_control/test/lexical/remote_control/build/error_test.exs
```diff
@@ -452,14 +383,24 @@ defmodule Lexical.RemoteControl.Build.ErrorTest do
...
-      assert diagnostic.message =~ "unknown key :c for struct Foo"
+      if Features.span_in_diagnostic?() do
+        [undefined, unknown] = diagnostics
+        assert unknown.message == "unknown key :c for struct Foo"
+        assert unknown.position == {7, 19}
 
-      if Features.with_diagnostics?() do
-        assert diagnostic.position == {7, 19}
+        assert undefined.message == "variable \"c\" is unused"
+        assert undefined.position == {7, 27, 7, 28}
```
scohen, line 394: Don't do assertions like this. Instead assert by extracting ranges and comparing against strings.

apps/remote_control/test/lexical/remote_control/build/error_test.exs
```diff
@@ -452,14 +383,24 @@ defmodule Lexical.RemoteControl.Build.ErrorTest do
...
       else
-        assert diagnostic.position == 7
+        [diagnostic] = diagnostics
+        assert diagnostic.message =~ "unknown key :c for struct Foo"
+
+        if Features.with_diagnostics?() do
```
scohen, line 399: This should likely go around test cases rather than assertions. I.e. have separate cases for the feature being enabled and disabled.

apps/remote_control/test/lexical/remote_control/build/parse_error_test.exs
```diff
@@ -0,0 +1,183 @@
+defmodule Lexical.RemoteControl.Build.ParseErrorTest do
```
scohen, line 1: Fo the entire case, don't assert against tuples of numbers. Those are super hard to reason about. We have RangeSupport that highlights ranges. Use that instead

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -96,10 +96,10 @@ defmodule Lexical.RemoteControl.CodeIntelligence.DefinitionTest do
     test "find the macro definition", %{project: project, uri: referenced_uri} do
       subject_module = ~q[
         defmodule UsesRemoteFunction do
-          alias MyDefinition
+          require MyDefinition
 
           def uses_macro() do
-            MyDefinition.print_hello|()
+            MyDefinition.print_hel|lo()
```
scohen, line 102: Why'd you move the cursor?

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -108,34 +108,13 @@ defmodule Lexical.RemoteControl.CodeIntelligence.DefinitionTest do
...
-    """
-    @tag :skip
     test "find the right arity function definition", %{project: project} do
       subject_module = ~q[
         defmodule UsesRemoteFunction do
           alias MultiArity
 
           def uses_multiple_arity_fun() do
-            MultiArity.sum|(1, 2, 3)
+            MultiArity.su|m(1, 2, 3)
```
scohen, line 117: Why'd you move the cursor?

apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
```diff
@@ -124,7 +124,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.ModuleOrBehavi
 
     local_name =
       child_module
-      |> String.slice(local_module_length..-1)
+      |> String.slice(local_module_length..-1//1)
```
scohen, line 127: Did range behaviour change in 1.16?

apps/remote_control/lib/lexical/remote_control/build/error_support.ex
scohen:
```elixir
case {context[:line], context[:column], context[:end_line], context[:end_collumn]} do 
  {nil, nil, nil, nil} -> nil # I don't like returning nil here. This leaves us without a place to put the error
  {line, nil, nil, nil} -> line
  {line, column, nil, nil} -> {line, column}
  range -> range
end
```

I kind of wonder if we should just be returning Positions and Ranges here.

apps/remote_control/lib/lexical/remote_control/build/error_support.ex
```diff
@@ -0,0 +1,30 @@
...
+    {line, column, end_line, end_column}
+  end
+
+  def position(line, column) do
+    {line, column}
+  end
+
+  def position(line, column, end_line, end_column) do
+    {line, column, end_line, end_column}
+  end
```
scohen, line 29: these need to go. Elsewhere in the code, you're pulling apart the tuples and calling these functions that put them together again.

apps/remote_control/lib/lexical/remote_control/build/parse_error.ex
scohen:
```elixir
{start_line_fun, end_line_fun} = 
  if Features.details_in_context?() do 
    {&build_end_line_diagnostics_from_context/4, &build_end_line_diagnostics_from_context}
  else 
    {&build_start_line_diagnostics/4, &build_end_line_diagnostics/4}
  end
```

apps/remote_control/lib/lexical/remote_control/build/error_support.ex
scohen: Support is a term we use exclusively in tests, I'd like to keep it that way.

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -32,6 +34,28 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Definition do
     {:ok, nil}
   end
 
+  defp maybe_move_cursor_to_next(type, document, line, column)
```
scohen, line 37: naming: What does this function do? What is next?

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -32,6 +34,28 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Definition do
...
+    case Ast.zipper_at(document, position) do
+      {:ok, zipper} ->
+        zipper = %{node: {entity_name, meta, _}} = Sourceror.Zipper.next(zipper)
+
+        if entity_name == :when do
+          %{node: {_entity_name, meta, _}} = Sourceror.Zipper.next(zipper)
+          {meta[:line], meta[:column]}
+        else
+          {meta[:line], meta[:column]}
+        end
```
scohen, line 50:
```suggestion
    meta = 
        if entity_name == :when do
          %{node: {_entity_name, meta, _}} = Sourceror.Zipper.next(zipper)
          meta
        else
          meta
        end
    {meta[:line], meta[:column]}
```
Though why don't you use the helper you defined to extract the position from the metadata?

apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
scottming: yes.

> [Enum] Deprecate passing a range with negative step on Enum.slice/2, give first..last//1 instead

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
scottming: move the cursor from

```elixir
   |def func() do
```

to 

```elixir
  def |func() do
```

ElixirSense changed its default behavior.

Do you think `maybe_move_cursor_to_next_node` is better?

scottming: Compared to using a zipper, do you think it would be more appropriate for us to define a `suffix_tokens` in `Lexical.Ast.Env` for cursor movement?

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
scohen: ...annoying.

apps/server/lib/lexical/server/code_intelligence/completion/translations/module_or_behaviour.ex
scohen: wow, that's a usability regression. Bummer.

scohen:
@scottming would it be easier to make two PRs, one that gets the current version working in 1.16 and another that does the diagnostic work?

scottming:
> @scottming would it be easier to make two PRs, one that gets the current version working in 1.16 and another that does the diagnostic work?

I don't think that's easy.

Because before 1.16, many errors were parsed from the message info, but after 1.16, the message info no longer contains that information.

Therefore, if we simply submit a PR with a simple modification to `@allowed_elixir` in boot.ex, it would result in the loss of a significant number of diagnostics during compilation. This would make the **"as-you-type"** feature almost unusable.

apps/remote_control/lib/lexical/remote_control/build/error_support.ex
scottming: I don't quite understand this comment. Could you provide more details about it?

apps/remote_control/lib/lexical/remote_control/build/error_support.ex
scottming: it is from stack_context or [string_to_quoted](https://github.com/elixir-lang/elixir/blob/927b10df80ee1c1c7396e68efe00d06bc3e80420/lib/elixir/lib/code.ex#L1147) 's `meta`

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
scottming: Applied.

> Though why don't you use the helper you defined to extract the position from the metadata?

Because this is the definition module, there is only one place where the position is extracted.

apps/remote_control/lib/lexical/remote_control/build/parse_error.ex
scottming: And any naming suggestions?

apps/remote_control/lib/lexical/remote_control/build/parse_error.ex
scottming: I'm curious How do you typically differentiate between normal functions and abnormal functions?

scottming:
I think it is almost ready.

cc @scohen @zachallaun

apps/remote_control/lib/lexical/remote_control/analyzer/imports.ex
```diff
@@ -110,11 +114,21 @@ defmodule Lexical.RemoteControl.Analyzer.Imports do
 
   defp function_and_arities_for_module(module, type) do
     ProcessCache.trans({module, :info, type}, fn ->
-      type
-      |> module.__info__()
-      |> Enum.reject(fn {name, _arity} ->
-        name |> Atom.to_string() |> String.starts_with?("_")
-      end)
+      if type == :sigils do
```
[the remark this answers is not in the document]
scottming: great, I rebased the main branch.

apps/remote_control/lib/lexical/remote_control/build/parse_error.ex
```diff
@@ -0,0 +1,189 @@
+defmodule Lexical.RemoteControl.Build.ParseError do
```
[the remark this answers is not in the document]
scottming: Considering that `ParseError` and `Error` are used at the same level, I think both `ParseError` and `Error.Parse` are acceptable. I don't have a strong preference.

apps/remote_control/lib/lexical/remote_control/build/error/location.ex
```diff
@@ -0,0 +1,71 @@
+defmodule Lexical.RemoteControl.Build.Error.Location do
```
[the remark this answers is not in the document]
scottming: I can change position to return `Position`, and then rename the function that returns `Range` to `range()`. However, what should we name it when the result is a single integer?

apps/remote_control/lib/lexical/remote_control/build/parse_error.ex
scohen: as opposed to macros or sigils

scohen: I don't import functions unless i'm in a test case.

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
scohen: The helper i'm talking about is the one that extracts range info from the metadata.

apps/remote_control/lib/lexical/remote_control/build/error/location.ex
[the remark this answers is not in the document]
scohen: we would make it a range that selects the entire line. Start is on `{line, 0}`, end is on `{line + 1, 0}`

apps/remote_control/lib/lexical/remote_control/build/parse_error.ex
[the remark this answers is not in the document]
scohen: well, i'm saying that they likely shouldn't be on the same level.

apps/remote_control/lib/lexical/remote_control/build/parse_error.ex
scohen:
```elixir
case Location.extract_range(context) do 
  {:ok, %Range{end: end_pos}} -> [Result.new(source.uri, pos, end_pos, message, :error, @elixir_source)]
   _ -> []
 end
```

apps/remote_control/lib/lexical/remote_control/build/error/location.ex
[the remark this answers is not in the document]
scottming: But this logic is handled in this file:https://github.com/lexical-lsp/lexical/blob/main/apps/server/lib/lexical/convertibles/lexical.plugin.diagnostic.result.ex#L50C8-L50C25. 

I think it's reasonable to do this processing at that level because before that, we still need to perform operations like `uniq` and `refine`, We need to know whether the Elixir compiler is giving us the `line` or the `position`. It makes sense to keep it as a **single integer** instead of converting it.

scohen:
@scottming can you go over this and apply my suggestions or comment about why you won't? There are a couple where you gave thumbs up and then didn't do anything.
I'd like to merge this soon.

scottming:
> @scottming can you go over this and apply my suggestions or comment about why you won't? There are a couple where you gave thumbs up and then didn't do anything.

@scohen sure, I'll try to complete it at today(January 16th)

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
scottming: So, do I still need to make some modifications?

apps/remote_control/lib/lexical/remote_control/build/error/location.ex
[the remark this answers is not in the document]
scottming: How about we handle the case of a range first?

https://github.com/lexical-lsp/lexical/pull/535/commits/6b7092e2b4d0dd2377b21c32625308cb5c0f36ee

scohen (approved):

megalithic:
💯

PR #536, opened 2024-01-02
Replaced refs with snowflakes

scohen:
Despite my idea of creating refs to represent IDs of syntactic elements, they are not globally unique, and can be reused when the VM restarts. This makes them inappropriate for long-lived ids, like we need in the index.

I've replaced them with snowflakes, which i benchmarked at 800k qps, are unique, and also represent the time they were created, which means we'll not need to store dates in the index. After the change, lexical's index was 50 MB smaller.


scottming (approved):
Snowflakes seems like a good option, but this library appears to have been unmaintained for a while.

scohen:
> but this library appears to have been unmaintained for a while.

I've used this library before, it's fast, it's correct and it's done.  Worst case, we fork it.

PR #537, opened 2024-01-02
Made schemas implement a bevaviour

scohen:
We were reaching into version specific schemas to do certain operations, and it seems like we really want the top-level schema module to do these things instead.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/store/backends/ets/schema.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/store/backends/ets/schema.ex
@@ -14,6 +14,7 @@ defmodule Lexical.RemoteControl.Search.Store.Backends.Ets.Schema do
     version = Keyword.fetch!(opts, :version)
 
     quote do
+      @behaviour unquote(__MODULE__)
       @version unquote(version)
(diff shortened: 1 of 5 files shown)


apps/remote_control/lib/lexical/remote_control/search/store/backends/ets/schema.ex
```diff
@@ -54,8 +76,15 @@ defmodule Lexical.RemoteControl.Search.Store.Backends.Ets.Schema do
     end
   end
 
-  alias Lexical.Project
-  alias Lexical.VM.Versions
+  @spec entries_to_rows(Enumerable.t(Entry.t()), module()) :: [tuple()]
+  def entries_to_rows(entries, schema_module) do
```
scottming, line 80: This PR should also be helpful for my Rename PR, so `entries_to_rows` is a general operation, right?

scottming (approved):

apps/remote_control/lib/lexical/remote_control/search/store/backends/ets/schema.ex
scohen: Yes.

PR #541, opened 2024-01-04
Block hierarchy

scohen:


apps/remote_control/lib/lexical/remote_control/search/indexer/source/reducer.ex
```diff
@@ -108,9 +113,16 @@ defmodule Lexical.RemoteControl.Search.Indexer.Source.Reducer do
   end
 
   defp push_block(%__MODULE__{} = reducer, %Block{} = block) do
-    parent = List.first(reducer.blocks)
+    parent = current_block(reducer)
     block = %Block{block | parent_id: parent.id}
-    %__MODULE__{reducer | blocks: [block | reducer.blocks]}
+    id_path = Enum.reduce(reducer.blocks, [], fn block, acc -> [block.id | acc] end)
+
+    hierarchy =
```
scottming, line 120: Considering that the struct of the block is very simple, what I would like to know is what can I do with `blocks` and `block_hierarchy` separately?

apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
```diff
@@ -20,40 +21,71 @@ defmodule Lexical.RemoteControl.Search.Indexer.Entry do
...
       path: path,
       range: range,
+      subject: subject,
       subtype: subtype,
       type: type,
       updated_at: timestamp()
     }
   end
 
+  def block?(%__MODULE__{} = entry) do
```
scottming, line 85: Where will we use this function?

apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
```diff
@@ -20,40 +21,71 @@ defmodule Lexical.RemoteControl.Search.Indexer.Entry do
...
+      block.parent_id,
+      subject,
+      type,
+      range,
+      application
+    )
   end
 
-  defp new(path, id, parent, subject, type, subtype, range, application) do
+  defp definition(path, id, block_id, subject, type, range, application) do
```
scottming, line 67: I'm not quite understanding the difference between "block_definition" and "definition."

apps/remote_control/lib/lexical/remote_control/search/indexer/source/reducer.ex
scottming: For example, to extract references of variables, if the definition is in the parent block, do I need to use the `store.backend` API to find the definition when extracting them?

apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
scohen: One is the definition of something that starts a block, a function, defmodule, fn, def, etc. the other is a definition of a thing that doesn't define a block, like a variable or a module attribute.

apps/remote_control/lib/lexical/remote_control/search/indexer/source/reducer.ex
scohen: That's correct. 
think of it this way, to search for a variable's instances, you need to start at the reference you're selecting, then find all references to it in each parent block until you get to its definition's block. Then, you need to replace all references in the definition's block, and all of its children.

apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
scohen: I figured it would be helpful for doing navigation.

scottming (approved):

PR #544, opened 2024-01-04
Support mise (new name for rtx)

x-ji:
Apparently the author of `rtx` changed its name to `mise` recently (rationale here: https://mise.jdx.dev/rtx.html). When you install `rtx` via `homebrew` and run `brew upgrade`, it automatically changes it to `mise` and with it the `rtx` command is changed to `mise` as well (but everything else works in the same way as `rtx`). This PR supports `mise`.

### DIFF
--- a/bin/activate_version_manager.sh
+++ b/bin/activate_version_manager.sh
@@ -2,7 +2,7 @@
 
 # The purpose of these functions is to detect and activate the correct
 # installed version manager in the current shell session. Currently, we
-# try to detect asdf and rtx.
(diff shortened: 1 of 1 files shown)


scohen (commented):
lgtm, @zachallaun ?

soundmonster:
In case you came here wondering why rtx doesn't work anymore after a `brew upgrade` and are looking for a temporary workaround until this PR is merged, a simple `alias rtx=mise` worked for me.

Issue #539, opened 2024-01-03
Crash when importing macros


scohen:
I was typing the following and got a crash:
```elixir
import Entry, only: :m|
```

the crash was 
```
    ** (EXIT) exited in: GenServer.call({:via, :global, {LXical.Document.Store, 1091}}, {:fetch, "file:///Users/steve/Projects/lexical/apps/remote_control/lib/lexical/remote_control/search/store/backends/ets/state.ex"}, 5000)
        ** (EXIT) an exception was raised:
            ** (CaseClauseError) no case clause matching: :m
                (lx_common 0.3.0) lib/lexical/ast/analysis/import.ex:26: anonymous fn/2 in LXical.Ast.Analysis.Import.expand_selector/1
                (elixir 1.14.5) lib/enum.ex:2468: Enum."-reduce/3-lists^foldl/2-0-"/3
                (lx_common 0.3.0) lib/lexical/ast/analysis/import.ex:22: LXical.Ast.Analysis.Import.expand_selector/1
                (lx_common 0.3.0) lib/lexical/ast/analysis/import.ex:17: LXical.Ast.Analysis.Import.new/3
                (lx_common 0.3.0) lib/lexical/ast/analysis.ex:217: LXical.Ast.Analysis.analyze_node/2
                (lx_common 0.3.0) lib/lexical/ast/analysis.ex:72: anonymous fn/2 in LXical.Ast.Analysis.traverse/2
                (elixir 1.14.5) lib/macro.ex:635: anonymous fn/4 in Macro.do_traverse_args/4
                (stdlib 4.3.1.2) lists.erl:1462: :lists.mapfoldl_1/3
```

It's a simple fix, we need to import `:all` if we don't match any of the patterns.

scohen:
Fixed by #546

PR #546, opened 2024-01-04
Fixed crash that would happen during imports

scohen:
If you import `only: :macros`, or `only: :functions`, lexical would crash, because there was no case for `only: :f`. Adding a fallthrough case fixes the issue

### DIFF
--- a/apps/common/lib/lexical/ast/analysis/import.ex
+++ b/apps/common/lib/lexical/ast/analysis/import.ex
@@ -3,7 +3,11 @@ defmodule Lexical.Ast.Analysis.Import do
   @type function_name :: atom()
   @type function_arity :: {function_name(), arity()}
   @type selector ::
-          :functions | :macros | [only: [function_arity()]] | [except: [function_arity()]]
+          :functions
+          | :macros
+          | :sigils
+          | [only: [function_arity()]]
+          | [except: [function_arity()]]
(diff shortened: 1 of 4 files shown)


scohen:
Fixes #539

apps/common/lib/lexical/ast/analysis/import.ex
```diff
@@ -35,6 +35,11 @@ defmodule Lexical.Ast.Analysis.Import do
                   {{:__block__, _, [function_name]}, {:__block__, _, [arity]}} ->
                     {function_name, arity}
                 end)
+
+              _ ->
```
scottming, line 39: Should we consider the situation with sigils?

https://github.com/elixir-lsp/elixir_sense/blob/c3771bc69ec7a907b693f6f11983e38da21848d4/test/support/fixtures/metadata_builder/import/only_sigils.ex#L3

https://github.com/lexical-lsp/lexical/pull/535/commits/026a4d709887b27176abbfef31cb9f7f3b5b3d66

scottming (approved):

apps/common/lib/lexical/ast/analysis/import.ex
scohen: yes, that one slipped by me

PR #548, opened 2024-01-09
Khepri backend

scohen:
Earlier in the week, I came to the realization that I was writing a database, which isn't what this project is about. An offhand comment from Hauleth in the elixir discord made me aware of the Khepri database by the folks who brought us rabbitMQ, and after looking at it, it seemed to be near ideal for our project.

This commit changes the backing store of the index to use the khepri database, which gives us a path-like structure to store data, journaling and snapshots, transactions, stored procedures and a whole host of other database-y things. It's a little (a lot) slower to load the initial data set (six seconds vs. a half second), but search time is quick, and the interface is fantastic.


apps/remote_control/lib/lexical/remote_control/search/store/backends/khepri.ex
```diff
@@ -0,0 +1,150 @@
...
+  require Logger
+
+  @behaviour Backend
+
+  @impl Backend
+  def new(%Project{} = project) do
+    result =
+      project
+      |> data_directory()
+      |> :khepri.start(khepri_store_id(project))
```
zachallaun, line 19: Maybe worth wrapping all this in a little helper to clean up this function body. 

```elixir
case start_khepri(project) do
  {:ok, _} = success ->
    success

  ...
end
```

apps/remote_control/lib/lexical/remote_control/search/store/backends/khepri/condition.ex
```diff
@@ -0,0 +1,7 @@
+defmodule Lexical.RemoteControl.Search.Store.Backends.Khepri.Condition do
+  require Record
+
+  for {name, fields} <- Record.extract_all(from_lib: "khepri/include/khepri.hrl") do
```
zachallaun, line 4: Can we add a moduledoc with a link to docs or something for all the records defined here?

zachallaun (commented):
Looks really clean! I still need to test locally to make sure that load time doesn't cause issues on slower machines (like mine), but this is promising.

apps/remote_control/lib/lexical/remote_control/search/store/backends/khepri/condition.ex
scohen: your "link" to "docs" is asking a lot. (Khepri uses the erlang docs, and they're _terrible_) I'll see what I can do.

zachallaun:
Even just link to this would be helpful for anyone looking at this the
first time:
https://github.com/rabbitmq/khepri/blob/main/include/khepri.hrl

scohen:
I think this one might be more helpful https://hexdocs.pm/khepri/khepri_condition.html

apps/remote_control/lib/lexical/remote_control/search/store/backends/khepri/operations.ex
```diff
@@ -0,0 +1,93 @@
...
+  shy away from using any of elixir's modules that come anywhere near protocols. That means
+  No inspect, no Enum, etc. Luckily tests will fail if we include anything we shouldn't.
+  """
+  alias Lexical.RemoteControl.Search.Indexer.Entry
+  alias Lexical.RemoteControl.Search.Store.Backends.Khepri.Condition
+
+  import Condition
+
+  def tx_insert_many(entries) do
+    :lists.foreach(&:khepri_tx.put(to_path(&1), &1), entries)
```
scottming, line 16: Enum.each

scottming (commented):
It looks really clean. I'd like to give it a try locally.

apps/remote_control/lib/lexical/remote_control/search/store/backends/khepri/operations.ex
scohen: Read the comment at the top of this file, we can't use Enum.each, because it references the Enumerable protocol.

scohen:
Closing this in favor of #554

PR #563, opened 2024-01-18
Resolve function definitions without parens

### LINKED ISSUE #561: Entity.resolve fails to resolve a function definition without parens
Entity.resolve fails to resolve the following
```elixir
defmodule Parent do
  def my_functio|n do
  end
end
```
It thinks it's a `local_or_var` when it should be a `{:call, Parent, :my_function, 0}`. If there are parens after the function name, it resolves correctly.

zachallaun:
Fixes #561


scohen:
Awesome, thanks!

PR #568, opened 2024-01-19
Indexing struct definitions and references

### LINKED ISSUE #560: Indexing: Struct definitions and references
find all calls to `defstruct` in the codebase, and references to those structs

Handle special cases of `%__MODULE__{}`, and aliased structs.

scohen:
We now index struct definitions and references, including structs created with the various implementations of Kernel.struct

Fixes #560

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/struct_definition.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/struct_definition.ex
@@ -1,6 +1,5 @@
 defmodule Lexical.RemoteControl.Search.Indexer.Extractors.StructDefinition do
-  alias Lexical.Document.Position
-  alias Lexical.Document.Range
+  alias Lexical.Ast
   alias Lexical.RemoteControl.Analyzer
   alias Lexical.RemoteControl.Search.Indexer.Entry
   alias Lexical.RemoteControl.Search.Indexer.Source.Reducer
@@ -9,7 +8,7 @@ defmodule Lexical.RemoteControl.Search.Indexer.Extractors.StructDefinition do
     document = reducer.analysis.document
     block = Reducer.current_block(reducer)
     {:ok, current_module} = Analyzer.current_module(reducer.analysis, Reducer.position(reducer))
-    range = range(document, definition)
+    range = Ast.Range.fetch!(definition, document)
 
     entry =
       Entry.definition(
@@ -26,16 +25,5 @@ defmodule Lexical.RemoteControl.Search.Indexer.Extractors.StructDefinition do
 
   def extract(_, _) do
     :ignored
-  end
-
-  defp range(document, definition) do
-    %{start: start_pos, end: end_pos} = Sourceror.get_range(definition)
-    [line: start_line, column: start_column] = start_pos
-    [line: end_line, column: end_column] = end_pos
-
-    Range.new(
-      Position.new(document, start_line, start_column),
-      Position.new(document, end_line, end_column)
-    )
   end
 end
--- a/apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/struct_definition.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/struct_definition.ex
@@ -1,0 +1,41 @@
+defmodule Lexical.RemoteControl.Search.Indexer.Extractors.StructDefinition do
+  alias Lexical.Document.Position
+  alias Lexical.Document.Range
+  alias Lexical.RemoteControl.Analyzer
+  alias Lexical.RemoteControl.Search.Indexer.Entry
+  alias Lexical.RemoteControl.Search.Indexer.Source.Reducer
+
+  def extract({:defstruct, _, [_fields]} = definition, %Reducer{} = reducer) do
+    document = reducer.analysis.document
+    block = Reducer.current_block(reducer)
+    {:ok, current_module} = Analyzer.current_module(reducer.analysis, Reducer.position(reducer))
+    range = range(document, definition)
+
+    entry =
+      Entry.definition(
+        document.path,
+        block,
+        current_module,
+        :struct,
+        range,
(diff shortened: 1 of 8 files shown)


Blond11516 (approved):
Looks and works nice!

2 thoughts for future improvements:
- I see from the associated issue that there are already intended uses for struct definitions outside of navigation, but I couldn't actually navigate to to a struct definition despite the new extractor. Anything I tried sent me to the module definition. I think it'd be nice to detect go to definition requests on struct references and actually navigate to the `defstruct` call rather than the module (I assume it would be doable because find references already behaves like this, at least for `%{}` references). I think it'd help a little when you're actually looking for the struct definition, and if you were just looking for the module, well you're in there anyway, just not at the top.
- Using find references requests on `defstruct` to find references to the corresponding struct. It currently does nothing but that would be the most useful behavior imo. I don't really see a use in finding every reference to the actual `defstruct` macro (which a simple text search can easily do anyway)

These are obviously not in scope for this PR but I'd like your thoughts on them!

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/struct_definition.ex
```diff
@@ -0,0 +1,41 @@
...
+      )
+
+    {:ok, entry}
+  end
+
+  def extract(_, _) do
+    :ignored
+  end
+
+  defp range(document, definition) do
```
zachallaun, line 31: We do this in enough places that, IMO, it should really be a utility function like `Lexical.Ast.range_in_document(node, document)`.

This function could also raise an exception with a useful message if `Sourceror.get_range` returns nil for the given node (which it can), which will help us see the actual syntax it's failing on instead of whatever downstream match failure occurs because it returns nil instead of a map.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/struct_reference.ex
```diff
@@ -0,0 +1,103 @@
...
+  end
+
+  # handles calls to Kernel.struct e.g. struct(MyModule) or struct(MyModule, foo: 3)
+  def extract({struct_fn_name, _, [struct_alias | _] = args} = reference, %Reducer{} = reducer)
+      when struct_fn_name in @struct_fn_names do
+    reducer_position = Reducer.position(reducer)
+    imports = Analyzer.imports_at(reducer.analysis, reducer_position)
+    arity = length(args)
+
+    with true <- Enum.member?(imports, {Kernel, :struct, arity}),
```
zachallaun, line 47: I think it should be this to also handle `struct!`?

```suggestion
    with true <- Enum.member?(imports, {Kernel, struct_fn_name, arity}),
```

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/struct_reference.ex
```diff
@@ -0,0 +1,103 @@
...
+      document.path,
+      block,
+      subject,
+      :struct,
+      range(document, reference),
+      Application.get_application(struct_module)
+    )
+  end
+
+  defp range(%Document{} = document, ast) do
```
zachallaun, line 75: See comment above; I think we should stop rewriting this function.

zachallaun (approved):

scohen:
> Looks and works nice!
> 
> 2 thoughts for future improvements:
> 
> * I see from the associated issue that there are already intended uses for struct definitions outside of navigation, but I 
couldn't actually navigate to to a struct definition despite the new extractor. 

Did you blow away the old index and reindex? I'm able to navigate to struct definitions after doing so. 

> Anything I tried sent me to the module definition. 

This is the old behavior, I think you might have an old version running.

> * Using find references requests on `defstruct` to find references to the corresponding struct. It currently does nothing but that would be the most useful behavior imo. I don't really see a use in finding every reference to the actual `defstruct` macro (which a simple text search can easily do anyway)

I like this a lot more, you're right, find references for 'defstruct' should behave differently than other calls. I'll update.

> These are obviously not in scope for this PR but I'd like your thoughts on them!

I disagree, both are in scope for the PR, and I'll address the second, because I think it already works like you expect, and you might need to reindex for it to work. Remember, you can go to the project's top-level `mix.exs` and click the reindex code lens.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/struct_reference.ex
```diff
@@ -0,0 +1,103 @@
...
+        {:ok, entry(reducer, struct_module, reference)}
+
+      _ ->
+        :ignored
+    end
+  end
+
+  # Call to Kernel.struct with a fully qualified module e.g. Kernel.struct(MyStruct, ...)
+  def extract(
+        {{:., _, [kernel_alias, struct_fn_name]}, _, [struct_alias]} = reference,
```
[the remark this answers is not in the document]
scohen: good catch

scohen:
> Using find references requests on defstruct 

Spoke too soon. This is actually another issue with CodeIntelligence.Entity. It cannot resolve `defstruc|t` as a call to defstruct, so we can't do the proper search. Now, if it did resolve it properly to a call to `Kernel.defstruct`, the current behavior would definitely find all references to that, but we could easily capture the specific call in `references.ex` and instead get the current module and get references for its struct usages. I'll write up an issue for the first problem.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/struct_definition.ex
scohen: I added `Lexical.Ast.Range.fetch` and `Lexical.Ast.Range.fetch!` for this.

scohen:
> Anything I tried sent me to the module definition

I need to read more carefully, @Blond11516. This is because we're using ElixirSense's go to definition and not our index. We'll have to switch away from elixir sense for go to definition in order to get the actual range for the defstruct, though we're definitely storing it correctly.

PR #583, opened 2024-02-08
Fixed bug due to elixir_sense upgrade

scohen:
When elixir_sense was upgraded, it added an :alias subtype for module completions, which we didn't handle, so all aliases failed to complete.


scohen:
Going to merge this, it's a pretty annoying bug.

PR #586, opened 2024-02-08
Entity.resolve resolves defstruct correctly

### LINKED ISSUE #569: Entity.resolve fails on `defstruct`
given the following code:
```elixir
defmodule Struct do 
  defstruct [:foo]
```

`Entity.resolve` cannot resolve the call to defstruct on the second line, no matter where the cursor position is.

Expected behavior: `{:ok, {:call, Kernel, :defstruct, 2}}`. Actual behavior: `{:error, :not_found}`

scohen:
Fixes #569

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
@@ -63,18 +63,28 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
     resolve_alias(chars, node_range, analysis, position)
   end
 
-  defp resolve({:local_or_var, 'defstruct'}, node_range, _analysis, _position) do
-    {:ok, {:call, Kernel, :defstruct, 1}, node_range}
-  end
-
   defp resolve({:local_or_var, chars}, node_range, analysis, position) do
     maybe_fun = List.to_atom(chars)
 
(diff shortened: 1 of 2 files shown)


apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -63,6 +63,10 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
     resolve_alias(chars, node_range, analysis, position)
   end
 
+  defp resolve({:local_or_var, 'defstruct'}, node_range, _analysis, _position) do
```
zachallaun, line 66: Why does this need to be special-cased? Shouldn't this be resolved by looking at imports?

For instance, while ill-advised, one could shadow `defstruct` and use their own wrapper, e.g.

```elixir
defmodule MyModule do
  import Kernel, except: [defstruct: 1]
  import MyDefstruct
  
  # from MyDefstruct
  defstruct x: 1, y: 2
end
```

scohen: It's a little unclear to me how entity should handle this generically though. There's a local or var case already, but that seems to want to return a 0 arity call.

scohen: worse, i cannot find a way to reliably determine the arity of the function in question if it doesn't have parens.

scohen: ok, i took another stab at it.

PR #592, opened 2024-02-13
Removed select_all() from backend in favor of a reduce function

scohen:
There are several places where we're pulling all the entries from the backend, processing them and throwing them out. This creates memory pressure on every process involved in handing the entries.

Instead of that, it makes more sense to expose a reduce function inside the backend that allows you to process the entries however you want, and only the data you care about is pulled through the processes.

This saves around 250mb running the lexical project on my laptop.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/indexer.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/indexer.ex
@@ -30,7 +30,7 @@ defmodule Lexical.RemoteControl.Search.Indexer do
   end
 
   defp do_update_index(%Project{} = project, backend) do
-    path_to_last_index_at =
+    path_to_ids =
       backend.reduce(%{}, fn %Entry{path: path} = entry, path_to_ids when is_integer(entry.id) ->
         Map.update(path_to_ids, path, entry.id, &max(&1, entry.id))
       end)
@@ -40,16 +40,16 @@ defmodule Lexical.RemoteControl.Search.Indexer do
       |> indexable_files
       |> MapSet.new()
 
-    previously_indexed_paths = MapSet.new(path_to_last_index_at, fn {path, _} -> path end)
+    previously_indexed_paths = MapSet.new(path_to_ids, fn {path, _} -> path end)
 
     new_paths = MapSet.difference(project_files, previously_indexed_paths)
 
     {paths_to_examine, paths_to_delete} =
-      Enum.split_with(path_to_last_index_at, fn {path, _} -> File.regular?(path) end)
+      Enum.split_with(path_to_ids, fn {path, _} -> File.regular?(path) end)
 
     changed_paths =
-      for {path, updated_at_timestamp} <- paths_to_examine,
-          newer_than?(path, updated_at_timestamp) do
+      for {path, id} <- paths_to_examine,
+          newer_than?(path, id) do
         path
       end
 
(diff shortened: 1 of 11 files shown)


apps/remote_control/lib/lexical/remote_control/search/indexer.ex
```diff
@@ -21,33 +23,33 @@ defmodule Lexical.RemoteControl.Search.Indexer do
...
   end
 
-  defp do_update_index(%Project{} = project, existing_entries) do
-    path_to_last_index_at =
-      existing_entries
-      |> Enum.group_by(& &1.path, & &1.updated_at)
-      |> Map.new(fn {k, v} -> {k, Enum.max(v)} end)
+  defp do_update_index(%Project{} = project, backend) do
+    path_to_ids =
+      backend.reduce(%{}, fn %Entry{path: path} = entry, path_to_ids when is_integer(entry.id) ->
```
vrcca, line 34: @scohen Thanks for this amazing project!

I am running the latest version from `main` and I noticed the search store was crashing, and it was because this line expecting that entry.id is always an integer, but it's crashing on `:root` atom:

```
==> .lexical/project.log <==
2024-02-16T14:51:56.623008+01:00 info: Compile completed with status ok Produced 0 diagnostics []
2024-02-16T14:51:56.639593+01:00 info: Plugins found 0 results in 0.011 ms
2024-02-16T14:52:03.841408+01:00 notice: disk_log: repairing "/Users/vitor/Developer/TheApp/.lexical/indexes/ets/26.1.2/1.16.0/2/updates.wal" ...
2024-02-16T14:52:19.797334+01:00 info: backend reports stale
2024-02-16T14:52:22.421439+01:00 error: GenServer LXical.RemoteControl.Search.Store.Backends.Ets terminating, ** (FunctionClauseError) no function clause matching in anonymous fn/2 in LXical.RemoteControl.Search.Indexer.do_update_index/2, (lx_remote_control 0.3.0) lib/lexical/remote_control/search/indexer.ex:34: anonymous fn(%LXical.RemoteControl.Search.Indexer.Entry{application: :exvcr, id: :root, block_id: nil, path: "/Users/vitor/Developer/TheApp/deps/exvcr/lib/exvcr/exceptions.ex", range: %LXical.Document.Range{start: #LXical.Document.Position<line: 5, character: 11, ...>, end: #LXical.Document.Position<line: 5, character: 37, ...>}, subject: ExVCR.RequestNotMatchError, subtype: :definition, type: :module}, "ahugemap, and parameters I can't share here" 
```

I applied this change locally:

```diff
diff --git a/apps/remote_control/lib/lexical/remote_control/search/indexer.ex b/apps/remote_control/lib/lexical/remote_control/search/indexer.ex
index bc93477..54b3bcb 100644
--- a/apps/remote_control/lib/lexical/remote_control/search/indexer.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/indexer.ex
@@ -31,8 +31,12 @@ defmodule Lexical.RemoteControl.Search.Indexer do

   defp do_update_index(%Project{} = project, backend) do
     path_to_ids =
-      backend.reduce(%{}, fn %Entry{path: path} = entry, path_to_ids when is_integer(entry.id) ->
-        Map.update(path_to_ids, path, entry.id, &max(&1, entry.id))
+      backend.reduce(%{}, fn %Entry{path: path} = entry, path_to_ids ->
+        if is_integer(entry.id) do
+          Map.update(path_to_ids, path, entry.id, &max(&1, entry.id))
+        else
+          path_to_ids
+        end
       end)

     project_files =
```

Should I open a ticket?

scohen: Please open an issue with a reproducible case

scohen: I see the issue here, I'll get working on it.

PR #616, opened 2024-02-22
Removed -e option

### LINKED ISSUE #615: start_lexical.sh exiting early when no version manager is installed
I'm using the vscode extension, and I have verified that it has installed the latest version of lexical ("0.5.0").

The vscode extension fails to start up since it's trying to run `start_lexical.sh` which exits with exit code 1.
I have looked into `start_lexical.sh` and `activate_version_manager.sh` and I have found that

* Since I have none of these version managers installed (I installed Elixir via homebrew) it goes through the whole `activate_version_manager.sh`, printing "No activated version manager detected. Searching for version manager..." and exits with exit code 1 at the end after trying mise
* At this point, I believe `start_lexical.sh` should print "Could not activate a version manager. Trying system installation." and continue to start lexical, but instead the script just exits with code 1 without printing anything.

From what I can tell this is happening because of `set -e` at the start of `start_lexical.sh`, if I change the second line to just `set -o pipefail` instead of `set -eo pipefail` the script works as expected for me. 

I have to say I'm not sure *why* because from what I understand `set -e` shouldn't cause an exit if the non-zero exiting command is called in an if-statement, which it is, but it does happen 🤷
For context I am on MacOS and my bash version is
```
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin23)
Copyright (C) 2007 Free Software Foundation, Inc.
```

scohen:
The -e option exits immediately on an error, which will cause the command to exit if no version manager is detected. This means we won't try to invoke elixir from the path.

From my limited knowledge of bash, it looks like this command was added in error.

Fixes #615

### DIFF
--- a/bin/start_lexical.sh
+++ b/bin/start_lexical.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -eo pipefail
+set -o pipefail
 
 script_dir="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
 


zachallaun (approved):
My understanding was that `-e` wouldn't cause an exit if a command used as an `if` condition returned non-zero, but apparently (as with all things bash) there are caveats to that. If this fixes the issue, then sounds good to me! The "benefit" of `-e` is mostly to not let errors that will cause downstream problems go undetected -- clearly that's not working here.

PR #621, opened 2024-02-23
Added variable extractor and ability to customize which extractors run

scohen:
Added an extractor that pulls out variables and references from AST. It's likely incomplete, but we can add missed cases as we go.

I also added the ability to customize which extractors run during an index. The idea here is to have a separate indexing run after a file is compiled, so we can pull out variables from it and power go to definition for variables without polluting the index with things that we won't use or need.


apps/remote_control/test/lexical/remote_control/search/indexer/extractors/variable_test.exs
```diff
@@ -0,0 +1,440 @@
...
+    assert entry.subject == variable_name
+  end
+
+  def assert_reference(entry, variable_name) do
+    assert entry.type == :variable
+    assert entry.subtype == :reference
+    assert entry.subject == variable_name
+  end
+
+  for def_type <- [:def, :defp, :defmacro, :defmacrop] do
```
scottming, line 30: I like the simplicity of this PR, but I'm curious, besides the unimplemented cross-scope variable extraction, what other scenarios have we deliberately not implemented?

Regarding testing, I think we're missing cases where definitions and references are extracted together, like this:

```elixir
  test "definition at `=`'s right side" do
    {:ok, [ref], doc} = ~q/
        fn
          1 = a -> a
        end
      / |> index_references()

    assert decorate(doc, ref.range) =~ "1 = a -> «a»"
  end

  test "simple reference" do
    {:ok, [ref], doc} = ~q/
      def myfunc(a) do
        [a]
      end
      / |> index_references()

    assert decorate(doc, ref.range) =~ "[«a»]"
  end
```

**Note**, the current code cannot pass these two tests.

Additionally, how would you differentiate between multiple variables with the same name appearing in a single scope?

apps/remote_control/lib/lexical/remote_control/search/indexer/source/reducer.ex
```diff
@@ -13,7 +13,7 @@ defmodule Lexical.RemoteControl.Search.Indexer.Source.Reducer do
   alias Lexical.RemoteControl.Search.Indexer.Metadata
   alias Lexical.RemoteControl.Search.Indexer.Source.Block
 
-  defstruct [:analysis, :entries, :position, :ends_at, :blocks, :block_hierarchy]
+  defstruct [:analysis, :entries, :position, :blocks, :block_hierarchy, extractors: []]
```
scottming, line 16: I noticed that the default value for `extractors` in many functions is nil, but here it is `[]`. Is there any particular reason for this

apps/remote_control/test/lexical/remote_control/search/indexer/extractors/variable_test.exs
scohen: @scottming, the first example isn't valid syntax, so it's going to fail and produce no results. The second is a valid bug though.

scohen:
> Additionally, how would you differentiate between multiple variables with the same name appearing in a single scope?

For the purposes of extraction, there's no handling to speak of, they're valid shadowing of variables. 

Once things are extracted, if you want to find a definition, you get all the definitions in the current scope, sort them by line and column, and find the nearest one. If that doesn't exist, you get the parent scope and try that again.

Finding references is much the same as definitions, but you carry an accumulator with you, and while the definition isn't found, add any references to the collection. Once a definition is found, you add all references that occur after the definition.

scohen:
> I like the simplicity of this PR, but I'm curious, besides the unimplemented cross-scope variable extraction

What do you mean? 

> what other scenarios have we deliberately not implemented?

I've tried to be complete here, if something isn't implemented, please tell me.

scottming:
> What do you mean?

> Once things are extracted, if you want to find a definition, you get all the definitions in the current scope, sort them by line and column, and find the nearest one that matches the variable name you're looking for

I think your answer above already addresses this question. I previously thought there was a need to differentiate the variable definitions for each scope.

apps/remote_control/test/lexical/remote_control/search/indexer/extractors/variable_test.exs
scottming: I don't think the first example is incorrect syntax; it can be run directly in iex. Also, with a simple function like this: `fn x -> x end`, we can't extract the variables and references. I think we need more tests with references and definitions together.

CleanShot 2024-02-28 at 20 36 38@2x

apps/remote_control/test/lexical/remote_control/search/indexer/extractors/variable_test.exs
```diff
@@ -0,0 +1,603 @@
...
+        ]
+        |> index_definitions()
+
+      assert_definition(param, :var)
+      assert decorate(doc, param.range) =~ "fn «var» ->"
+    end
+
+    test "in a struct's values" do
+      {:ok, [param], doc} =
+        ~q[
```
scottming, line 276: The same issue, if I make some changes to this example, it fails to extract the reference.

```elixir
      {:ok, [param], doc} =
        ~q[
        fn %Pattern{foo: var} ->
          var
        end
        ]
        |> index_references()
```

scottming:
> I've tried to be complete here, if something isn't implemented, please tell me.

This PR doesn't seem to handle the variable definitions in these blocks, for example: `[:else, :rescue, :catch, :after]`

```elixir
    test "When it appears in the else block of a with statement." do
      {:ok, [func, var], doc} =
        ~q[
        func = fn _ -> :not_ok end

        with :ok <- func.(1) do
          :ok
        else
          var -> var
        end
        ]
        |> index_definitions()

      assert_definition(func, :func)
      assert decorate(doc, func.range) =~ "«func» = fn _ ->"

      # NOTE: current PR doesn't handle the `var` in the `else` block
      assert_definition(var, :var)
      assert decorate(doc, var.range) =~ "  «var» ->"
    end
```

apps/remote_control/test/lexical/remote_control/search/indexer/extractors/variable_test.exs
scohen: The syntax you typed was not inside an anonymous function, and that works, even in the old code.

scohen: You're right, I missed the lone `fn` on the line above.

scohen:
> This PR doesn't seem to handle the variable definitions in these blocks, for example: [:else, :rescue, :catch, :after]

Correct me if i'm wrong, but I don't think you can define a variable in an after block.

The problem isn't really the blocks either, it's that the stab (`->`) operator wasn't handled properly. It was special-cased for the `fn` case. Removing the special case allows for more generic handling of defining variables in stabs.

scohen:
@scottming I believe i've addressed the latest issues.

apps/remote_control/test/lexical/remote_control/search/indexer/extractors/variable_test.exs
```diff
@@ -0,0 +1,760 @@
...
+        assert decorate(doc, second_ref.range) =~ "  [^first, other] = «second»"
+
+        assert_reference(other_ref, :other)
+        assert decorate(doc, other_ref.range) =~ " «other»"
+      end
+    end
+  end
+
+  describe "variable definitions in anonymous function parameters are extracted" do
+    test "hen definition on the right side of the equals" do
```
scottming, line 261:
```suggestion
    test "when definition on the right side of the equals" do
```

scottming (approved):
I can't think of any more cases for now. I think we can go ahead and merge this PR.

PR #628, opened 2024-02-29
Fixed unused variable code action

scohen:
The quote character around the variable name was changed, this broke the code action for replace unused variables.


scottming:
The initial reason for making this change was that I thought using backticks for the token aligns better with markdown syntax, indicating it's part of the code. This will be advantageous for editors supporting *treesitter*. 

More importantly, I find that compared to `"`, backticks are easier on the eyes to recognize, as `"` can be disruptive to identify. You can see the comparison in the following few images:

|  editor   | " | backtick |
|  ----  | ----  |---|
| nvim  | image 4 | image |
| vscode  | image 3 | image 2 |

scohen:
How is it advantageous for a treesitter based editor? Diagnostics  aren't part of the code, and aren't displayed via markdown.

scottming (approved):

PR #640, opened 2024-03-04
Implement callback completions

Moosieus:
Callback completions should now expand to their full def-blocks, preceded with an impl tag.

In a module `use`-ing a behaviour, typing `def |` should result in the following completion options:
* `def (define a function)`
* `def callback(args) module_label`
* `(remaining callbacks)`
* Other `def*` macros

### DIFF
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/callback.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/callback.ex
@@ -37,7 +37,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Callback do
     end
 
     # add tab stops and join with ", "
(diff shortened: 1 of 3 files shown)


Moosieus:
CI seems to be failing due to an unrelated issue:
```
 Could not fetch /builds/otp/ubuntu-22.04/builds.txt from any hex.pm mirror
 ```
 
 In the meantime, I need to take a second pass at the callback completion and add additional tests. Presently `def handle|` doesn't overwrite the original `def ` properly.

scohen:
@Moosieus I re-ran the tests, these failures seem legit.

Moosieus:
Keeping the sorting as-is for now since I've already pitched the overhaul in #643.

scohen (approved):

PR #642, opened 2024-03-06
Fix `do` completing to `defmodule` in VSCode

Moosieus:
Pursuant to: https://discord.com/channels/269508806759809042/918183981344772117/1214563122686206033
> objectuser — Today at 8:20 AM
I'm using Lexical in VSCode. When I'm creating a do block, the first completion is defmodule and there's no completion listed in the pop-up for just do...end. Should I just use user snippets for a do...end block or is there a better way to configure these completions?

(From Elixir Discord)

Fixes include:
* Include partial `d|` in `should_emit_do_end_snippet`
* Add `filter_text: "do"` for more accurate fuzzy matching

This may warrant user-testing across different editors to ensure existing completions remain consistent, as they may convert user-input-over-time to `textDocument/completion` requests differently.

### DIFF
--- a/apps/server/lib/lexical/server/code_intelligence/completion.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion.ex
@@ -58,7 +58,11 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
         do_end_snippet = "do\n  $0\nend"
 
         env
-        |> Builder.snippet(do_end_snippet, label: "do/end block")
+        |> Builder.snippet(
+          do_end_snippet,
+          label: "do/end block",
+          filter_text: "do"
+        )
         |> List.wrap()
 
       Env.in_context?(env, :struct_field_key) ->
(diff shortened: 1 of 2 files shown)


apps/server/lib/lexical/server/code_intelligence/completion.ex
```diff
@@ -58,7 +58,11 @@ defmodule Lexical.Server.CodeIntelligence.Completion do
         do_end_snippet = "do\n  $0\nend"
 
         env
-        |> Builder.snippet(do_end_snippet, label: "do/end block")
+        |> Builder.snippet(
+          do_end_snippet,
+          label: "do/end block",
+          filter_text: "do"
```
scohen, line 64: shouldn't the filter_text be whatever the prefix was? (likely `d`) ?

Moosieus: AFAICT there's no hard and fast rules to filter texts, they just have to fuzzy match well to the expected inputs.

I ran a quick test on this with vscode's client:
* `filter_text: "do"` provides the completion for both `d|` and `do|`.
* `filter_text: "d"` only provides the completion for `d|` (doesn't pass the fuzzy match margin).

Hence I'd say stick with `filter_text: "do"`.

I'll have to get myself setup with neovim so I can validate there as well.

Broader scope, perhaps an LSP client test suite's in order at some point - that's a big uplift though.

scohen: vscode has been the most finicky with this, emacs already works.

scohen (approved):

Issue #641, opened 2024-03-04
Unsaved text buffers in vscode crash Lexical


Moosieus:
Relevant log, note the second parameter which should be `%State{}` is `nil`:
```
18:45:30.371 [error] GenServer LXical.Server terminating
** (FunctionClauseError) no function clause matching in LXical.Server.Provider.Queue.add/2
    (lx_server 0.5.0) lib/lexical/server/provider/queue.ex:154: LXical.Server.Provider.Queue.add(%LXical.Protocol.Requests.CodeLens{lsp: %LXical.Protocol.Requests.CodeLens.LSP{id: 68, jsonrpc: "2.0", method: "textDocument/codeLens", partial_result_token: nil, text_document: #Protocol.Types.TextDocument.Identifier<[uri: "untitled:Untitled-1"]>, work_done_token: nil}, document: nil, id: 68, jsonrpc: "2.0", method: "textDocument/codeLens", partial_result_token: nil, text_document: nil, work_done_token: nil}, nil)
    (lx_server 0.5.0) lib/lexical/server.ex:144: LXical.Server.handle_message/2
    (lx_server 0.5.0) lib/lexical/server.ex:67: LXical.Server.handle_cast/2
    (stdlib 5.2) gen_server.erl:1121: :gen_server.try_handle_cast/3
    (stdlib 5.2) gen_server.erl:1183: :gen_server.handle_msg/6
    (stdlib 5.2) proc_lib.erl:241: :proc_lib.init_p_do_apply/3
Last message: {:"$gen_cast", {:protocol_message, %LXical.Protocol.Requests.CodeLens{lsp: %LXical.Protocol.Requests.CodeLens.LSP{id: 68, jsonrpc: "2.0", method: "textDocument/codeLens", partial_result_token: nil, text_document: #Protocol.Types.TextDocument.Identifier<[uri: "untitled:Untitled-1"]>, work_done_token: nil}, document: nil, id: 68, jsonrpc: "2.0", method: "textDocument/codeLens", partial_result_token: nil, text_document: nil, work_done_token: nil}}}
State: %LXical.Server.State{configuration: nil, initialized?: false, shutdown_received?: false, in_flight_requests: %{}}
```

The URI appears as `untitled:Untitled-1`.

So far as I can immediately tell, these unsaved files are sent over as `textDocument/didOpen` notifications, followed by relevant `textDocument/didChange` events when edits are made:
```
[...]

PR #644, opened 2024-03-10
Fix crashing in unsaved vscode files

Moosieus:
Pursuant to #641

* match on and handle `untitled:Untitled-N` URIs. Previously they were errantly being treated as file path URIs.
* add `is_binary` guards to `Document.Path` functions. This makes for slightly more explicit errors.

### DIFF
--- a/projects/lexical_shared/lib/lexical/document/path.ex
+++ b/projects/lexical_shared/lib/lexical/document/path.ex
@@ -59,7 +59,7 @@ defmodule Lexical.Document.Path do
   end
 
   def from_uri(%URI{scheme: scheme}) do
-    raise ArgumentError, message: "unexpected URI scheme #{inspect(scheme)}"
+    raise ArgumentError, message: "unsupported URI scheme #{inspect(scheme)}"
   end
 
   def from_uri(uri) when is_binary(uri) do
--- a/projects/lexical_shared/lib/lexical/document/path.ex
+++ b/projects/lexical_shared/lib/lexical/document/path.ex
@@ -14,7 +14,7 @@ defmodule Lexical.Document.Path do
 
   def ensure_uri("untitled:" <> _ = uri), do: uri
 
-  def ensure_uri(path) when is_binary(path), do: to_uri(path)
+  def ensure_uri(path), do: to_uri(path)
 
   @doc """
   Given a uri or a path, either return the path unmodified or converts the uri to a path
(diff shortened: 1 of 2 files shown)


Moosieus:
Test at `projects/lexical_shared/test/lexical/document/path_test.exs:118` is failing, although vscode's working exactly as intended with the patch in place.

Might need to use a different test case since `untitled:*` is explicitly handled now, or just forego it entirely.

projects/lexical_shared/lib/lexical/document/path.ex
```diff
@@ -52,26 +54,37 @@ defmodule Lexical.Document.Path do
     convert_separators_to_native(path)
   end
 
+  def from_uri(%URI{scheme: "untitled", path: name}) do
+    "untitled:" <> name
```
scottming, line 58: It should be noted that 'untitled' is not an official specification of lsp. There is a related discussion here of neovim, but I'm not sure if there's a better solution.

* https://github.com/neovim/neovim/issues/21276

Moosieus: No exact scheme (`http`, `sftp`, even `file`, etc) of URI is officially supported by LSP, the spec officially supports any valid URI as defined by rfc3986.

Turns out `untitled:Name` [is actually a valid URI](https://datatracker.ietf.org/doc/html/rfc3986#section-3), *which I didn't know until a few hours ago!* :D

Passing it to `URI.parse/1` even works:
```elixir
iex(1)> URI.parse("untitled:Untitled-1")
%URI{
  scheme: "untitled",
  userinfo: nil,
  host: nil,
  port: nil,
  path: "Untitled-1",
  query: nil,
  fragment: nil
}
```

For now I think matching against `untitled:` is clean enough, and it shores up a major usability/adoption issue.

scohen:
Also, tests failed, though they look spurious now, in light of the current revelations.

projects/lexical_shared/test/lexical/document/path_test.exs
```diff
@@ -115,14 +115,8 @@ defmodule ElixirLS.LanguageServer.SourceFile.PathTest do
       end)
     end
 
-    test "wrong schema" do
-      assert_raise ArgumentError, fn ->
-        from_uri("untitled:Untitled-1")
-      end
-
-      assert_raise ArgumentError, fn ->
-        from_uri("unsaved://343C3EE7-D575-486D-9D33-93AFFAF773BD")
```
scohen, line 124: this also looks oddly specific, can we keep it?

Moosieus: Added back, slightly adjusted.

Moosieus:
Hard for me to say what the ramifications are - here's what I got though:
* All tests pass.
* Lexical no longer crashes due to unsaved files in vscode.
* Makes `document.path` a misnomer since values could be `untitled:Untitled-N` now.
* Other files in vscode seemingly don't receive completions from unsaved files (I think this is alright for now).

Everything else seems to work as anticipated.

scohen:
the misnomer is making me worried, we've assumed paths are... paths, and can be operated on with the Path module

scohen:
I'm going to land this, and we'll see how it goes.

scohen (approved):

PR #645, opened 2024-03-12
Find references for variables

scohen:
This PR implements find references for variables and fixes a couple bugs and missed cases in the variable indexer.

There's a notable edge case present in the implementation. A rebound variable that's assigned to a block will effectively shadow the variable of that name inside the block. I believe this is a limitation of the information we get back from the indexer, so another approach might be necessary in the future if this becomes an issue. An example follows:

```elixir
name = "Stinky"
name = if some_condition() do
  "Smelly"
else
  name
end
```

Searching for references for the `name` variable on the first line will return no results, but searching for references to the rebind on the second line will return the reference in the `else` clause in the if.

### DIFF
--- a/apps/remote_control/test/lexical/remote_control/code_intelligence/variable_test.exs
+++ b/apps/remote_control/test/lexical/remote_control/code_intelligence/variable_test.exs
@@ -1,0 +1,489 @@
+defmodule Lexical.RemoteControl.CodeIntelligence.VariableTest do
+  alias Lexical.Ast
+  alias Lexical.RemoteControl.CodeIntelligence.Variable
+
+  use ExUnit.Case
+
+  import Lexical.Test.CodeSigil
+  import Lexical.Test.CursorSupport
+  import Lexical.Test.RangeSupport
+
+  def find_definition(code) do
+    {position, document} = pop_cursor(code, as: :document)
+    analysis = Ast.analyze(document)
+    {:ok, {:local_or_var, var_name}} = Ast.cursor_context(analysis, position)
+
+    case Variable.definition(analysis, position, List.to_atom(var_name)) do
+      {:ok, entry} -> {:ok, entry.range, document}
+      error -> error
+    end
+  end
+
+  def find_references(code, include_definition? \\ false) do
+    {position, document} = pop_cursor(code, as: :document)
+    analysis = Ast.analyze(document)
+    {:ok, {:local_or_var, var_name}} = Ast.cursor_context(analysis, position)
+
+    ranges =
+      analysis
+      |> Variable.references(position, List.to_atom(var_name), include_definition?)
+      |> Enum.map(& &1.range)
+
+    {:ok, ranges, document}
+  end
+
+  describe "definitions in a single scope" do
+    test "are returned if it is selected" do
+      {:ok, range, doc} =
+        ~q[
+          def foo(param|) do
+            param
+          end
+        ]
+        |> find_definition()
+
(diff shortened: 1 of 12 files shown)


apps/remote_control/test/lexical/remote_control/code_intelligence/variable_test.exs
```diff
@@ -0,0 +1,489 @@
...
+
+      assert decorate(doc, range) =~ "  case «param» do"
+    end
+
+    test "are constrained to a single arm of a case statement" do
+      {:ok, [range], doc} =
+        ~q[
+          def something(param) do
+            case param do
+             param| when is_number(param) -> param + 1
```
scottming, line 298: There might be something odd here where this `is_number(params)` cannot be found, and similarly, when the cursor is at this position: `|param + `, it can't find the definition on the left either.

scottming:
> There's a notable edge case present in the implementation. A rebound variable that's assigned to a block will effectively shadow the variable of that name inside the block. I believe this is a limitation of the information we get back from the indexer, so another approach might be necessary in the future if this becomes an issue. An example follows:

I took a closer look at the implementation of that algorithm. If I understand correctly, the main reason is that by the time it gets to the second definition, it has already been shadowed, so it cannot further collect references.

Also, when the cursor is on a reference, this algorithm doesn't check whether the identically named definition before the cursor is the definition for that reference. Maybe adding this check could solve the problem, but it seems tricky because there are too many ways of defining things.

For example, like in this case:

```elixir
code =
  ~q{
    entries = [1, 2, 3]
    entries =
      if something() do
        [4 | entries]
      else
        |entries # __CURSOR__ 
      end
  }

{position, document} = pop_cursor(code, as: :document)
quoted = document |> Ast.analyze()
cursor_path = Ast.cursor_path(quoted, position)
#Here we will find the second line, so we exclude this line and continue searching upward for the real definition.
Macro.path(cursor_path, &match?({:=, _, _}, &1)) 
```

This is just a simple idea, hoping to provide some food for thought rather than adding more burden.😂

scottming:
Today, while using it at work, I realized we missed some **test-related** cases.

For example, variables within the context of a single test - some of them should be variables defined in setup.

```elixir
describe "test something" do
  setup do
    v = 1
    %{v: v}
  end

  test "case 1", %{v: v} do
    v
  end
end

**Comment by scohen** [comment]
I disagree, that variable is defined in the test declaration, there can be multiple setup blocks, and later ones can override previous ones.  That will be impossible to figure out.

**Comment by scottming** [comment]
> I disagree, that variable is defined in the test declaration, there can be multiple setup blocks, and later ones can override previous ones. That will be impossible to figure out.

In that case, then I think we can first ignore the approach of finding definitions in setup, and only focus on the context and test block, something like this:

```
test "case 1", %{n: n} do # definition
    v + 1 # ref1
    assert v == 1 #ref2
end
 ```

scohen:
@scottming added a special case for tests.

scottming (approved):
I think we can merge it now and further optimize accuracy when we want to implement the rename variable feature in the future.

scohen:
Yeah, that's what I'm hoping. This is good, but could be better.

apps/remote_control/test/lexical/remote_control/code_intelligence/variable_test.exs
scohen: Yep, we weren't handling guards properly. They have a very odd AST representation that needs to be handled specially in both the `def` cases and the anonymous function cases.

PR #648, opened 2024-03-15
Add summaries to callable completions.

Moosieus:
* Include `:summary` from elixir_sense callback candidates.
* Append `@spec`'s to them as well.
* Format `:documentation` as markdown in LSP completions.
* Include the type of callable as a completion detail.

### DIFF
--- a/apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
+++ b/apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
@@ -54,6 +54,7 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Callable do
       kind: kind,
       detail: "(#{callable.type})",
       sort_text: sort_text(callable),
+      filter_text: "#{callable.name}",
       documentation: maybe_summary(callable),
       tags: tags
     )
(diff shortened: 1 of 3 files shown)


apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
```diff
@@ -136,4 +143,18 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Callable do
...
+      if Map.has_key?(callable, :summary) do
+        callable.summary
+      end
+
+    spec =
+      if Map.has_key?(callable, :spec) do
+        "```elixir\n#{callable.spec}\n```"
+      end
+
+    "#{summary}\n#{spec}"
```
scohen, line 158: this seems wonky, if you don't have a summary or a spec, you will generate a spurious empty line, right?
how about
```elixir
case callable do 
  %{summary: summary, spec: spec} when is_binary(summary) and is_binary(spec) ->  
     "#{summary}\n```elixir\n#{spec}\n```"
  %{summary: summary} when is_binary(summary) ->
     summary
  %{spec: spec} when is_binary(spec)
    "```elixir\n#{spec}\n```
  _ -> 
    ""
end
```
Also, this function is named incorrectly, `maybe_` implies that it might or might not do something, this always returns a value. Prefer `summary` or `build_summary` or `extract_summary` or something like that.

Moosieus: I've renamed `maybe_summary` to `build_docs` to match the field name, and broke it into a few function clauses.

scohen (approved):

PR #650, opened 2024-03-17
Upgrade dialyzer for Elixir 1.16

scottming:
When I develop locally with Elixir 1.16, I often encounter this error. Upgrading can solve this problem.

```elixir
~/Code/lexical rename-module-v2 +3 !1 ❯ mix dialyzer                                     21h 8m 13s   25.3.2.7  1.16.0-otp-25
DEPENDENCY MISSING
------------------------
If you are reading this message, then Elixir and Erlang are installed but the
Erlang Dialyzer is not available. Probably this is because you installed Erlang
with your OS package manager and the Dialyzer package is separate.

On Debian/Ubuntu:

  `apt-get install erlang-dialyzer`

Fedora:

   `yum install erlang-dialyzer`

Arch and Homebrew include Dialyzer in their base erlang packages. Please report a Github
issue to add or correct distribution-specific information.
```


scohen:
As an aside, why are you using apt to install erlang or dialyzer or anything? That's a recipe for a miserable team; you should be using asdf or mise.

scohen (approved):

PR #651, opened 2024-03-19
Rename file when renaming module

scottming:
This PR is an implementation of renaming the file while renaming the module.

It first determines whether the module meets certain hard constraints, such as having no parent, then performs some conventional checks. If both are satisfied, it proceeds to return the file renaming.

https://github.com/lexical-lsp/lexical/assets/12830256/3d9d0221-fe3f-43be-8a51-3121b375d65d


scohen:
In the end, don't we just want the module name to match the file path? I was thinking this could be accomplished by indexing the file with the Module extractor, then, for each module, getting its name, and if it's not in the right file, doing one of the following:

If there is one module in the file:
1. If the file has the correct name, then do nothing
2. If the file name or path is incorrect, determine the right path and issue a RenameFIle, then emit edits for the file with the new contents.

If there is more than one module in the file:
1. Determine if any of the modules have the correct name
2. If one does, emit an edit for the current file with the others removed
3. For all modules in the current file that don't have the right name, follow the above steps.

There's a caveat here, since there can be global aliases / imports, etc. in the file, and if there are, they need to be relocated to the new files as well.

scohen:
Why do you think it's challenging?

scottming:
> Why do you think it's challenging?

Perhaps I was overthinking it at the time, feeling that there were too many conventions. However, if we start with some simple conventions, like in this PR, then the complexity should be manageable.

scohen:
Why do we care if there are parents or siblings before we rename a file?

apps/remote_control/lib/lexical/remote_control/code_mod/rename/file.ex
```diff
@@ -0,0 +1,100 @@
+defmodule Lexical.RemoteControl.CodeMod.Rename.File do
+  alias Lexical.Ast
+  alias Lexical.Document
+  alias Lexical.RemoteControl
+  alias Lexical.RemoteControl.Search.Store
+
+  def maybe_rename(entry, new_suffix) do
```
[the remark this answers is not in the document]
scottming: That means I need to pass the `project` down from `RemoteControl.Api.rename` all the way through. I'm not very keen on doing that; it doesn't seem to simplify the testing. prefer to your above suggestion.

apps/remote_control/lib/lexical/remote_control/code_mod/rename/document_changes.ex
```diff
@@ -0,0 +1,12 @@
+defmodule Lexical.RemoteControl.CodeMod.Rename.DocumentChanges do
```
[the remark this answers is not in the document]
scottming: Do you mean adding a 'rename_file' field to that structure? Then I need to use  `Document.Store.open_temporary(uri)` again to get the document when constructing it. Is that okay?

scottming:
> Why do we care if there are parents or siblings before we rename a file?

When there are some sibling modules at the first level, we usually don't use one of those modules as a mapping for the file name.

Regarding the issue of parent modules, consider examples like `state` modules inside `GenServer` modules.

apps/remote_control/lib/lexical/remote_control/code_mod/rename/document_changes.ex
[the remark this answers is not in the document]
scottming: how about this? @scohen 

https://github.com/lexical-lsp/lexical/pull/651/commits/4d4116805db09ad94a65a9e569fc0101d75d2c94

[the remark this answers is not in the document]
scohen: yes, one small change, make it a struct, so it's a little clearer

scohen:
> Regarding the issue of parent modules, consider examples like state modules inside GenServer modules.

This is why I was suggesting using the indexer, you can index the file with the `module` extractor, and see if there are multiple modules in the file (as well as get their hierarchy). If there are, you would treat the file as the parent module, and only rename that. 

There are multiple cases for this, of course. You can have a file whose name has no bearing on the modules contained inside as well.

apps/remote_control/lib/lexical/remote_control/code_mod/rename/document_changes.ex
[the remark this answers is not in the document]
scottming: Are you suggesting we change `rename_file` to a struct, rather than using a `tuple`?

scottming:
> This is why I was suggesting using the indexer, you can index the file with the module extractor, and see if there are multiple modules in the file (as well as get their hierarchy)

To be honest, I'm not sure exactly how to do it. It sounds like I need to add some fields to `Indexer.Entry`, right? And append this information when `Entry.block_definition` is called. Do you want me to take care of this in this PR?

scohen:
> It sounds like I need to add some fields to Indexer.Entry, right? 

No, I don't think so. You just need to run the indexer on the document, giving it only the `module` extractor, and it will return all the module definitions in the document. 
You can then look at the entries to find out if they're nested (you can get all the top-level modules by collecting those modules whose block_id is `:root`. any others are nested. 

to index using a subset of extractors, do the following
```elixir
document
modules =
  |> Indexer.Source.index([Indexer.Extractors.Module])
  |> Enum.reject(& &1.type == :metadata)
```

apps/remote_control/lib/lexical/remote_control/code_mod/rename/document_changes.ex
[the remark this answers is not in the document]
scohen: Yes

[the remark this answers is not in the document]
scottming: @scohen 2e6cb4c73db976178dfe10eb59be3795dd2b1c38

apps/remote_control/lib/lexical/remote_control/code_mod/rename/file.ex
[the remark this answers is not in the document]
scohen: it would simplify testing because you don't need to set something that's effectively global.

apps/remote_control/lib/lexical/remote_control/build.ex
```diff
@@ -81,17 +85,35 @@ defmodule Lexical.RemoteControl.Build do
...
 
+  @impl GenServer
+  def handle_cast({:compile, _force?}, %State{} = state) do
+    {:noreply, state}
+  end
+
+  @impl GenServer
+  def handle_cast(:suspend_compilation_temporarily, %State{} = state) do
+    new_state = State.disable_compilation(state)
+    Process.send_after(self(), :enable_compilation, 1000)
```
scohen, line 101: This seems very fraught and arbitrary. Won't it make sense to disable compilation until it's re-enabled by the thing that disabled it in the first place? 

```elixir
Build.disable_compilation fn -> 
  # compilation is disabled here
end)
```
seems much better.   

Furthermore, you've disabled compilation. I'm going to assume a couple of things have happened that would have changed the build state. You likely need to recompile the app after compilation resumes.

apps/remote_control/lib/lexical/remote_control/build/state.ex
```diff
@@ -15,7 +15,10 @@ defmodule Lexical.RemoteControl.Build.State do
 
   use RemoteControl.Progress
 
-  defstruct project: nil, build_number: 0, uri_to_source_and_edit_time: %{}
+  defstruct project: nil,
+            build_number: 0,
+            uri_to_source_and_edit_time: %{},
+            compilation_enabled?: true
```
scohen, line 21:
```suggestion
            enabled?: true
```

apps/remote_control/lib/lexical/remote_control/code_mod/rename/file.ex
```diff
@@ -0,0 +1,172 @@
...
+  alias Lexical.Ast
+  alias Lexical.Document
+  alias Lexical.ProcessCache
+  alias Lexical.Project
+  alias Lexical.RemoteControl
+  alias Lexical.RemoteControl.Search.Indexer
+  alias Lexical.RemoteControl.Search.Indexer.Entry
+
+  @spec maybe_rename(Entry.t(), String.t()) :: Document.Changes.rename_file()
+  def maybe_rename(entry, new_suffix) do
```
scohen, line 11:
```suggestion
  def maybe_rename(%Entry{} = entry, new_suffix) do
```

apps/remote_control/lib/lexical/remote_control/code_mod/rename/file.ex
```diff
@@ -0,0 +1,172 @@
...
+    uri = Document.Path.ensure_uri(entry.path)
+    text_edits = [Document.Edit.new(new_suffix, entry.range)]
+
+    with {:ok, document} <- Document.Store.open_temporary(uri),
+         {:ok, edited_document} =
+           Document.apply_content_changes(document, document.version + 1, text_edits),
+         {:ok, %{context: {:alias, alias}}} <-
+           Ast.surround_context(edited_document, entry.range.start) do
+      {:ok, to_string(alias)}
+    else
```
scohen, line 79: This feels incredibly wasteful to me, you're opening a new document, applying content edits to it, then parsing it just to get the new name of a file. Couldn't you do something similar by pulling using the range of the document and modifying that?

apps/remote_control/lib/lexical/remote_control/code_mod/rename/file.ex
```diff
@@ -0,0 +1,172 @@
...
+      _ -> :error
+    end
+  end
+
+  defp fetch_conventional_prefix(path) do
```
scohen, line 84: what makes a prefix "conventional"?

scohen, line 84: a comment here would be nice.

apps/remote_control/lib/lexical/remote_control/code_mod/rename/file.ex
```diff
@@ -0,0 +1,172 @@
...
+        remain, {elements, prefix} ->
+          {remain ++ elements, prefix}
+      end)
+
+    case result do
+      {_, []} ->
+        :error
+
+      {module_paths, prefix} ->
+        prefix = prefix |> Enum.reverse() |> Enum.join("/")
```
scohen, line 108:
```suggestion
        prefix = prefix |> Enum.reverse() |> Path.join()
```

apps/remote_control/lib/lexical/remote_control/code_mod/rename/file.ex
```diff
@@ -0,0 +1,172 @@
...
+
+        phoenix_live_view?(old_module_paths, new_name) ->
+          "live"
+
+        true ->
+          ""
+      end
+
+    suffix
+    |> String.split("/")
```
scohen, line 135:
```suggestion
    |> Path.split()
```

apps/remote_control/lib/lexical/remote_control/code_mod/rename/file.ex
```diff
@@ -0,0 +1,172 @@
...
+    |> List.insert_at(1, insertions)
+    |> Enum.reject(&(&1 == ""))
+    |> Enum.join("/")
```
scohen, line 138:
```suggestion
    |> Path.join()
```

apps/remote_control/lib/lexical/remote_control/code_mod/rename/file.ex
```diff
@@ -0,0 +1,172 @@
...
+  end
+
+  defp phoenix_component?(old_module_paths, new_name) do
+    under_components? = "components" in old_module_paths
+    component? = String.ends_with?(new_name, ["Components", "Layouts", "Component"])
```
scohen, line 143: i feel the same way about this heuristic as I did with controllers, there's no real indication that any of these are actually components, layouts or controllers. 
You can definitely check the modules though

apps/remote_control/lib/lexical/remote_control/code_mod/rename/file.ex
```diff
@@ -0,0 +1,172 @@
...
+        parent = Enum.at(new_name_list, -2)
+        local_module = Enum.at(new_name_list, -1)
+
+        # `LiveDemoWeb.SomeLive` or `LiveDemoWeb.SomeLive.Index`
+        String.ends_with?(parent, "Live") or String.ends_with?(local_module, "Live")
+      else
+        false
+      end
+
+    under_live_views? and live_view?
```
scohen, line 170: same comment about the heuristics here as well.

apps/remote_control/lib/lexical/remote_control/code_mod/rename/file.ex
```diff
@@ -0,0 +1,172 @@
...
+
+    under_controllers? and controller?
+  end
+
+  defp phoenix_live_view?(old_module_paths, new_name) do
+    under_live_views? = "live" in old_module_paths
+    new_name_list = String.split(new_name, ".")
+
+    live_view? =
+      if match?([_, _ | _], new_name_list) do
```
scohen, line 160: wat does ithis do? what is it testing for?

apps/remote_control/lib/lexical/remote_control/code_mod/rename/file.ex
```diff
@@ -0,0 +1,172 @@
...
+        parent = Enum.at(new_name_list, -2)
+        local_module = Enum.at(new_name_list, -1)
+
+        # `LiveDemoWeb.SomeLive` or `LiveDemoWeb.SomeLive.Index`
```
scohen, line 164: you can't nest live views more than two levels?

scohen (commented):
There are a bunch of hard-coded "/" characters here, this makes me very worried, as it's not portable.

scohen:
@scottming tried this again in both vscode and emacs (renaming `Lexical.Ast` to `Lexical.AST`, and both failed in different ways, but both brought down the language server.

apps/remote_control/lib/lexical/remote_control/commands/rename.ex
```diff
@@ -0,0 +1,117 @@
...
+  def start_link(_) do
+    GenServer.start_link(__MODULE__, %State{}, name: __MODULE__)
+  end
+
+  @impl true
+  def init(state) do
+    {:ok, state}
+  end
+
+  def set_rename_progress(uri_with_operation_counts) do
```
scottming, line 63: Note that when a large number of files are renamed, VSCode will delay sending some of the `DidClose` for old files and `DidOpen` for new files (for example, when renaming `Ast` -> `AST`, the last batch of messages will be received after 3-4 minutes). So, should we set a countdown here, for instance, reset the `uri_with_operation_counts` to `%{}` after 1 minute?

any ideas? @scohen

apps/remote_control/lib/lexical/remote_control/api.ex
```diff
@@ -13,6 +13,37 @@ defmodule Lexical.RemoteControl.Api do
...
+  @spec maybe_mark_closed(Project.t(), Lexical.uri()) :: :ok
+  def maybe_mark_closed(%Project{} = project, uri) do
+    if in_rename_progress?(project) do
+      RemoteControl.call(project, Commands.Rename, :mark_closed, [uri])
+    else
+      :ok
+    end
+  end
+
+  defp in_rename_progress?(%Project{} = project) do
```
scohen, line 43:
```suggestion
  defp rename_in_progress?(%Project{} = project) do
```

I'd also accept `rename_running?` to be more like `reindex`.

apps/remote_control/lib/lexical/remote_control/api.ex
```diff
@@ -13,6 +13,37 @@ defmodule Lexical.RemoteControl.Api do
...
+  def maybe_compile_document(project, document) do
+    if in_rename_progress?(project) do
+      RemoteControl.call(project, Commands.Rename, :mark_changed, [document.uri])
+    else
+      compile_document(project, document)
+    end
+  end
+
+  @spec maybe_mark_closed(Project.t(), Lexical.uri()) :: :ok
+  def maybe_mark_closed(%Project{} = project, uri) do
```
scohen, line 35: the names here need to be specific to renaming, like `mark_rename_closed`

apps/remote_control/lib/lexical/remote_control/api.ex
```diff
@@ -13,6 +13,37 @@ defmodule Lexical.RemoteControl.Api do
 
   require Logger
 
+  @spec maybe_schedule_compile(Project.t(), Lexical.uri()) :: :ok
+  def maybe_schedule_compile(project, triggered_file_uri) do
+    if in_rename_progress?(project) do
+      RemoteControl.call(project, Commands.Rename, :mark_saved, [triggered_file_uri])
+    else
+      schedule_compile(project, false)
+    end
```
scohen, line 22: why do these if statements need to happen here rather than on the remote_control side?

apps/remote_control/lib/lexical/remote_control/code_mod/format.ex
```diff
@@ -12,13 +13,22 @@ defmodule Lexical.RemoteControl.CodeMod.Format do
...
-    with :ok <- Build.compile_document(project, document),
+    with :ok <- ensure_not_in_rename_process(),
+         :ok <- Build.compile_document(project, document),
          {:ok, formatted} <- do_format(project, document) do
       edits = Diff.diff(document, formatted)
       {:ok, Changes.new(document, edits)}
     end
   end
 
+  defp ensure_not_in_rename_process do
```
scohen, line 24:
```suggestion
  defp ensure_not_renaming do
```

apps/remote_control/lib/lexical/remote_control/code_mod/rename.ex
```diff
@@ -11,12 +12,51 @@ defmodule Lexical.RemoteControl.CodeMod.Rename do
...
-      {:ok, rename_module.rename(range, new_name, entity)}
+      results = rename_module.rename(range, new_name, entity)
+      set_rename_progress(results, client_name)
+      {:ok, results}
     end
   end
+
+  defp set_rename_progress(document_changes_list, client_name) do
+    uri_with_operation_counts = uri_with_operation_counts(client_name, document_changes_list)
+    Commands.Rename.set_rename_progress(uri_with_operation_counts)
```
scohen, line 28:
```suggestion
    client_name
    |> uri_with_operation_counts(document_changes_list)
    |> Commands.Rename.set_rename_progress()
```

apps/remote_control/lib/lexical/remote_control/code_mod/rename.ex
```diff
@@ -11,12 +12,51 @@ defmodule Lexical.RemoteControl.CodeMod.Rename do
...
+  end
+
+  defp uri_with_operation_counts("Visual Studio Code", document_changes_list) do
```
scohen, line 31: this seems highly likely to change without any notice.

apps/remote_control/lib/lexical/remote_control/code_mod/rename/file.ex
scottming: Perhaps we should pass the `document` params in  `maybe_rename` function, as this would simplify many things.

apps/remote_control/lib/lexical/remote_control/api.ex
scottming: Because the Build's API, such as `compile_document`, is called using `RemoteControl.call`, do you have a better suggestion?

apps/remote_control/lib/lexical/remote_control/code_mod/rename/file.ex
scottming: added

apps/remote_control/lib/lexical/remote_control/api/messages.ex
```diff
@@ -15,6 +15,8 @@ defmodule Lexical.RemoteControl.Api.Messages do
 
   defrecord :file_changed, uri: nil, from_version: nil, to_version: nil, open?: false
 
+  defrecord :file_opened, uri: nil, version: nil
```
scohen, line 18: don't forget to add types for this too.

apps/remote_control/lib/lexical/remote_control/code_mod/format.ex
```diff
@@ -5,20 +5,30 @@ defmodule Lexical.RemoteControl.CodeMod.Format do
...
+         :ok <- Build.compile_document(project, document),
          {:ok, formatted} <- do_format(project, document) do
       edits = Diff.diff(document, formatted)
       {:ok, Changes.new(document, edits)}
     end
   end
 
+  defp ensure_not_renaming do
+    if Rename.in_progress?() do
+      {:error, "Cannot format while renaming"}
```
scohen, line 26: `{:error, :rename_in_progress}` is clear enough

scohen:
I tried this branch this weekend. Renamed `Lexical.Ast` to `Lexical.AST`.

Fails pretty hard in emacs. Lots of errors about documents not being open. I can get you stack traces if you want
 
Works in vscode, but I see a bunch of errors like the following:
```
17:33:37.046 [error] Formatter failed "Cannot format while renaming"
17:33:37.046 [error] Formatter failed "Cannot format while renaming"
17:33:37.046 [error] Formatter failed "Cannot format while renaming"
17:33:37.046 [error] Formatter failed "Cannot format while renaming"
17:33:37.046 [error] Formatter failed "Cannot format while renaming"
17:33:37.046 [error] Formatter failed "Cannot format while renaming"
17:33:37.046 [error] Formatter failed "Cannot format while renaming"
17:33:37.047 [error] Formatter failed "Cannot format while renaming"
17:33:37.047 [error] Formatter failed "Cannot format while renaming"
17:33:37.047 [error] Formatter failed "Cannot format while renaming"
17:33:37.048 [error] Formatter failed "Cannot format while renaming"
17:33:37.048 [error] Formatter failed "Cannot format while renaming"
17:33:37.048 [error] Formatter failed "Cannot format while renaming"
```
The formatter should detect these errors, but you've made it tougher by making the reason a string.

scohen:
Here are the errors I encountered in emacs:
```
2024-04-15T17:29:32.801242-07:00 error: Error in process <0.4908.5> on node 'project-lexical-29783@127.0.0.1' with exit value:, {#{'__exception__' => true,'__struct__' => 'Elixir.Module.Types.Error',message => <<"found error while checking  (UndefinedFunctionError) function Lexical.Ast.Analysis.__struct__/0 is undefined (module Lexical.Ast.Analysis is not available)">>},[
  {'Elixir.Lexical.Ast.Analysis','__struct__',[],[]},
{'Elixir.Module.Types.Of',struct,3,[{file,"lib/module/types/of.ex"},{line,131}]},
{'Elixir.Module.Types.Pattern',of_shared,4,[{file,"lib/module/types/pattern.ex"},{line,754}]},{'Elixir.Module.Types.Pattern',of_pattern,3,[{file,"lib/module/types/pattern.ex"},{line,47}]},{'Elixir.Module.Types.Helpers',do_map_reduce_ok,3,[{file,"lib/module/types/helpers.ex"},{line,93}]},{'Elixir.Module.Types.Pattern',of_head,4,[{file,"lib/module/types/pattern.ex"},{line,12}]},{'Elixir.Module.Types',warnings_from_clause,6,[{file,"lib/module/types.ex"},{line,62}]},{'Elixir.Module.Types','-warnings/5-fun-0-',8,[{file,"lib/module/types.ex"},{line,22}]}]}
```
Lots of these: 
```
17:29:33.161 [info] No formatter found in /Users/steve/Projects/lexical/apps/common/test/lexical/ast/detection/alias_test.exs error was false
17:29:33.162 [info] No formatter found in /Users/steve/Projects/lexical/apps/common/test/lexical/ast/detection error was false
17:29:33.165 [info] No formatter found in /Users/steve/Projects/lexical/apps/common/test/lexical/ast error was false
17:29:33.165 [info] No formatter found in /Users/steve/Projects/lexical/apps/common/test/lexical error was false
17:29:33.166 [info] No formatter found in /Users/steve/Projects/lexical/apps/common/test error was false
17:29:33.169 [info] found formatter in /Users/steve/Projects/lexical/apps/common
```

Then TONS of these:
```
18:12:48.311 [error] Failed to handle Elixir.LXical.Protocol.Notifications.DidClose, {{:error, :not_open}, %LXical.Server.State{configuration: %LXical.Server.Configuration{project: %LXical.Project{root_uri: "file:///Users/steve/Projects/lexical", mix_exs_uri: "file:///Users/steve/Projects/lexical/mix.exs", mix_project?: true, mix_env: nil, mix_target: nil, env_variables: %{}, project_module: nil, entropy: 4195, project_config: []}, support: %LXical.Server.Configuration.Support{code_action_dynamic_registration: true, hierarchical_symbols: true, snippet: true, deprecated: true, tags: false, signature_help: #Protocol.Types.SignatureHelp.ClientCapabilities<[dynamic_registration: true, signature_information: #Protocol.Types.SignatureHelp.ClientCapabilities.SignatureInformation<[parameter_information: #Protocol.Types.SignatureHelp.ClientCapabilities.ParameterInformation<[label_offset_support: true]>]>]>, work_done_progress: true}, client_name: "emacs", additional_watched_extensions: nil, dialyzer_enabled?: false}, initialized?: true, shutdown_received?: false, in_flight_requests: %{"12" => {%LXical.Protocol.Requests.CodeLensRefresh{lsp: %LXical.Protocol.Requests.CodeLensRefresh.LSP{id: "12", jsonrpc: "2.0", method: "workspace/codeLens/refresh"}, id: "12", jsonrpc: "2.0", method: "workspace/codeLens/refresh"}, #Function<0.7576677/2 in LXical.Server.server_request/1>}}}}
18:12:48.311 [error] Could not handle message LXical.Protocol.Notifications.DidClose :ok
```

I also notice that you're emitting file renames even if the file names haven't changed. This could also be messing stuff. up. 
Examples:

```
    {
      "kind": "rename",
      "newUri": "file:///Users/steve/Projects/lexical/apps/common/lib/lexical/ast/detection/pipe.ex",
      "oldUri": "file:///Users/steve/Projects/lexical/apps/common/lib/lexical/ast/detection/pipe.ex",
      "options": {
        "overwrite": true
      }
    },
    {
      "kind": "rename",
      "newUri": "file:///Users/steve/Projects/lexical/apps/common/lib/lexical/ast/detection/spec.ex",
      "oldUri": "file:///Users/steve/Projects/lexical/apps/common/lib/lexical/ast/detection/spec.ex",
      "options": {
        "overwrite": true
      }
    },
```
I think these are unnecessary, and will likely cause emacs to freak out.

Emacs then opens all the documents and does code lens requests on them, followed by a bunch of didOpen requests, followed by didChange requests. I then start seeing cancellation requests for the codeLens requests intermixed with the didChange notifications. It appears that emacs sends the cancellation request immediately after the code lens request.

Emacs does seem to be requesting notifications in the following order: `didOpen` -> `didChange` -> `didSave` -> `didClose` If this is the case, why do we need to open things temporarily?

scottming:
> I also notice that you're emitting file renames even if the file names haven't changed. This could also be messing stuff. up.
Good catch, I'll change it tomorrow.

> Emacs does seem to be requesting notifications in the following order: didOpen -> didChange -> didSave -> didClose If this is the case, why do we need to open things temporarily?


So, are you suggesting that in this PR, we should replace 'Store.open_temporarily' with 'Store.open' everywhere?

scohen:
> So, are you suggesting that in this PR, we should replace 'Store.open_temporarily' with 'Store.open' everywhere?

Logically, all editors should be doing the same actions, so yes. 
Are they not?

Issue #643, opened 2024-03-07
New completion sorting approach


Moosieus:
Here's a brief completion sorting proposal I've made up:

`"00"`: local variables
`"01"`: local declarations (declared within the immediate module or `use`)
`"10"`: remote declarations (from existing invocations of `alias`, `require`, and `import`)
`"11"`: global declarations (def, do/end, if/do/else/end, SpecialForms, etc)
`"20"`: auto declarations (short for auto-import added by the LSP)

variables = variables. Made relatively easy by Elixir's scoping rules.
declarations = functions, attributes, atoms(?), and all classes of macros (macros, guards, structs, protocols, etc).

- The first digit roughly categorizes the *"distance"* at which something's defined, for lack of better term. Stuff most immediate to the user's work is sorted first.
- The second digit roughly categorizes the taxonomy of a completion *(left very general at the moment)*.
- Everything within the same "taxonomy" is then sorted by name.
- Every completion falls into one of these categories, and there's no way to boost things ad-hoc.

This is mainly inspired by the way TypeScript (tsserver) handles its completion suggestions, which are taxonomized and sorted as:
```ts
LocalDeclarationPriority: "10" as SortText,
LocationPriority: "11" as SortText,
OptionalMember: "12" as SortText,
MemberDeclaredBySpreadAssignment: "13" as SortText,
SuggestedClassMembers: "14" as SortText,
GlobalsOrKeywords: "15" as SortText,
AutoImportSuggestions: "16" as SortText,
ClassMemberSnippets: "17" as SortText,
JavascriptIdentifiers: "18" as SortT
```
[(source)](https://github.com/microsoft/TypeScript/blob/6357c89c52c86f7c82ea17c3660efcc2a1a12915/src/services/completions.ts#L406)

I like what TS does here b/c:
- It seems to work well and satisfy TS developers.
- It's easily understood and maintained.
- Works when factoring in jsDoc and JSX (workalikes to typespecs and heex).

I think the approach works well for Elixir and Lexical b/c:
- Elixir has even fewer constructs to sort.
- New levels can be easily defined and existing ones reordered.

Some personal notes:
- TS shows deprecated completions but sorts them last. I'd refrain from showing them at all.
- This approach concerns fuzzy matching/searching exclusively to filtering, rather than *filtering and sorting*.
  - That means *'I boosted x by 5% but that ruined sorting somewhere else'* isn't a problem.
[...]

PR #653, opened 2024-03-20
New completion sorting approach

Moosieus:
Addressing #643. This PR replaces the completion 'boost' function `set_sort_scope` and an enumerated set of options, based on relative "distance" (for lack of better term) to a user's locally opened module.

At time of writing, I'm definitely looking for feedback and further improvements before this is merged.

I've a few immediate things I'm considering for improvement:
* I chose the new naming conventions (ex: `SortScope`, `set_sort_scope`) on the spot. Perhaps there's better names?
* Local sorting is relatively ad-hoc. Should it be more formally defined?
* I wonder if Module/Behaviour completion deserves its own category and local category?


apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
```diff
@@ -118,11 +128,35 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Callable do
 
   @default_functions ["module_info", "behaviour_info"]
 
-  defp maybe_boost(item, %_{name: name}, default_boost \\ 5) do
-    if String.starts_with?(name, "__") or name in @default_functions do
-      item
-    else
-      Builder.boost(item, default_boost)
+  defp maybe_boost(item, callable, %Env{position_module: position_module}) do
```
scohen, line 131: nit: i prefer using function-level pattern matching to make decisions about which function head is called, and not to extract fields. 
```suggestion
  defp maybe_boost(item, callable, %Env{}) do
  position_module = env.position_module
```

apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
```diff
@@ -118,11 +128,35 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Callable do
...
+
+    local_priority =
+      cond do
+        dunder? -> 9
+        callback? -> 8
+        true -> 1
+      end
+
+    cond do
+      origin === "Kernel" or origin === "Kernel.SpecialForms" or name in @default_functions ->
```
scohen, line 151: what if you've imported a module or function? Will it be local?

apps/server/test/lexical/server/code_intelligence/completion/builder_test.exs
```diff
@@ -34,41 +35,35 @@ defmodule Lexical.Server.CodeIntelligence.Completion.BuilderTest do
...
+    test "scope order follows variable -> local -> remote -> global -> auto -> default" do
+      i = set_sort_scope(item("g"), SortScope.module())
+      ii = set_sort_scope(item("f"), SortScope.variable())
+      iii = set_sort_scope(item("e"), SortScope.local())
+      iv = set_sort_scope(item("d"), SortScope.remote())
+      v = set_sort_scope(item("c"), SortScope.global())
+      vi = set_sort_scope(item("b"), SortScope.auto())
+      vii = set_sort_scope(item("a"), SortScope.default())
+
+      assert [^i, ^ii, ^iii, ^iv, ^v, ^vi, ^vii] = sort_items([vii, vi, v, iv, iii, ii, i])
```
scohen, line 48: I can't tell if I like this or not, but it's exceedingly concise and clever.

scohen (commented):
I haven't had a chance to play around with this very much, but from my cursory play, it seems much improved. 

I especially like the `SortScope` idea. Giving names to things works very well.

Going to hold off on approving until I get more time with it.

scohen:
Suggestions: 
  * It would be _exceedingly cool_ if we had different ideas of `local`. For example, functions defined in the current module are more local than functions that have been imported. 
  * Test cases should not be completed ever. Currently, they're coming in above snippets like defining new tests. I think they were completed before, but were ranked lower than the `test` snippets.

apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
Moosieus: Completions from imports are sorted as remote for now, so they're below local functions, macros, and callbacks, but before globally available callables.

I mostly did this out of convenience, and that the results seemed serviceable. I found it was easy to identify global and local functions, and have other callable completions default to remote.

I could see a case for imported stuff being local, but I think the results look nice now.

apps/server/test/lexical/server/code_intelligence/completion/builder_test.exs
Moosieus: I debated keeping the `alpha_first`/`alpha_second` naming scheme, but I felt this looked nicer, particularly keeping the assert on a single line.

apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
Moosieus: Fixed!

apps/server/lib/lexical/server/code_intelligence/completion/translations/callable.ex
scohen: I think this sort order is good. I was worried they'd be considered local. 
My thoughts here have been "make the closest things have higher priority", and this does that.

apps/server/test/lexical/server/code_intelligence/completion/builder_test.exs
scohen: Like i said, clever, though it might be difficult for those not exposed to roman numerals (@scottming ?) to understand.

scohen (changes requested):
This looks very good, and is a welcome improvement. 
My one ask is to remove test cases from completion. After that, it's good to go.

scohen (approved):

PR #658, opened 2024-03-25
For module definitions, use the `Indexer` version instead of `ElixirSense`

### LINKED ISSUE #649: Improve the precision of `go to module definition`. When the cursor is on the parent, it should jump to the parent module.
For example: 

`Lexical.|RemoteControl.Search` should go to `RemoteControl` not `RemoteControl.Search`

scottming:
Fixes #649

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
@@ -10,6 +10,7 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Definition do
   alias Lexical.RemoteControl.Search.Store
   alias Lexical.Text
 
+  @spec definition(Document.t(), Position.t()) :: {:ok, [Location.t()]} | {:error, String.t()}
   def definition(%Document{} = document, %Position{} = position) do
     with {:ok, _, analysis} <- Document.Store.fetch(document.uri, :analysis),
          {:ok, entity, _range} <- Entity.resolve(analysis, position) do
@@ -21,9 +22,23 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Definition do
        when type in [:struct, :module] do
     module = Formats.module(entity)
 
-    case Store.exact(module, type: type, subtype: :definition) do
-      [entry] -> to_location(entry)
-      _other -> :error
+    locations =
+      for entry <- Store.exact(module, type: type, subtype: :definition),
+          result = to_location(entry),
+          match?({:ok, _}, result) do
+        {:ok, location} = result
+        location
+      end
+
+    case locations do
+      [] ->
+        {:error, "No definition found for #{inspect(module)}"}
+
+      [location] ->
+        {:ok, location}
+
+      _ ->
+        {:ok, locations}
     end
   end
 
--- a/apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
@@ -1,17 +1,39 @@
 defmodule Lexical.RemoteControl.CodeIntelligence.Definition do
   alias Future.Code, as: Code
   alias Lexical.Ast
+  alias Lexical.Ast.Analysis
   alias Lexical.Document
   alias Lexical.Document.Location
   alias Lexical.Document.Position
+  alias Lexical.Formats
   alias Lexical.RemoteControl.CodeIntelligence.Entity
+  alias Lexical.RemoteControl.Search.Store
   alias Lexical.Text
 
-  def definition(%Document{} = document, %Position{} = position) do
-    document
+  def definition(%Analysis{} = analysis, %Position{} = position) do
+    with {:ok, entity, _range} <- Entity.resolve(analysis, position) do
+      fetch_definition(entity, analysis, position)
+    end
+  end
+
+  defp fetch_definition({type, entity}, %Analysis{} = _analysis, %Position{} = _position)
+       when type in [:struct, :module] do
+    module = Formats.module(entity)
+
+    case Store.exact(module, type: type, subtype: :definition) do
+      [entity] ->
+        to_location(entity)
+
+      [] ->
+        {:ok, nil}
+    end
+  end
+
+  defp fetch_definition(_, %Analysis{} = analysis, %Position{} = position) do
+    analysis.document
     |> Document.to_string()
     |> ElixirSense.definition(position.line, position.character)
-    |> parse_location(document)
+    |> parse_location(analysis.document)
   end
 
   defp parse_location(%ElixirSense.Location{} = location, document) do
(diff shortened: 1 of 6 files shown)


apps/server/lib/lexical/server/provider/handlers/go_to_definition.ex
```diff
@@ -1,18 +1,21 @@
...
   alias Lexical.RemoteControl
 
   require Logger
 
   def handle(%GoToDefinition{} = request, env) do
-    case RemoteControl.Api.definition(env.project, request.document, request.position) do
-      {:ok, native_location} ->
-        {:reply, Responses.GoToDefinition.new(request.id, native_location)}
-
+    with {:ok, _document, %Ast.Analysis{valid?: true} = analysis} <-
```
scottming, line 11: This means that when the document can't be analyzed, we can't jump. Do we have any other options?

scohen: fall back to elixir_sense?

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -1,17 +1,39 @@
...
+
+    case Store.exact(module, type: type, subtype: :definition) do
+      [entity] ->
+        to_location(entity)
+
+      [] ->
+        {:ok, nil}
+    end
+  end
+
```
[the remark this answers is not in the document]
scottming: I don't want to for now.

apps/server/lib/lexical/server/provider/handlers/go_to_definition.ex
scottming: Fixed the issue using the `Store.fetch` method.

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -1,17 +1,40 @@
...
+  defp fetch_definition({type, entity}, %Analysis{} = _analysis, %Position{} = _position)
+       when type in [:struct, :module] do
+    module = Formats.module(entity)
+
+    with [entry] <- Store.exact(module, type: type, subtype: :definition),
+         {:ok, location} <- to_location(entry) do
+      {:ok, location}
+    else
+      _ ->
+        {:ok, nil}
```
scohen, line 29: Shouldn't this be :error?

scottming: Only if the index doesn't take effect will it reach here, though an error also makes sense.

Changed.

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -310,12 +350,28 @@ defmodule Lexical.RemoteControl.CodeIntelligence.DefinitionTest do
...
+        text = File.read!(path)
+        {path, text}
+      else
+        {document.path, Document.to_string(document)}
+      end
+
+    with {:ok, entries} <- Search.Indexer.Source.index(path, text) do
+      Search.Store.replace(entries)
+    end
+  end
```
scohen, line 376: why read directly off the filesystem when we have the document store? Can't you use `open_temporarily`?

scottming: Ok, changed

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -310,12 +358,27 @@ defmodule Lexical.RemoteControl.CodeIntelligence.DefinitionTest do
     end
   end
 
-  defp definition(project, code) do
+  defp definition(project, code, referenced_uri \\ nil) do
     with {position, code} <- pop_cursor(code),
          {:ok, document} <- subject_module(project, code),
+         :ok <- index(document, referenced_uri),
```
scohen, line 364: this is super confusing. Sometimes you're indexing the subject, other times you're indexing the referenced file, which doesn't change. Shouldn't you _always_ index the referenced file and the subject?  I think that would make things a lot clearer and easier.

apps/server/test/lexical/server/provider/handlers/go_to_definition_test.exs
```diff
@@ -61,11 +61,6 @@ defmodule Lexical.Server.Provider.Handlers.GoToDefinitionTest do
       {:ok, request} = build_request(uses_file_path, 4, 17)
 
       {:reply, %{result: %Location{} = location}} = handle(request, project)
-
-      assert location.range.start.line == 15
-      assert location.range.start.character == 7
-      assert location.range.end.line == 15
-      assert location.range.end.character == 12
```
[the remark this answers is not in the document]
scottming: This was actually learned from `find_reference_test.exs`: https://github.com/lexical-lsp/lexical/blob/f8ef46b81b74b7a129ce17710f63826999b5b31a/apps/server/test/lexical/server/provider/handlers/find_references_test.exs#L70

Additionally, I feel that verifying these rows and columns doesn't really make much sense. It's sufficient to verify that the data structure is returned. 
Precise row and column verification has already been done in the `remote_control` app. Moreover, since these are transformed LSP rows and columns, they can't be directly used with the `decorate` function in `RangeSupport`.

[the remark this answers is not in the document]
scohen: Yeah, you know i'm not in favor of those kinds of assertions.

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -1,17 +1,36 @@
...
+         {:ok, entity, _range} <- Entity.resolve(analysis, position) do
+      fetch_definition(entity, analysis, position)
+    end
+  end
+
+  defp fetch_definition({type, entity}, %Analysis{} = _analysis, %Position{} = _position)
+       when type in [:struct, :module] do
+    module = Formats.module(entity)
+
+    with [entry] <- Store.exact(module, type: type, subtype: :definition) do
```
scohen, line 24: note: `Store.exact` can theoretically return multiple entries, which will be returned here instead of the ok/error tuple. It's unlikely to happen with modules, but there's nothing that states that you cannot name a module the same as another existing module on disk. 

prefer to swap this out with a case
```elixir
case Store.exact(module, type: type, subtype: :definition) do
  [entry] -> to_location(entry)
  _other -> {:error, :multiple_results}
end
```

scottming: Theoretically speaking, such a situation could indeed occur.

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -1,17 +1,37 @@
...
+    end
+  end
+
+  defp fetch_definition({type, entity}, %Analysis{} = _analysis, %Position{} = _position)
+       when type in [:struct, :module] do
+    module = Formats.module(entity)
+
+    case Store.exact(module, type: type, subtype: :definition) do
+      [entry] -> to_location(entry)
+      _other -> :error
```
scohen, line 26: I think, per our discussion, this is also fine:
```elixir
module
|> Store.exact(type: type, subtype: :definition)
|> Enum.map(&to_location/1)
```

scohen (approved):

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
scottming: CleanShot 2024-04-12 at 16 10 22@2x
Feel free to try the last commit: 57c82f462e7ff6d67abfdf19f4ecf3a3932c6f7e, if no more issues, then I'll merge it.

PR #659, opened 2024-03-25
Support resolving Phoenix Controller modules.

scottming:
ElixirSense supports this, so I think we should too.


apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -190,6 +193,37 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
...
+    scope_path = Future.Macro.path(path, &match?({:scope, _, [_ | _]}, &1))
+
+    if scope_path do
+      {:scope, _, [_, {:__aliases__, _, segments} | _]} = hd(scope_path)
+      segments
+    end
+  end
+
+  defp controller_module_at_cursor?(charlist) do
+    charlist |> last_module_at_position() |> String.ends_with?("Controller")
```
scohen, line 220: this heuristic isn't particularly good, lots of things can end with `Controller` and not be phoenix controllers. We have access to these modules and should check for certain functions inside them before deeming them to be controllers.

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -190,6 +193,37 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
...
+    path = Ast.path_at(analysis, position)
+    scope_path = Future.Macro.path(path, &match?({:scope, _, [_ | _]}, &1))
+
+    if scope_path do
+      {:scope, _, [_, {:__aliases__, _, segments} | _]} = hd(scope_path)
+      segments
+    end
+  end
+
+  defp controller_module_at_cursor?(charlist) do
```
scohen, line 219: nit: `defp phoenix_controller_module?`

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -190,6 +193,37 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
     end
   end
 
+  defp maybe_pad_phoenix_scope_module(module_string, charlist, analysis, position) do
+    with true <- controller_module_at_cursor?(charlist),
+         scope_segments = phoenix_scope_segments(analysis, position),
+         false <- is_nil(scope_segments),
+         {:ok, module} <-
+           RemoteControl.Analyzer.expand_alias(scope_segments, analysis, position) do
+      [module, module_string] |> Module.concat() |> inspect()
```
scohen, line 202: rather than `inspect()` I think you want `Formats.module`

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -190,6 +193,37 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
...
+      {:scope, _, [_, {:__aliases__, _, segments} | _]} = hd(scope_path)
+      segments
+    end
+  end
+
+  defp controller_module_at_cursor?(charlist) do
+    charlist |> last_module_at_position() |> String.ends_with?("Controller")
+  end
+
+  defp last_module_at_position(charlist) do
```
scohen, line 223: the splitting part of this function should probably go in `module.ex` in common

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
scohen: So, rather than look for a trailing `Controller` find some of these functions

https://hexdocs.pm/phoenix/Phoenix.Controller.html#functions

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -190,6 +193,37 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
...
+         {:ok, module} <-
+           RemoteControl.Analyzer.expand_alias(scope_segments, analysis, position) do
+      [module, module_string] |> Module.concat() |> inspect()
+    else
+      _ ->
+        module_string
+    end
+  end
+
+  defp phoenix_scope_segments(analysis, position) do
```
scohen, line 209: what is this doing?

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -190,6 +193,37 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
     end
   end
 
+  defp maybe_pad_phoenix_scope_module(module_string, charlist, analysis, position) do
+    with true <- controller_module_at_cursor?(charlist),
+         scope_segments = phoenix_scope_segments(analysis, position),
+         false <- is_nil(scope_segments),
```
scohen, line 199:
```suggestion
         {:ok, scope_segments} <- fetch_phoenix_scope_segments(analysis, position),
```

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -190,6 +193,37 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
...
+  end
+
+  defp phoenix_scope_segments(analysis, position) do
+    path = Ast.path_at(analysis, position)
+    scope_path = Future.Macro.path(path, &match?({:scope, _, [_ | _]}, &1))
+
+    if scope_path do
+      {:scope, _, [_, {:__aliases__, _, segments} | _]} = hd(scope_path)
+      segments
+    end
```
scohen, line 216: case scope_path 
```suggestion
    case scope_path do
      [{:scope, _, [_, {:__aliases__, _, segments} | _]} | _] -> {:ok, segments}
      _ -> :error
      segments
    end
```

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -190,6 +193,37 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
     end
   end
 
+  defp maybe_pad_phoenix_scope_module(module_string, charlist, analysis, position) do
```
scohen, line 196:
```suggestion
  defp maybe_prepend_phoenix_scope_module(module_string, charlist, analysis, position) do
```

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
scottming: Added some docs.

    # fetch the alias segments from the `scope` macro
    # e.g. `scope "/foo", FooWeb.Controllers`
    # the alias module is `FooWeb.Controllers`, and the segments is `[:FooWeb, :Controllers]`

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
scottming: > So, rather than look for a trailing Controller find some of these functions

But we can't do this before we expand the `alias` of the scope. Since we have `fetch_phoenix_scope_xx`for further checks, then is doing an `ends_with` check enough? 

Of course, we can rename this function to make it clearer.

scohen: You can do that, get the expected module. Then see if it has one of those functions exported

scottming: > You can do that, get the expected module.

What I mean is, before I know the `alias` of the scope, I can't obtain the specific module. The current module is a `Router` module, not a certain `Controller`.

scohen: Don't you have the pieces in the parent scope and the current route?

scottming: Okay, I think I understand what you mean now. You're suggesting to do this check **after** the fetch. I'll make some modifications in the next few days.

Additionally, I think we also need to consider the issues with *nested scopes* and the `LiveView` modules.

scottming: Additionally, regarding this kind of function_exported check, do you have any suggestions for testing? It seems I can't simply use a patch to simulate it.

scohen: Can you not patch `Kernel.function_exported?/2`

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -199,6 +204,57 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
...
+    else
+      _ ->
+        module_string
+    end
+  end
+
+  defp fetch_phoenix_scope_alias_segments(analysis, position) do
+    # fetch the alias segments from the `scope` macro
+    # e.g. `scope "/foo", FooWeb.Controllers`
+    # the alias module is `FooWeb.Controllers`, and the segments is `[:FooWeb, :Controllers]`
```
scohen, line 224: Fantastic comment.

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -199,6 +204,57 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
...
+    path =
+      analysis
+      |> Ast.cursor_path(position)
+      |> Enum.filter(&match?({:scope, _, [_ | _]}, &1))
+      # There might be nested `scope` macros, we need the ancestor one
```
scohen, line 229:
```suggestion
      # There might be nested `scope` macros, we need the immediate ancestor
```

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -397,4 +453,8 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
       _ -> :error
     end
   end
+
+  defp function_exported_check?(module, function, arity) do
```
scohen, line 457: nit: just call this `function_exported?` and add a comment indicating that it's here for testing.

scohen (approved):
This got a lot less complicated. Very nice.
Address the comment and you're good to go.

Moosieus (approved):

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
scottming: I initially thought the same, but this would conflict with Kernel.function_exported? and cause the compilation to fail. Do you have any other name suggestions?

scohen: `exported?`

PR #661, opened 2024-03-25
FIX: Invalid reads for requests that contain multi-byte characters

scohen:
We would get sporadic failures where the transport would keep reading headers after they were done. This was due to the previous request containing unicode characters at specific locations, which would confuse the transport, since it was reading the body of the request with IO.read, which reads _characters_, but it was passing in the content-length, which is counted in bytes. This would cause us to over-read the body, and consume part of the next body as we were reading the headers, and then the whole thing would fail when we tried to parse the headers.

The change is simple, switch to `binread`, which takes the number of bytes to read.

### DIFF
--- a/apps/server/lib/lexical/server/transport/std_io.ex
+++ b/apps/server/lib/lexical/server/transport/std_io.ex
@@ -69,7 +69,7 @@ defmodule Lexical.Server.Transport.StdIO do
         loop([], device, callback)
 
       :eof ->
-        maybe_halt()
+        maybe_stop()
 
       line ->
         loop([line | buffer], device, callback)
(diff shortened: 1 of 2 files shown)


Moosieus (approved):

Moosieus:
Looks good aside from the Dialyzer complaining about no return. Is there a good reason for using `System.halt/0` over its more graceful `System.stop/0` counterpart?

scohen:
>  Is there a good reason for using System.halt/0 over its more graceful System.stop/0 counterpart?

Nope, changed it.

Moosieus (approved):

Issue #639, opened 2024-03-04
Undefined function in .html.heex files


crbelaus:
I am noticing a weird behaviour on Heex files. When I open any file and save I get a error on the first function call saying:

```
undefined function WHATEVER_FUNCTION (there is no such import)
```

I can consistently replicate this behaviour on a brand new Phoenix project. I ran the following command to create the project:

```
mix phx.new --no-ecto example
```

Then, navigated to the project root, ran `mix do deps.get, compile` and confirmed that it compiles cleanly.

```
$ mix compile --force --warnings-as-errors
Compiling 14 files (.ex)
Generated example app
```

When I open `lib/example_web/controllers/page_html/home.html.heex` in my editor (VSCode) and just save it I get the error mentioned above. It is interesting that the LSP works fine regardless including formatter, autocompletion and whatnot.

![image](https://github.com/lexical-lsp/lexical/assets/2965623/01c38e89-d681-45c5-80ba-c910cc91ae38)

Not sure if this is a problem in Lexical or in the VSCode extension, so feel free to move the issue where it is more appropriate.

crbelaus:
Just want to add that, if I call any function above the first error it disappears and shows in the new function call instead. For example:

![image](https://github.com/lexical-lsp/lexical/assets/2965623/19e62619-3e8b-4c65-a115-4100c89cae68)

crbelaus:
In case anyone has the same problem. This has been discussed on #495 where it is explained that the autoformatting is causing this problem.

When you save the file it triggers a project recompilation. Then, the autoformat triggers a file recompilation which doesn't have enough information to figure out what functions have been imported in the HEEx template, causing the problems described above.

The immediate remedy to avoid this problems is to disable autoformat in your code editor.

scohen:
Fixed by #662

Moosieus:
As-you-type completions are disabled in heex for now, until we can create the environment scottming indicated [here](https://github.com/lexical-lsp/lexical/issues/495#issuecomment-1828853060).

Issue #547, opened 2024-01-05
Undefined function gettext/1 in `app.html.heex`


Terbium-135:
While editing `app.html.heex` (I guess this will happen editing any heex file) an error is shown:

![gettext-error](https://github.com/lexical-lsp/lexical/assets/29901271/85452c61-7657-4d6c-829d-438f65dd5731)

After a recompile the error is gone.
Reporting this more or less for completeness

scottming:
@Terbium-135 Once you save the file, the error diagnostics will disappear, right?

I think this is a known issue, https://github.com/lexical-lsp/lexical/issues/495#issuecomment-1828853060

Terbium-135:
Aye, the diagnostics are disappearing on a save.

scohen:
Duplicate of #495

Issue #495, opened 2023-11-19
Undefined function errors in heex file


bangalcat:
Hi, thank you for all of your efforts on this awesome lsp.

I'm not sure that lexical fully supports heex files, though I'd like to report this for the records.

When I use function components by import, lexical diagnoses it as compile error:

<img width="837" alt="image" src="https://github.com/lexical-lsp/lexical/assets/16166083/3109c372-1f90-48d2-b176-0cd7a2217062">

If I use it with the module name, it doesn't emit error messages:

<img width="831" alt="image" src="https://github.com/lexical-lsp/lexical/assets/16166083/9abbc08c-ddcb-4740-98cc-8560659e4035">

I guess this occurs all functions without module name in heex file:

<img width="1041" alt="image" src="https://github.com/lexical-lsp/lexical/assets/16166083/afebb502-0cbf-4b28-903f-4d8ccaf6e6e4">

But mix compile works well anyway.

### My Environments

- lexical build: elixir 1.15.7 / otp 26.1.2 / on latest commit
- project : elixir 1.15.7 / otp 26.1.2
- MacOS Sonoma, using neovim, asdf

scottming:
Hi, @bangalcat Heex is not fully supported, and this behavior is expected. After saving, this error should disappear. Currently, it is difficult to determine which `ex/heex` files are referenced by a `heex` file while you are editing it. We will further optimize the compilation diagnostics feature for `heex` once indexing is completed.

bangalcat:
@scottming thanks for the quick response.

> After saving, this error should disappear. 

Hmm, as for me, the errors didn't disappear after saving. When I just open the file, the errors don't show, but when I start typing or saving the file, the errors appear. Am I missing something?

scottming:
What happens when you run `mix compile` in the terminal? Essentially, when you save a file in Lexical, it parses the diagnostics generated by `mix compile`.

bangalcat:
It passes without any error or warning messages. I've also tried `mix do clean, compile` but it's still the same.

scottming:
@bangalcat Did you find any error in the `.lexical/lexical.log` or `.lexical/project.log`?

> as for me, the errors didn't disappear after saving.
since I can't replicate this: 

![CleanShot 2023-11-19 at 19 23 25@2x](https://github.com/lexical-lsp/lexical/assets/12830256/b993e067-4008-497b-b8ed-b0437bd198a4)

bangalcat:
[...]

PR #662, opened 2024-03-26
Use `languageId` field provided by LSP clients.

Moosieus:
Presently, Lexical reads file extensions to evaluate files as eex, heex, or elixir. This approach doesn't work for documents that haven't been saved to the file system. Consequently unsaved eex and heex files don't receive proper diagnostics.

This commit adds `language_id` to the `Document` struct and sets it when provided by LSP. If not available, the file extension will be used as a fallback.

### DIFF
--- a/apps/server/lib/lexical/server/state.ex
+++ b/apps/server/lib/lexical/server/state.ex
@@ -176,7 +176,7 @@ defmodule Lexical.Server.State do
       uri: uri,
       version: version,
       language_id: language_id
-    } = did_open.lsp.text_document
+    } = did_open.text_document
 
     case Document.Store.open(uri, text, version, language_id) do
       :ok ->
--- a/apps/server/lib/lexical/server/state.ex
+++ b/apps/server/lib/lexical/server/state.ex
@@ -170,11 +170,20 @@ defmodule Lexical.Server.State do
     end
   end
 
-  def apply(%__MODULE__{} = state, %DidOpen{lsp: event}) do
-    %TextDocument.Item{text: text, uri: uri, version: version} =
-      text_document = event.text_document
+  def apply(%__MODULE__{} = state, %DidOpen{} = did_open) do
+    %DidOpen{
+      lsp: %DidOpen.LSP{
+        text_document:
+          %TextDocument.Item{
+            text: text,
+            uri: uri,
+            version: version,
+            language_id: language_id
+          } = text_document
+      }
+    } = did_open
 
-    case Document.Store.open(uri, text, version) do
+    case Document.Store.open(uri, text, language_id, version) do
       :ok ->
         Logger.info("opened #{uri}")
         {:ok, state}
--- a/apps/server/lib/lexical/server/state.ex
+++ b/apps/server/lib/lexical/server/state.ex
@@ -178,7 +178,7 @@ defmodule Lexical.Server.State do
       language_id: language_id
     } = event.text_document
 
-    case Document.Store.open(uri, text, language_id, version) do
+    case Document.Store.open(uri, text, version, language_id) do
       :ok ->
         Logger.info("opened #{uri}")
         {:ok, state}
--- a/apps/server/lib/lexical/server/state.ex
+++ b/apps/server/lib/lexical/server/state.ex
@@ -170,18 +170,13 @@ defmodule Lexical.Server.State do
     end
   end
 
-  def apply(%__MODULE__{} = state, %DidOpen{} = did_open) do
-    %DidOpen{
-      lsp: %DidOpen.LSP{
-        text_document:
-          %TextDocument.Item{
-            text: text,
-            uri: uri,
-            version: version,
-            language_id: language_id
-          } = text_document
-      }
-    } = did_open
+  def apply(%__MODULE__{} = state, %DidOpen{lsp: event}) do
+    %TextDocument.Item{
+      text: text,
+      uri: uri,
+      version: version,
+      language_id: language_id
+    } = event.text_document
 
     case Document.Store.open(uri, text, language_id, version) do
       :ok ->
@@ -189,7 +184,7 @@ defmodule Lexical.Server.State do
         {:ok, state}
 
       error ->
-        Logger.error("Could not open #{text_document.uri} #{inspect(error)}")
+        Logger.error("Could not open #{uri} #{inspect(error)}")
         error
     end
   end
--- a/apps/server/lib/lexical/server/state.ex
+++ b/apps/server/lib/lexical/server/state.ex
@@ -170,13 +170,13 @@ defmodule Lexical.Server.State do
     end
   end
 
-  def apply(%__MODULE__{} = state, %DidOpen{lsp: event}) do
+  def apply(%__MODULE__{} = state, %DidOpen{} = did_open) do
     %TextDocument.Item{
       text: text,
       uri: uri,
       version: version,
       language_id: language_id
-    } = event.text_document
+    } = did_open.lsp.text_document
 
     case Document.Store.open(uri, text, version, language_id) do
       :ok ->
--- a/projects/lexical_shared/lib/lexical/document.ex
+++ b/projects/lexical_shared/lib/lexical/document.ex
@@ -44,28 +44,22 @@ defmodule Lexical.Document do
   as a binary and the vewrsion.
   """
   @spec new(Lexical.path() | Lexical.uri(), String.t(), version()) :: t
-  def new(maybe_uri, text, version) do
+  def new(maybe_uri, text, version, language_id \\ nil) do
     uri = DocumentPath.ensure_uri(maybe_uri)
     path = DocumentPath.from_uri(uri)
-    lang_id = language_id_from_path(path)
+
+    language_id =
+      if language_id === nil do
+        language_id_from_path(path)
+      else
+        language_id
+      end
 
     %__MODULE__{
       uri: uri,
       version: version,
       lines: Lines.new(text),
       path: path,
-      language_id: lang_id
-    }
-  end
-
-  def new(maybe_uri, text, language_id, version) do
-    uri = DocumentPath.ensure_uri(maybe_uri)
-
-    %__MODULE__{
-      uri: uri,
-      version: version,
-      lines: Lines.new(text),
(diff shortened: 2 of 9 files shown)


Moosieus:
This PR also fixes two issues.
Presently, as-you-type completions don't work with heex files, see:
  - #495
  - #547
  - #639
  
  This PR disables document-level completions for heex files, meaning diagnostics won't show until the file's saved. This is a preferable stop-gap until full support can be implemented, as mentioned [here](https://github.com/lexical-lsp/lexical/issues/495#issuecomment-1828853060).

Diagnostics in eex were also resulting in LSP errors, fixed by a slight tweak to the error messaging regex.

projects/lexical_shared/lib/lexical/document.ex
```diff
@@ -43,12 +44,27 @@ defmodule Lexical.Document do
...
+    path = DocumentPath.from_uri(uri)
+    lang_id = DocumentPath.language_id_from_path(path)
 
     %__MODULE__{
       uri: uri,
       version: version,
       lines: Lines.new(text),
-      path: DocumentPath.from_uri(uri)
+      path: path,
+      language_id: lang_id
```
scohen, line 55: suggestion:
```elixir
path = DocumentPath.from_uri(uri)
language_id = uri |> DocumentPath.from_uri() |> language_id_from_path()
new(maybe_uri, text, language_id, version)
```

projects/lexical_shared/lib/lexical/document/path.ex
```diff
@@ -125,6 +127,35 @@ defmodule Lexical.Document.Path do
     end
   end
 
+  @spec language_id_from_path(Lexical.path()) :: String.t()
+  def language_id_from_path(path) do
```
scohen, line 131: Language id seems to be a property of `Lexical.Document` and not really related to paths. This module is really about path manipulation and conversion to and from URIs. 
I also think you can move this to `Lexical.Document` and use the URI module to get the path.

projects/lexical_shared/lib/lexical/document/path.ex
```diff
@@ -125,6 +127,35 @@ defmodule Lexical.Document.Path do
...
+      ".eex" ->
+        "eex"
+
+      ".heex" ->
+        "phoenix-heex"
+
+      extension ->
+        Logger.warning("can't infer lang ID for #{path}, ext: #{extension}.")
+
+        "unsupported (#{extension})"
```
scohen, line 148: 🤔 should we include mappings for all the language ids defined here?
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentItem

If so, maybe this belongs in its own module.

apps/server/lib/lexical/server/state.ex
```diff
@@ -170,11 +170,20 @@ defmodule Lexical.Server.State do
...
+      lsp: %DidOpen.LSP{
+        text_document:
+          %TextDocument.Item{
+            text: text,
+            uri: uri,
+            version: version,
+            language_id: language_id
+          } = text_document
+      }
+    } = did_open
```
scohen, line 184: This is a lot more wordy than the previous version.
```suggestion
  def apply(%__MODULE__{} = state, %DidOpen{lsp: event} = did_open) do
    %TextDocument.Item{
            text: text,
            uri: uri,
            version: version,
            language_id: language_id
          } = event.text_document
```

scohen: Also, this might be my issue, but we might not need to reach into the LSP fields. 
We can probably just do `did_open.text_document`

projects/lexical_shared/lib/lexical/document/path.ex
Moosieus: Agreed, moved this function from `Document.Path` to `Document`.

I'd prefer this function assume its input is a path, and have callers invoke `Document.Path.ensure_path/1` should they need to disambiguate it.

projects/lexical_shared/lib/lexical/document/path.ex
Moosieus: I think it's best to only map what's supported by Lexical, and handle any unexpected input generically.

apps/server/lib/lexical/server/state.ex
Moosieus: I prefer de-structuring over direct de-referencing, as it keeps the body of the function more concise. I do like the de-structuring you suggested more though, applied it.

Moosieus: A vs B:
```elixir
def apply(%__MODULE__{} = state, %DidOpen{lsp: event}) do
  %TextDocument.Item{
    text: text,
    uri: uri,
    version: version,
    language_id: language_id
  } = event.text_document
  
  case Document.Store.open(uri, text, language_id, version) do
    :ok ->
      Logger.info("opened #{uri}")
      {:ok, state}
  
  error ->
    Logger.error("Could not open #{uri} #{inspect(error)}")
    error
  end
end
```

```elixir
def apply(%__MODULE__{} = state, %DidOpen{lsp: event}) do
  text_document = event.text_document

  case Document.Store.open(
         text_document.uri,
         text_document.text,
         text_document.language_id,
         text_document.version
       ) do
    :ok ->
      Logger.info("opened #{text_document.uri}")
      {:ok, state}

    error ->
      Logger.error("Could not open #{text_document.uri} #{inspect(error)}")
      error
  end
end
```

I feel it's easier to grok at a glance with destructuring.

projects/lexical_shared/lib/lexical/document.ex
Moosieus: I've updated the function for creating a new document, positioning `language_id` as an optional argument. lmk what you think.

apps/server/lib/lexical/server/state.ex
scohen: My suggested change is not representative of suggestion B, so i don't understand the comment.

projects/lexical_shared/lib/lexical/document/store.ex
```diff
@@ -258,7 +259,15 @@ defmodule Lexical.Document.Store do
 
   @spec open(Lexical.uri(), String.t(), pos_integer()) :: :ok | {:error, :already_open}
   def open(uri, text, version) do
-    GenServer.call(name(), {:open, uri, text, version})
+    language_id = Document.Path.language_id_from_uri(uri)
```
scohen: can we push this down into the document store by defaulting to nil here?

projects/lexical_shared/lib/lexical/document/store.ex
```diff
@@ -258,7 +261,15 @@ defmodule Lexical.Document.Store do
 
   @spec open(Lexical.uri(), String.t(), pos_integer()) :: :ok | {:error, :already_open}
   def open(uri, text, version) do
-    GenServer.call(name(), {:open, uri, text, version})
+    language_id = Document.language_id_from_uri(uri)
+
+    GenServer.call(name(), {:open, uri, text, version, language_id})
```
scohen, line 266: I think having a default language id of nil will simplify things. Then we can have a three arity and four arity open function, and push determining language id down to the document, which would then not have to expose its `language_id_from_*` functions.

projects/lexical_shared/lib/lexical/document.ex
```diff
@@ -219,6 +231,35 @@ defmodule Lexical.Document do
     |> IO.iodata_to_binary()
   end
 
+  @spec language_id_from_uri(Lexical.uri()) :: String.t()
+  def language_id_from_uri(uri) do
+    %URI{path: path} = URI.parse(uri)
```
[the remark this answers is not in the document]
Moosieus: Moving `language_id_from_*` internal to `Document` means we can nix this function entirely. I was having some issues with it anyway, so I think it's better just to derive it from paths exclusively.

projects/lexical_shared/lib/lexical/document/store.ex
Moosieus: Yep, done.

projects/lexical_shared/lib/lexical/document/store.ex
Moosieus: Done as well.

projects/lexical_shared/lib/lexical/document.ex
Moosieus: [Moving forward w/ this here](https://github.com/lexical-lsp/lexical/pull/662#discussion_r1541386976).

projects/lexical_shared/lib/lexical/document/store.ex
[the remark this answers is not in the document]
Moosieus: Much cleaner, done.

projects/lexical_shared/lib/lexical/document/path.ex
```diff
@@ -2,6 +2,7 @@ defmodule Lexical.Document.Path do
   @moduledoc """
   A collection of functions dealing with converting filesystem paths to URIs and back
   """
+
```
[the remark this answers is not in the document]
Moosieus: Reverted this with another commit, didn't wanna mess up the history. Let me know if that works.

apps/server/lib/lexical/server/state.ex
Moosieus: Whoops, my bad. I think I see what you mean now, updated.

apps/server/lib/lexical/server/state.ex
```diff
@@ -170,17 +170,21 @@ defmodule Lexical.Server.State do
...
-      text_document = event.text_document
-
-    case Document.Store.open(uri, text, version) do
+  def apply(%__MODULE__{} = state, %DidOpen{} = did_open) do
+    %TextDocument.Item{
+      text: text,
+      uri: uri,
+      version: version,
+      language_id: language_id
+    } = did_open.lsp.text_document
```
scohen, line 179: do we need the lsp version? I think this will work with `did_open.text_document`

scohen (approved):
Address the one remaining issue, then we're good to go.

apps/server/lib/lexical/server/state.ex
Moosieus: I see what you mean, done.

PR #665, opened 2024-03-27
Fix: Current module not identified in defimpl

scohen:
Our defimpl resolver was not generating the current module name. It was returning the same module as the protocol, which is incorrect


apps/remote_control/test/lexical/remote_control/analyzer_test.exs
```diff
@@ -29,13 +29,13 @@ defmodule Lexical.RemoteControl.AnalyzerTest do
...
 
-        def pack(atom) do
-          |
+          def pack(atom) do
+            |
+          end
         end
-      end
-      ]
+        ]
```
scohen, line 38: This is just a formatting change.

Moosieus (approved):

PR #666, opened 2024-03-28
Added extractors for defimpl and defprotocol

### LINKED ISSUE #664: Add extractors for defimpl and defprotocol
We need extractors for protocols and protocol implementations.

scohen:
Added extractors for defimpl and defprotocol blocks. Also added unit tests

Fixes #664

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
@@ -16,27 +16,34 @@ defmodule Lexical.RemoteControl.Search.Indexer.Extractors.Module do
 
   require Logger
 
+  @definition_mappings %{
+    defmodule: :module,
+    defprotocol: :protocol
+  }
+  @module_definitions Map.keys(@definition_mappings)
+
   # extract a module definition
   def extract(
-        {:defmodule, defmodule_meta,
+        {definition, defmodule_meta,
          [{:__aliases__, module_name_meta, module_name}, module_block]} = defmodule_ast,
         %Reducer{} = reducer
-      ) do
+      )
+      when definition in @module_definitions do
     %Block{} = block = Reducer.current_block(reducer)
 
     case resolve_alias(reducer, module_name) do
       {:ok, aliased_module} ->
         module_position = Metadata.position(module_name_meta)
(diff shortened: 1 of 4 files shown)


apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
```diff
@@ -16,27 +16,34 @@ defmodule Lexical.RemoteControl.Search.Indexer.Extractors.Module do
...
-        range = to_range(reducer, module_name, module_position)
+        detail_range = to_range(reducer, module_name, module_position)
 
         entry =
           Entry.block_definition(
             reducer.analysis.document.path,
             block,
             Subject.module(aliased_module),
-            :module,
+            Map.get(@definition_mappings, definition, :module),
```
Moosieus, line 44: nit: Since the guard clause holds that `definition` must be a key of `@definition_mappings`, I think `Map.fetch/2` is slightly more appropriate.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/protocol.ex
Moosieus, line 1: Was this included in error?

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/module.ex
scohen: heh, i was wondering if `@definition_mappings[definition]` was more appropriate.
I'm not a huge fan of bang functions tbh. I kind of prefer a belt and suspenders approach, people can remove guards in the future, or add keys or all sorts of things.

Moosieus: I don't think there'd ever be an appropriate default where the definition mapping's absent, hence my preference for `Map.fetch!/2` over `Map.get/3` (default of `:module`) or a de-reference (default of `nil`). In that case, better to fail explicitly than let bad values propagate.

My nit here's really small and speculative though - any way gets the job done, so I'll leave it up to you.

scohen: Fixed it.

Moosieus (approved):

PR #672, opened 2024-03-29
Fixing crash when dealing with unicode files

scohen:
The change to binread introduced a fairly complicated unicode bug that would cause decoding of messages to partially fail. This was due to the fact that we were previously using read, which would occasionally return unencoded data. We patched this by doing a utf8 -> latin1 conversion, but this would fail once we acutally read the bytes rather than going through elixir's utf8 encoding friendly IO.read functions.

The change here is to get rid of the encoding step, as we're reading the raw bytes as we should.

### DIFF
--- a/apps/server/lib/lexical/server/transport/std_io.ex
+++ b/apps/server/lib/lexical/server/transport/std_io.ex
@@ -13,7 +13,7 @@ defmodule Lexical.Server.Transport.StdIO do
   end
 
   def init({callback, device}) do
-    :io.setopts([:binary, encoding: :latin1])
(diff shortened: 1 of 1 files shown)


scohen:
@scottming can you take a look at this? You deal with unicode a lot more than I do, I want to make sure things are still good.

FYI, on main, opening `ast_test.exs` would cause my server to crash.

scottming (approved):

scottming:
I did some tests with simple Chinese, and it seems to work very well, similar to the effects of the previous release.

PR #673, opened 2024-03-31
Update nix mix deps sha256

zoedsoupe:
On commit 2178b4b there was an update to dialyzer mix dependency. That causes a new state for `mix.lock` and therefore a new sha256 hash for the mix FOD deps build with `nix`.


scohen:
@hauleth the detection scripts don't seem to be working.

skbolton:
I am starting to think that it would be better to not try and have the nix definition stay up to date with main. It could make more sense to go to a release model and pin a tag and then as part of the release process update the nix definition.

scohen:
I fixed the permissions error. The other error doesn't look like anything to me.

scohen:
One reason is that main isn't in a great state right now

skbolton:
Sounds good @scohen.just wanted to make sure you weren't thinking things were resolved by the action fix.

hauleth:
Ahh, yeah. I need to update the script a little, as it is expected to fail, so it should take that into the account.

hauleth:
#683 supersedes this PR

scohen:
Closing in favor of #683

PR #674, opened 2024-04-01
Workspace symbols

scohen:
Workspace symbols support. Workspace symbols finds all things that match a query in a project.

### DIFF
--- a/apps/remote_control/test/lexical/remote_control/search/indexer/extractors/function_definition_test.exs
+++ b/apps/remote_control/test/lexical/remote_control/search/indexer/extractors/function_definition_test.exs
@@ -160,7 +160,7 @@ defmodule Lexical.RemoteControl.Search.Indexer.Extractors.FunctionDefinitionTest
 
       assert function.type == :public_function
       assert function.subject == "MyProtocol.Structs.Mystruct.do_proto/2"
-      assert decorate(doc, function.range) =~ "def «do_proto(a, b)» do"
+      assert "do_proto(a, b)" = extract(doc, function.range)
       assert decorate(doc, function.block_range) =~ "«def do_proto(a, b) do\n    a + b\n  end»"
     end
 
--- a/apps/remote_control/test/lexical/remote_control/search/indexer/extractors/function_definition_test.exs
+++ b/apps/remote_control/test/lexical/remote_control/search/indexer/extractors/function_definition_test.exs
@@ -147,6 +147,23 @@ defmodule Lexical.RemoteControl.Search.Indexer.Extractors.FunctionDefinitionTest
       assert expected == extract(code, multi_line.range)
     end
 
+    test "finds functions defined in protocol implementations" do
+      {:ok, [function], doc} =
+        ~q[
+          defimpl MyProtocol, for: Structs.Mystruct do
+            def do_proto(a, b) do
+              a + b
+            end
+          end
+        ]
+        |> index_functions()
+
+      assert function.type == :public_function
+      assert function.subject == "MyProtocol.Structs.Mystruct.do_proto/2"
+      assert decorate(doc, function.range) =~ "def «do_proto(a, b)» do"
+      assert decorate(doc, function.block_range) =~ "«def do_proto(a, b) do\n    a + b\n  end»"
+    end
+
     test "skips public functions defined in quote blocks" do
       code =
         ~q[
--- a/apps/server/lib/lexical/server/provider/queue.ex
+++ b/apps/server/lib/lexical/server/provider/queue.ex
@@ -136,7 +136,28 @@ defmodule Lexical.Server.Provider.Queue do
       Queue.Supervisor.run_in_task(handler)
     end
 
+    alias Lexical.Protocol.Types.ErrorCodes
+    require ErrorCodes
+
     defp write_reply(response) do
+      case Convert.to_lsp(response) do
+        {:ok, lsp_response} ->
+          Transport.write(lsp_response)
+
+        error ->
+          Logger.critical(
+            "Failed to convert into a response #{inspect(error)}  #{inspect(response)}"
+          )
+
+          response_error =
+            ResponseError.new(
+              code: :internal_error,
+              message: "Failed to convert #{response.__struct__} due to #{inspect(error)}"
+            )
+
+          Transport.write(%{id: response.id, error: response_error})
+      end
+
       with {:ok, lsp_response} <- Convert.to_lsp(response) do
         Transport.write(lsp_response)
       end
--- a/apps/server/lib/lexical/server/provider/queue.ex
+++ b/apps/server/lib/lexical/server/provider/queue.ex
@@ -157,10 +157,6 @@ defmodule Lexical.Server.Provider.Queue do
 
           Transport.write(%{id: response.id, error: response_error})
       end
-
-      with {:ok, lsp_response} <- Convert.to_lsp(response) do
-        Transport.write(lsp_response)
-      end
     end
   end
 
--- a/apps/protocol/lib/generated/lexical/protocol/types/location/link.ex
+++ b/apps/protocol/lib/generated/lexical/protocol/types/location/link.ex
@@ -1,0 +1,11 @@
+# This file's contents are auto-generated. Do not edit.
+defmodule Lexical.Protocol.Types.Location.Link do
+  alias Lexical.Proto
+  alias Lexical.Protocol.Types
+  use Proto
+
+  deftype origin_selection_range: optional(Types.Range),
+          target_range: Types.Range,
+          target_selection_range: Types.Range,
+          target_uri: string()
+end
--- a/apps/protocol/lib/generated/lexical/protocol/types/workspace/symbol.ex
+++ b/apps/protocol/lib/generated/lexical/protocol/types/workspace/symbol.ex
@@ -1,0 +1,19 @@
+# This file's contents are auto-generated. Do not edit.
+defmodule Lexical.Protocol.Types.Workspace.Symbol do
+  alias Lexical.Proto
+  alias Lexical.Protocol.Types
+
+  defmodule Location do
+    use Proto
+    deftype uri: string()
+  end
+
+  use Proto
+
+  deftype container_name: optional(string()),
+          data: optional(any()),
+          kind: Types.Symbol.Kind,
+          location: one_of([Types.Location, Lexical.Protocol.Types.Workspace.Symbol.Location]),
+          name: string(),
+          tags: optional(list_of(Types.Symbol.Tag))
+end
--- a/apps/protocol/lib/generated/lexical/protocol/types/workspace/symbol/params.ex
+++ b/apps/protocol/lib/generated/lexical/protocol/types/workspace/symbol/params.ex
@@ -1,0 +1,10 @@
+# This file's contents are auto-generated. Do not edit.
+defmodule Lexical.Protocol.Types.Workspace.Symbol.Params do
+  alias Lexical.Proto
+  alias Lexical.Protocol.Types
+  use Proto
+
+  deftype partial_result_token: optional(Types.Progress.Token),
+          query: string(),
+          work_done_token: optional(Types.Progress.Token)
+end
--- a/apps/protocol/lib/lexical/protocol/requests.ex
+++ b/apps/protocol/lib/lexical/protocol/requests.ex
@@ -82,6 +82,12 @@ defmodule Lexical.Protocol.Requests do
     defrequest "textDocument/documentSymbol", Types.Document.Symbol.Params
   end
 
+  defmodule WorkspaceSymbol do
+    use Proto
+
+    defrequest "workspace/symbol", Types.Workspace.Symbol.Params
+  end
+
   # Server -> Client requests
 
   defmodule RegisterCapability do
--- a/apps/protocol/lib/lexical/protocol/responses.ex
+++ b/apps/protocol/lib/lexical/protocol/responses.ex
@@ -56,6 +56,12 @@ defmodule Lexical.Protocol.Responses do
     defresponse optional(list_of(Types.Document.Symbol))
   end
 
+  defmodule WorkspaceSymbol do
+    use Proto
+
+    defresponse optional(list_of(Types.Workspace.Symbol))
+  end
+
   defmodule Shutdown do
     use Proto
     # yeah, this is odd... it has no params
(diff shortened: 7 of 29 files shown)


dimitarvp:
OK, some more screwing around reveals that this is kinda related to opening Neovim in a vertical split mode, I guess?

So I start my editor this way: `lvim -O2 lib/file1.ex lib/file2.ex`

I start off in `file1.ex`, I click on `file2.ex`, then I click back on `file1.ex`'s viewport and the error shows up.

Same happens with `-o2`.

The error does NOT manifest itself if the editor is opened with a single pane.

Finally, trying to invoke listing of file symbols shows this big error:

```plain
LSP[lexical]: Error NO_RESULT_CALLBACK_FOUND: {
  id = 7,
  jsonrpc = "2.0",
  result = { {
      children = { {
          children = {},
          kind = 12,
          name = "receive_srr(%Conn{} = conn, %{oem: _} = params)",
          range = {
            ["end"] = {
              character = 5,
              line = 39
            },
            start = {
              character = 2,
              line = 31
            }
          },
          selectionRange = {
            ["end"] = {
              character = 53,
              line = 31
            },
            start = {
              character = 6,
              line = 31
            }
          }
        }, {
          children = {},
          kind = 12,
          name = "receive_mo(%Conn{} = conn, %{oem: _} = params)",
          range = {
            ["end"] = {
              character = 5,
              line = 62
            },
            start = {
              character = 2,
              line = 54
            }
          },
          selectionRange = {
            ["end"] = {
              character = 52,
              line = 54
            },
            start = {
              character = 6,
              line = 54
            }
          }
        } },
      kind = 2,
      name = "OssM2m.V1.ShortMessages.SmsController",
      range = {
        ["end"] = {
          character = 3,
          line = 63
        },
        start = {
          character = 0,
          line = 0
        }
      },
      selectionRange = {
        ["end"] = {
          character = 47,
          line = 0
        },
        start = {
          character = 10,
          line = 0
        }
      }
    } }
}
```

Hope any of that helps.

dimitarvp:
Oh btw, this also happens not just when clicking the buffers with the mouse. Doing <kbd>C-h</kbd> and <kbd>C-l</kbd> causes the error as well.

scohen:
This looks to be a problem with the plugin, as evidenced by the fact that it seems to only manifest itself when something editor-specific is happening. The LSP has no idea what's going on with the editor. The response you sent looks like a valid workspace symbols response.  I'm not a vim user, so maybe @scottming has ideas.

The ["end"] in selection range is suspicious though.

dimitarvp:
Sorry, which plugin?

I can make a super mini Elixir project and see if we can have that reproduced on your machine (if you use NeoVim at all, that is).

scohen:
whatever you're using to enable language servers with vim. Maybe "LSP client" is better terminology.

dimitarvp:
Well I suppose it's NeoVim's builtin LSP software.

Any ideas how can I troubleshoot this further? I am busy enough and not being versed in the implementation details of my editor is a conscious choice that I am not looking to change, I am afraid.

Is there any other way I can help?

scohen:
One of our contributors, @scottming has been helpful with this stuff in the past.
What's interesting is that the response you've highlighted is from a document symbols response, not a workspace symbols response. This is probably why clicking around is crashing things, as those requests are fired whenever the cursor moves.

scottming:
@scohen @dimitarvp I can replicate this error, will take a look today or tomorrow.

dimitarvp:
@scohen My apologies for not posting the output of listing workspace symbols. My bad. I'll do that and post the output Later™.

apps/server/lib/lexical/server/provider/queue.ex
```diff
@@ -136,7 +136,28 @@ defmodule Lexical.Server.Provider.Queue do
...
+
+          response_error =
+            ResponseError.new(
+              code: :internal_error,
+              message: "Failed to convert #{response.__struct__} due to #{inspect(error)}"
+            )
+
+          Transport.write(%{id: response.id, error: response_error})
+      end
+
```
scottming, line 160: @dimitarvp @scohen 

After spending some time, I found out that these lines of code were causing `nvim` to crash repeatedly. This code responds more than twice to the same `id`, which we need to avoid.

CleanShot 2024-04-03 at 21 19 12@2x

When I tried checking out this file from the main branch, everything worked fine

```bash
git checkout main apps/server/lib/lexical/server/provider/queue.ex
```

scohen: @dimitarvp can you look in the lexical logs for the error that's causing the double write?

dimitarvp: @scohen There are several hundred log lines that seem relevant (at least they both contain `elixir-ls` and `lexical`), do you want them all?

dimitarvp: I have zero such lines in my log though. Hmmm. I'll try to do another experiment today or tomorrow and see if I can make it manifest itself (and will also post any findings about trying to list workspace symbols).

scohen: @dimitarvp i've pushed a fix to this branch that removes the double-write, so you should be able to use lexical now. 

just let it run and whenever you have a chance, grep your logs for "Failed to convert" and open an issue.

scohen: Thanks, @scottming for debugging this.

apps/remote_control/test/lexical/remote_control/search/indexer/extractors/function_definition_test.exs
```diff
@@ -147,6 +147,23 @@ defmodule Lexical.RemoteControl.Search.Indexer.Extractors.FunctionDefinitionTest
...
+            def do_proto(a, b) do
+              a + b
+            end
+          end
+        ]
+        |> index_functions()
+
+      assert function.type == :public_function
+      assert function.subject == "MyProtocol.Structs.Mystruct.do_proto/2"
+      assert decorate(doc, function.range) =~ "def «do_proto(a, b)» do"
```
scottming, line 163: This seems to be inconsistent with the other `function`'s definition  range.

scottming (commented):
CleanShot 2024-04-12 at 15 09 42@2xI tried out this branch and found that it works very well when searching for function definition symbols, but when I want to search for modules, it still first returns functions and the functions from dependencies. I think when the first letter is *capitalized*, it should prioritize returning modules, and it should prioritize returning the last module with a high degree of match.

Additionally, when I enter the attribute @user, it doesn't return anything. Is this intentional?

apps/remote_control/test/lexical/remote_control/search/indexer/extractors/function_definition_test.exs
scohen: how so? I think this test is just using decorate rather than extract.

scohen:
> but when I want to search for modules, it still first returns functions

That's not how workspace symbols works or is supposed to work. You're searching for all symbols, not a specific type, and if a symbol has some text in it, you're going to get that symbol back. We can control ordering to an extent though different editors sort things differently.

> it still first returns functions and the functions from dependencies.

I've noticed this too, i don't know what reverted to cause this.

> Additionally, when I enter the attribute `@user`, it doesn't return anything. Is this intentional?

I get module attributes when i search for their names (`user` in this case), but not when using the `@` prefix. I'll investigate.

scohen:
@scottming I fixed the dependencies issue, I added nil to the app for testing, which pulled in anything that didn't have an app, which included a _lot_ of dependencies. I removed nil from the application list, so you shouldn't see any deps in your completions any more.

As for the module attributes, I pulled out the leading `@` sign when creating the subject to normalize them, so that's why they wouldn't appear as results if the query included a leading `@`. I'm not sure why i did that, it doesn't affect quality, so i removed that.

scottming:
> That's not how workspace symbols works or is supposed to work. You're searching for all symbols, not a specific type, and if a symbol has some text in it, you're going to get that symbol back. We can control ordering to an extent though different editors sort things differently.


So if I want to quickly search for a module using this feature, how should I do it, especially when I don't quite remember the context of the module(Like only remember the `User`, not `LiveDemoAccounts`)?

scohen:
> So if I want to quickly search for a module using this feature, how should I do it, especially when I don't quite remember the context of the module(Like only remember the User, not LiveDemoAccounts

Searching for "User" should find it just fine.

scottming:
> Searching for "User" should find it just fine.

But as I commented before, there are some issues with the sorting when searching like this. I actually don't quite understand why the `user_fixture` function in the fixtures file would be ranked ahead of the `User` module.

CleanShot 2024-04-15 at 21 40 02@2x

scottming:
BTW, aside from filtering out dependencies, it seems unnecessary for constants like `@impl ...` to appear in the results

CleanShot 2024-04-15 at 21 46 08@2x

scohen:
> why the user_fixture function in the fixtures file would be ranked ahead of the User module.

Because your editor is ranking them, we merely provide them.

scohen:
> BTW, aside from filtering out dependencies, it seems unnecessary for constants like @impl ... to appear in the results

That's your opinion. It's quite possible that someone wants to see where they use `@impl` or search for `@impl GenServer`.

scottming:
> I want to be very clear here, we are providing a list of workspace symbols that match a query. Your editor then takes this list and orders it the way it sees fit. You're disagreeing with that ranking; we don't have much control over it.

Actually, I double-checked and found that `VSCode` does indeed reorder, while `nvim` does not; it simply displays the data returned by lexical. When I search for `Product`, this is the specific response received by the client. 
<details>
<summary>Response</summary>

```lua
[DEBUG][2024-04-17 10:15:48] .../vim/lsp/rpc.lua:387  "rpc.receive" {
  id = 32,
  jsonrpc = "2.0",
  result = { {
      kind = 12,
      location = {
        range = {
          ["end"] = {
            character = 35,
            line = 9
          },
          start = {
            character = 6,
            line = 9
          }
        },
        uri = "file:///Users/scottming/Code/live_demo/test/support/fixtures/catalog_fixtures.ex"
      },
      name = "LiveDemo.CatalogFixtures.product_fixture/1"
    }, {
      kind = 12,
      location = {
        range = {
          ["end"] = {
            character = 22,
            line = 37
          },
          start = {
            character = 6,
            line = 37
          }
        },
        uri = "file:///Users/scottming/Code/live_demo/lib/live_demo/catalog.ex"
      },
      name = "LiveDemo.Catalog.get_product!/1"
    }, {
      kind = 12,
      location = {
        range = {
          ["end"] = {
            character = 49,
            line = 72
          },
          start = {
            character = 7,
            line = 72
          }
        },
        uri = "file:///Users/scottming/Code/live_demo/lib/live_demo_web/live/product_live/form_component.ex"
      },
      name = "LiveDemoWeb.ProductLive.FormComponent.save_product/3"
    }, {
      kind = 12,
      location = {
        range = {
          ["end"] = {
            character = 50,
            line = 57
          },
          start = {
            character = 7,
            line = 57
          }
        },
        uri = "file:///Users/scottming/Code/live_demo/lib/live_demo_web/live/product_live/form_component.ex"
      },
      name = "LiveDemoWeb.ProductLive.FormComponent.save_product/3"
    }, {
      kind = 12,
      location = {
        range = {
          ["end"] = {
            character = 19,
            line = 19
          },
          start = {
            character = 6,
            line = 19
          }
        },
        uri = "file:///Users/scottming/Code/live_demo/lib/live_demo/catalog.ex"
      },
      name = "LiveDemo.Catalog.list_products/0"
    }, {
      kind = 12,
      location = {
        range = {
          ["end"] = {
            character = 49,
            line = 69
          },
          start = {
            character = 6,
            line = 69
          }
        },
        uri = "file:///Users/scottming/Code/live_demo/lib/live_demo/catalog.ex"
      },
      name = "LiveDemo.Catalog.update_product/2"
    }, {
      kind = 12,
      location = {
        range = {
          ["end"] = {
            character = 34,
            line = 51
          },
          start = {
            character = 6,
            line = 51
          }
        },
        uri = "file:///Users/scottming/Code/live_demo/lib/live_demo/catalog.ex"
      },
      name = "LiveDemo.Catalog.create_product/1"
    }, {
      kind = 12,
      location = {
        range = {
          ["end"] = {
            character = 42,
            line = 87
          },
          start = {
            character = 6,
            line = 87
          }
        },
        uri = "file:///Users/scottming/Code/live_demo/lib/live_demo/catalog.ex"
      },
      name = "LiveDemo.Catalog.delete_product/1"
    }, {
      kind = 12,
      location = {
        range = {
          ["end"] = {
            character = 56,
            line = 100
          },
          start = {
            character = 6,
            line = 100
          }
        },
        uri = "file:///Users/scottming/Code/live_demo/lib/live_demo/catalog.ex"
      },
      name = "LiveDemo.Catalog.change_product/2"
    }, {
      kind = 2,
      location = {
        range = {
          ["end"] = {
            character = 38,
            line = 0
          },
          start = {
            character = 10,
            line = 0
          }
        },
        uri = "file:///Users/scottming/Code/live_demo/lib/live_demo_web/live/product_live/show.ex"
      },
      name = "LiveDemoWeb.ProductLive.Show"
    }, {
      kind = 2,
      location = {
        range = {
          ["end"] = {
            character = 39,
            line = 0
          },
          start = {
            character = 10,
            line = 0
          }
        },
        uri = "file:///Users/scottming/Code/live_demo/lib/live_demo_web/live/product_live/index.ex"
      },
      name = "LiveDemoWeb.ProductLive.Index"
    }, {
      kind = 2,
      location = {
        range = {
          ["end"] = {
            character = 47,
            line = 0
          },
          start = {
            character = 10,
            line = 0
          }
        },
        uri = "file:///Users/scottming/Code/live_demo/lib/live_demo_web/live/product_live/form_component.ex"
      },
      name = "LiveDemoWeb.ProductLive.FormComponent"
    }, {
      kind = 2,
      location = {
        range = {
          ["end"] = {
            character = 34,
            line = 0
          },
          start = {
            character = 10,
            line = 0
          }
        },
        uri = "file:///Users/scottming/Code/live_demo/lib/live_demo/catalog/product.ex"
      },
      name = "LiveDemo.Catalog.Product"
    } }
}
```
</details>
Below is the display effect. 

<details>
<summary>effect</summary>

image

</details>
You can use these command to easily create a Phoenix project to replicate these issues.


```bash
mix phx.new live_demo
mix phx.gen.live Catalog Product products title:string description:string price:decimal views:integer
```

scohen:
I explained why the ordering is the way it is. I feel like we can go back and forth on this, but ordering is more a property of the fuzzy matcher while this pr is about enabling workspace symbols. 
emacs also reorders for what it’s worth.

scottming (approved):
Sorting issues may require more thought, but can be addressed in the future PR or client-side

PR #679, opened 2024-04-02
Added module attribute detection

scohen:
Our environment can now detect when it's in a module attribute context

This will enable us to check for randomly named module attributes and fix some of our longstanding issues.


Moosieus:
Should `@callback` and `@macrocallback` have their own context_types as well?

Moosieus (approved):

PR #684, opened 2024-04-03
Log errors in Server.Transport.StdIO.

Moosieus:
* Added error log in read loop.
* Improved error values in header and body parsing.


apps/server/lib/lexical/server/transport/std_io.ex
```diff
@@ -76,27 +83,33 @@ defmodule Lexical.Server.Transport.StdIO do
...
   defp read_body(device, byte_count) do
     case IO.binread(device, byte_count) do
-      data when is_binary(data) or is_list(data) ->
+      data when is_binary(data) ->
         {:ok, data}
 
-      other ->
-        other
+      :eof ->
+        {:error, :eof}
```
scohen, line 102: I'm pretty sure we want to shut down in this case, as it signals that the stream is closed.

Moosieus: Good point, think this would work?
```elixir
:eof ->
  Logger.critical("stdio received :eof, server will stop.")
  maybe_stop()
```

scohen: Sure, though it's likely you won't see that message.

Moosieus: Added

apps/server/lib/lexical/server/transport/std_io.ex
```diff
@@ -59,44 +61,57 @@ defmodule Lexical.Server.Transport.StdIO do
...
+    case List.keyfind(headers, "content-length", 0) do
+      {_, len_str} ->
+        {:ok, String.to_integer(len_str)}
 
-  defp header_value(headers, header_name, converter) do
-    case List.keyfind(headers, header_name, 0) do
-      nil -> :error
-      {_, value} -> {:ok, converter.(value)}
+      nil ->
+        {:error, :no_content_length_header}
```
scohen, line 93: one more nitpick

```elixir 
with {_, len_str} <- List.keyfind(headers, "content-length", 0),
       {content_length, ""} <- Integer.parse(len_str) do
  {:ok, content_length}
else _ -> 
  {:error, :no_content_length_header}
end
```

scohen: (String.to_integer raises)

Moosieus: That's what was there - could rewrite it like this though:
```elixir
defp content_length(headers) do
  with {:ok, len_str} <- find_header(headers, "content-length"),
       {:ok, int} <- parse_length(len_str) do
    {:ok, int}
  end
end

defp find_header(headers, name) do
  case List.keyfind(headers, name, 0) do
    {_, len_str} -> {:ok, len_str}
    nil -> {:error, {:header_not_found, name}}
  end
end

defp parse_length(len_str) when is_binary(len_str) do
  case Integer.parse(len_str) do
    {int, ""} -> {:ok, int}
    :error -> {:error, {:cant_parse_length, len_str}}
  end
end
```

scohen: this is fine

Moosieus: Added

scohen (approved):

PR #687, opened 2024-04-05
Fixed: Error while calculating block spans

scohen:
While trying to enhance the indexing of protocol modules, I found a fairly serious bug with alias / import detection. When we'd encounter a block of code (usually starting with do), we used soureror's detection of start and end. The problem is that it starts that block before the do, so for example, `defmodule MyModule do..` would start at the `defmodule` call.

This meant that the module we were defining starts at the defmodule call, so our current module detection would be off by almost an entire line. This caused errors when builidng module names for protocols and when people would do `defmodule __MODULE__.Child do`.

We were also making nonsensical module names like `Elixir.__MODULE__` when we called `current_module` despite the fact that not all code is inside a module.

Now, module detection is character accurate and can fail, and some code had to have error detection added. For future reference, all calls to `current_module(analysis, position)` can now fail, so doing a hard pattern match on `{:ok, module}` will fail in contexts outside of a module.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
@@ -40,9 +40,17 @@ defmodule Lexical.RemoteControl.Search.Indexer.Entry do
   defguard is_structure(entry) when entry.type == :metadata and entry.subtype == :block_structure
   defguard is_block(entry) when entry.id == entry.block_id
 
-  def copy(%__MODULE__{} = orig, overrides) do
+  @doc """
+  Creates a new entry by copying the passed-in entry.
+
+  The returned entry will have the same fields set as the one passed in,
+  but a different id.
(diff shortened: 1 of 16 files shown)


apps/remote_control/lib/lexical/remote_control/search/indexer/entry.ex
```diff
@@ -40,6 +40,11 @@ defmodule Lexical.RemoteControl.Search.Indexer.Entry do
   defguard is_structure(entry) when entry.type == :metadata and entry.subtype == :block_structure
   defguard is_block(entry) when entry.id == entry.block_id
 
+  def copy(%__MODULE__{} = orig, overrides) do
+    %__MODULE__{orig | id: Identifier.next_global!()}
+    |> Map.merge(overrides)
+  end
+
```
Moosieus, line 47: Could you add a brief doc string here?

scottming (approved):

PR #692, opened 2024-04-12
Stopped completions inside strings

### LINKED ISSUE #657: Completions appear in strings
image

Relevant to #638, as hovers there should not provide hints or analysis either.

scohen:
We no longer issue completions inside of strings.
Also fixed a bug where aliases would be detected in multi-line heredocs.

Fixes #657


apps/common/lib/lexical/ast/detection/alias.ex
```diff
@@ -23,37 +23,42 @@ defmodule Lexical.Ast.Detection.Alias do
...
-        {:cont, false}
-    end)
+        _, _ ->
+          {:cont, false}
+      end)
+
+    case result do
+      b when is_boolean(b) -> b
+      _ -> false
+    end
```
zachallaun, line 62: Being unfamiliar with this code, it took a while for me to figure out why this was necessary. I think being explicit about the non-boolean value that we might expect would make this more clear.

```suggestion
    case result do
      b when is_boolean(b) -> b
      :eol -> false
    end
```

apps/common/lib/lexical/ast/detection.ex
```diff
@@ -65,6 +66,20 @@ defmodule Lexical.Ast.Detection do
     end)
   end
 
+  def ast_to_range(ast) do
+    ast_to_range(ast, 0, 0)
+  end
+
+  def ast_to_range(ast, start_offset, end_offset) do
+    %{start: [line: start_line, column: start_col], end: [line: end_line, column: end_col]} =
+      Sourceror.get_range(ast)
```
zachallaun, line 75: `Sourceror.get_range/1` can return `nil`, which needs to be handled here. It looks like we already have a `Lexical.Ast.Range.fetch/2`, which I think should be used instead. It could be modified to support the start/end offsets.

This will obviously mean we'll have to modify the usage in the detection modules to handle an `:error` case.

apps/common/lib/lexical/ast/tokens.ex
```diff
@@ -145,4 +145,26 @@ defmodule Lexical.Ast.Tokens do
   defp normalize_type(:flt), do: :float
   defp normalize_type(:bin_string), do: :string
   defp normalize_type(type), do: type
+
+  defp interpolation_ranges(interpolations) do
+    {_, ranges} =
+      Enum.reduce(interpolations, {{1, 1}, []}, fn
+        literal, {{line, column}, acc} when is_binary(literal) ->
+          end_pos = {line, column + String.length(literal)}
+
```
zachallaun, line 154:
```suggestion
```

zachallaun (changes requested):

apps/common/lib/lexical/ast/detection/alias.ex
scohen: yes, but `fetch_range` takes a document, which makes the api a lot more cumbersome.
Can you provide an example of sourceror returning nil in these cases? I know it's possible, but I was under the impression that it will only do so if there is no information in the AST, which is not the case for our usages.

apps/common/lib/lexical/ast/detection.ex
scohen: yes, but fetch_range takes a document, which makes the api cumbersome.
Can you provide an example of sourceror returning nil in these cases? I know it's possible, but I was under the impression that it will only do so if there is no information in the AST, which is not the case for our usages.

zachallaun: > yes, but fetch_range takes a document, which makes the api cumbersome.

Fair!

> Can you provide an example of sourceror returning nil in these cases? 

So the detection abstractions are new to me and it's definitely possible that `nil` doesn't need to be handled in these cases. That said, here's one example and what `get_range` returns:

```elixir
iex(15)> {:ok, ast, _} = Lexical.Ast.from(~S|"foo#{}"|)
{:ok,
 {:<<>>, [delimiter: "\"", line: 1, column: 1],
  [
    "foo",
    {:"::", [line: 1, column: 5],
     [
       {{:., [line: 1, column: 5], [Kernel, :to_string]},
        [closing: [line: 1, column: 7], line: 1, column: 5],
        [{:__block__, [], []}]},
       {:binary, [line: 1, column: 5], nil}
     ]}
  ]}, []}

iex(17)> ast |> Macro.prewalker() |> Enum.map(&{Sourceror.get_range(&1), &1})
[
  {%{end: [line: 1, column: 9], start: [line: 1, column: 1]},
   {:<<>>, [delimiter: "\"", line: 1, column: 1],
    [...snip]}},

  {nil, "foo"},

  {%{end: [line: 1, column: 11], start: [line: 1, column: 1]},
   {:"::", [line: 1, column: 5],
    [...snip]}},

  {%{end: [line: 1, column: 8], start: [line: 1, column: 1]},
   {{:., [line: 1, column: 5], [Kernel, :to_string]},
    [closing: [line: 1, column: 7], line: 1, column: 5],
    [{:__block__, [], []}]}},

  {nil, {:., [line: 1, column: 5], [Kernel, :to_string]}},

  {nil, Kernel},

  {nil, :to_string},

  {nil, {:__block__, [], []}},

  {%{end: [line: 1, column: 11], start: [line: 1, column: 5]},
   {:binary, [line: 1, column: 5], nil}}
]
```

zachallaun (approved):
This looks good to me! I'll let you decide on the `get_range` wrapper and whether nils need to be handled there.

apps/common/lib/lexical/ast/detection.ex
scohen: Thanks, that's helpful, let me think on this a bit.

PR #698, opened 2024-04-16
textDocument/definition: respond with nil result when not found

### LINKED ISSUE #697: Request textDocument/definition failed ... Message: "No definition found for..."
I don't think this is quite the same as #693, but using VSCode on main, `ef747aa`, I get the following when holding CMD and hovering over module definitions. These modules don't contain HEEx.

This is a clean Phoenix project 1.7, nothing added.

```
2024-04-16 06:59:14.386 [info] warning: the log level :warn is deprecated, use :warning instead
  (logger 1.16.1) lib/logger.ex:1165: Logger.elixir_level_to_erlang_level/1
  (logger 1.16.1) lib/logger.ex:627: Logger.compare_levels/2
  (logger_file_backend 0.0.13) lib/logger_file_backend.ex:36: LoggerFileBackend.handle_event/2
  (stdlib 5.2) gen_event.erl:814: :gen_event.server_update/4
  (stdlib 5.2) gen_event.erl:796: :gen_event.server_notify/4
  (stdlib 5.2) gen_event.erl:538: :gen_event.handle_msg/6


2024-04-16 06:59:15.346 [info] [Error - 6:59:15 AM] Request textDocument/definition failed.
2024-04-16 06:59:15.346 [info]   Message: {:unsupported, {:keyword, ~c"true"}}
  Code: -32803 
2024-04-16 06:59:15.650 [info] [Error - 6:59:15 AM] Request textDocument/definition failed.
2024-04-16 06:59:15.650 [info]   Message: "No definition found for \"Supervisor\""
  Code: -32803 
```

zachallaun:
Error responses should only be used for exceptional circumstances (e.g. a literal raised exception was caught) as opposed to "this thing wasn't found". In VS Code, for instance, error responses cause a pop-up to appear that you have to manually dismiss.

I checked all the other handlers and it looks like this was the only one still using the error response for something like this.

Fixes #697.

### DIFF
--- a/apps/server/test/lexical/server/provider/handlers/go_to_definition_test.exs
+++ b/apps/server/test/lexical/server/provider/handlers/go_to_definition_test.exs
@@ -56,12 +56,34 @@ defmodule Lexical.Server.Provider.Handlers.GoToDefinitionTest do
   describe "go to definition" do
     setup [:with_referenced_file]
 
-    test "find the function defintion", %{project: project, uri: referenced_uri} do
+    test "finds user-defined functions", %{project: project, uri: referenced_uri} do
       uses_file_path = file_path(project, Path.join("lib", "uses.ex"))
       {:ok, request} = build_request(uses_file_path, 4, 17)
 
       {:reply, %{result: %Location{} = location}} = handle(request, project)
       assert Location.uri(location) == referenced_uri
+    end
+
+    test "finds user-defined modules", %{project: project, uri: referenced_uri} do
+      uses_file_path = file_path(project, Path.join("lib", "uses.ex"))
+      {:ok, request} = build_request(uses_file_path, 4, 4)
+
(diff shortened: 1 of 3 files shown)


apps/server/test/lexical/server/provider/handlers/go_to_definition_test.exs
zachallaun, line 1: These are "old-style" tests that could be made much more readable and reliable like the [hover tests](https://github.com/lexical-lsp/lexical/blob/main/apps/server/test/lexical/server/provider/handlers/hover_test.exs#L125-L136), for instance, but I don't have time to do it right now and so stuck with what was already there.

IMO we should let this PR go forward and then add an issue to refactor these tests.

scottming (approved):
Welcome back, Zach. I also find errors quite annoying when I occasionally use VSCode

PR #704, opened 2024-04-18
Included non-deps symbols in fuzzy matcher

scohen:
I noticed that the fuzzy matcher wouldn't match any unit tests. This is because their application is nil, as they're scripts. This change checks for nil applications, and if it finds one, only excludes the mapped entry if the path looks like it comes from a dependencies directory.

Note, the deps directory name is hard coded. I originally queried the project for the deps directory name, but this would include dependencies in the projects directory, which was quite annoying, hence the hard-coding.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/fuzzy.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/fuzzy.ex
@@ -359,8 +359,26 @@ defmodule Lexical.RemoteControl.Search.Fuzzy do
         names
       end
 
-    fn mapped(application: app, subtype: subtype) ->
-      subtype == :definition and MapSet.member?(app_names, app)
+    # we do the path thing to ensure that
+    # we get a valid OS-specifc path component
+    # so we can do `String.contains?` on a path
+    deps_part =
+      [" ", "deps", " "]
+      |> Path.join()
+      |> String.trim()
+
+    fn
+      mapped(application: nil, subtype: :definition, grouping_key: path) ->
+        # if we don't have an app name, just make sure we're not
+        # in what looks like a deps directory
+
+        not String.contains?(path, deps_part)
+
+      mapped(application: app, subtype: :definition) ->
+        MapSet.member?(app_names, app)
+
+      _ ->
+        false
     end
   end
 end
--- a/apps/remote_control/lib/lexical/remote_control/search/fuzzy.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/fuzzy.ex
@@ -370,27 +370,16 @@ defmodule Lexical.RemoteControl.Search.Fuzzy do
     # and lexical won't understand this. This was done because loading
     # each sub-project is expensive and changes our global directory.
 
-    root_deps_path = Mix.Project.deps_path()
-    {root_deps_path, Project.mix_exs_path(project)}
-
     [Project.root_path(project), "**", "mix.exs"]
     |> Path.join()
     |> Path.wildcard()
-    |> Enum.map(&Path.absname/1)
-    |> Enum.reduce([], fn mix_exs_path, deps_paths ->
-      deps_dir =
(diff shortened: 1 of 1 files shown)


apps/remote_control/lib/lexical/remote_control/search/fuzzy.ex
```diff
@@ -359,8 +359,26 @@ defmodule Lexical.RemoteControl.Search.Fuzzy do
...
+      [" ", "deps", " "]
+      |> Path.join()
+      |> String.trim()
+
+    fn
+      mapped(application: nil, subtype: :definition, grouping_key: path) ->
+        # if we don't have an app name, just make sure we're not
+        # in what looks like a deps directory
+
+        not String.contains?(path, deps_part)
```
zachallaun, line 375: What happens if our module has `Deps` in it? e.g. we're in `test/my_lib/parent/deps/child_test.exs`

scohen: it won't show up.

scohen: Technically, we'd need to interrogate each project's mix config to figure out what directory its deps reside in, `deps` is just the default, and I don't know how to go from app name -> mix project or from a directory to a mix project.

I can try to come up with a 80% heuristic though, that should be no worse than the above.

scohen:
@zachallaun please take another look.

scohen:
FYI, @zachallaun the previous approach wouldn't allow searches for `Lexical.Document` to succeed either.

apps/remote_control/lib/lexical/remote_control/search/fuzzy.ex
```diff
@@ -336,31 +337,63 @@ defmodule Lexical.RemoteControl.Search.Fuzzy do
...
+        |> Path.join("deps")
+
+      if File.exists?(deps_dir) do
+        [deps_dir | deps_paths]
+      else
+        deps_paths
+      end
+    end)
+    |> Enum.reverse()
+    |> Enum.uniq()
```
zachallaun, line 393: A slight simplification of this pipeline. Also, I think maybe the `uniq` can be excluded? I could be missing something obvious, but it seems like there shouldn't be duplicates.
```suggestion
    |> Enum.map(fn relative_mix_path ->
      relative_mix_path
      |> Path.absname()
      |> Path.dirname()
      |> Path.join("deps")
    end)
    |> Enum.filter(&File.exists?/1)
```

apps/remote_control/lib/lexical/remote_control/search/fuzzy.ex
```diff
@@ -336,31 +337,63 @@ defmodule Lexical.RemoteControl.Search.Fuzzy do
...
+  end
+
+  defp deps_roots(%Project{mix_project?: true} = project) do
+    # Note: This function assumes that the deps directories for all
+    # found projects is `deps`. Projects may override this directory
+    # and lexical won't understand this. This was done because loading
+    # each sub-project is expensive and changes our global directory.
+
+    root_deps_path = Mix.Project.deps_path()
+    {root_deps_path, Project.mix_exs_path(project)}
```
zachallaun, line 374: I think this is dead code or a leftover from debugging.
```suggestion
```

zachallaun (commented):
Made a couple suggestions.

I do think this is a better approach than the previous iteration. One downside is that it finds the existing deps directories once when fuzzy is initialized, which means that any directories added later (from project restructuring/etc. maybe?) won't be ignored.

scohen:
> One downside is that it finds the existing deps directories once when fuzzy is initialized

There's really no way to fix this. There is a lot of trouble with LSP file watching, several editors (emacs included) simply don't deliver `didChangeWatchedFiles` messages reliably, which is a shame.

zachallaun (approved):

PR #707, opened 2024-04-19
Fallback to `ElixirSense` when unable to find any definitions with Indexer

scottming:
But I don't know how to test this fallback more extensively, because `Patch` also cannot patch a call to a function within a project node, and I also don't want to introduce a dependency on ecto for navigation project.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
@@ -9,6 +9,7 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Definition do
   alias Lexical.RemoteControl.CodeIntelligence.Entity
   alias Lexical.RemoteControl.Search.Store
   alias Lexical.Text
+  require Logger
 
   @spec definition(Document.t(), Position.t()) :: {:ok, [Location.t()]} | {:error, String.t()}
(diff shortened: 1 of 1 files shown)


apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -43,6 +45,10 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Definition do
   end
 
   defp fetch_definition(_, %Analysis{} = analysis, %Position{} = position) do
+    fall_back_to_elixir_sense(analysis, position)
+  end
+
+  defp fall_back_to_elixir_sense(%Analysis{} = analysis, %Position{} = position) do
```
scohen, line 51: nitpick: `defp elixir_sense_definition`

scohen (approved):

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -32,7 +33,8 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Definition do
 
     case locations do
       [] ->
-        {:error, "No definition found for #{inspect(module)}"}
+        Logger.error("No definition found for #{inspect(module)} with Indexer.")
```
scohen, line 36: this is either a warning or an info.

PR #708, opened 2024-04-19
Added boost for matches directly following the last period

scohen:
If something matches starting after the last period (of a likely module) it is boosted. This makes a match for `MyModule.Structs.User` take precedence over `MyModule.User.Something` while querying for `User`

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/fuzzy/scorer.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/fuzzy/scorer.ex
@@ -23,7 +23,7 @@ defmodule Lexical.RemoteControl.Search.Fuzzy.Scorer do
 
   import Record
 
-  defrecord :subject, graphemes: nil, normalized: nil, last_period_position: nil
+  defrecord :subject, graphemes: nil, normalized: nil, period_positions: [-1]
 
   @typedoc "A match score. Higher numbers mean a more relevant match."
   @type score :: integer
@@ -47,12 +47,11 @@ defmodule Lexical.RemoteControl.Search.Fuzzy.Scorer do
       |> List.to_tuple()
 
     normalized = normalize(subject)
-    last_period = last_period_position(normalized)
 
     subject(
       graphemes: graphemes,
       normalized: normalized,
-      last_period_position: last_period
+      period_positions: period_positions(normalized)
     )
   end
 
@@ -171,7 +170,7 @@ defmodule Lexical.RemoteControl.Search.Fuzzy.Scorer do
 
     match_amount_boost = consecutive_count * pattern_length
 
-    match_boost = match_boost(score, subject)
+    match_boost = match_boost(score, subject, pattern_length)
 
     camel_case_boost = camel_case_boost(score.matched_character_positions, subject)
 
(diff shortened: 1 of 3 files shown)


scottming (commented):
I think this PR solves this problem:
> This makes a match for MyModule.Structs.User take precedence over MyModule.User.Something while querying for User

but it doesn't address the issue with *mixed case*. Instead, it pushes modules with capital letters further down. The screenshot illustrates the problem; the module with 'User' at the top is sorted towards the end.

CleanShot 2024-04-20 at 20 23 28@2x

You can also reproduce the issue with this unit test:

```elixir
    test "matches when there're mixed case characters" do
      results = score_and_sort(~w(create_user save_user Demo.Accounts.User), "User")
      assert results == ~w(Demo.Accounts.User create_user save_user)
    end
```

scohen:
@scottming i made some changes to the scoring that should improve the situations you've mentioned.

apps/remote_control/test/lexical/remote_control/search/fuzzy/scorer_test.exs
```diff
@@ -81,5 +86,11 @@ defmodule Lexical.RemoteControl.Search.Fuzzy.ScorerTest do
 
       assert ["First.Second.Third" | _] = results
     end
+
+    test "modules are boosted" do
```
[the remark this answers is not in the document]
scohen: The boost matches at the beginning of the string and boos matches after the last dot are conflicting with each other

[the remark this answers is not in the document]
scohen: This issue still gives you an ok sorting, and only affects nvim.

apps/remote_control/lib/lexical/remote_control/search/fuzzy/scorer.ex
```diff
@@ -150,36 +163,52 @@ defmodule Lexical.RemoteControl.Search.Fuzzy.Scorer do
...
   defp normalize(string) do
     String.downcase(string)
   end
 
+  @last_period_boost 15
+  @max_match_boost_boost 10
+  defp match_boost(%__MODULE__{} = score, subject(last_period_position: nil)) do
+    # penalize first matches further in the string by making them negative.
+    [first_match_position | _] = score.matched_character_positions
+    max(0 - first_match_position, @max_match_boost_boost)
```
zachallaun, line 195: Won't this always return `@max_match_boost_boost`, since this is essentially `max(something_negative, 10)` and `10` will always be greater? Is this meant to be `max(-first_match_position, -@max_match_boost_boost)`?

There's also an instance of this below.

zachallaun: Looking at the code, I'm wondering if this code should just always return `@max_match_boost_boost` (since it's a subject without a period, which is like the last segment matching). And then the clause below should maybe just return 0 because it doesn't meet the criteria for the boost?

scohen: Yes good catch.

scottming (approved):
I tested it and I think the latest version works really well, thank you. I believe that boosting the last module is extremely effective because once we pinpoint the module, we can narrow the scope and more quickly locate the specific symbol.

PR #709, opened 2024-04-20
chore: update FOD hash

hauleth:
It seems that there is still issue with autoupdater. It seems that
`main` is completely locked for direct commits, which mean that GitHub
Bot cannot update the hash automatically. I need to review the
possibility to automatically create PR from the GHA to handle that. But
for now we need to update hash manually.


zachallaun:
It looks like [this action](https://github.com/peter-evans/create-pull-request) can be used to automatically open a PR. @scohen may need to set up permissions for it. [Info here.](https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#workflow-permissions)

zachallaun (approved):

PR #715, opened 2024-04-23
Alias range

scohen:
When doing refactor operations like organize imports, it would be
helpful if we could get the entire range for an alias, because if we do,
replacing the alias is a trivial matter of replacing the range with an
empty string. Previously, aliases just had a line number, but this won't
work for aliases with blocks.


scohen:
Note, this is for after the 0.6 release.

apps/common/lib/lexical/ast/analysis/alias.ex
```diff
@@ -1,19 +1,61 @@
...
     Module.concat(alias.module)
   end
+
+  @implicit_aliases [:__MODULE__, :"@for", :"@protocol"]
+  defp range_for_ast(document, ast, _alias, as) when as in @implicit_aliases do
+    range_for_implicit_alias(document, ast)
+  end
+
+  defp range_for_ast(document, ast, alias, _as) do
+    if List.last(alias) == alias do
```
zachallaun, line 30: I think this must be a typo as it will always be false.

apps/common/lib/lexical/ast/analysis/alias.ex
```diff
@@ -1,19 +1,61 @@
...
+      range_for_implicit_alias(document, ast)
+    else
+      range_for_explicit_alias(document, ast)
+    end
+  end
+
+  defp range_for_explicit_alias(%Document{} = document, ast) do
+    case Ast.Range.fetch(ast, document) do
+      {:ok, %Range{end: end_pos} = range} ->
+        %Range{range | end: %Position{end_pos | character: end_pos.character}}
```
zachallaun, line 40: I found this pretty hard to parse, and I think it's actually a no-op? Breaking it down:

```elixir
# this update doesn't change end_pos
end_pos == %Position{end_pos | character: end_pos.character}
# which means the range update also doesn't change anything
range == %Range{range | end: end_pos}
```

apps/common/lib/lexical/ast/analysis/alias.ex
```diff
@@ -1,19 +1,61 @@
...
+
+  defp range_for_ast(document, ast, alias, _as) do
+    if List.last(alias) == alias do
+      range_for_implicit_alias(document, ast)
+    else
+      range_for_explicit_alias(document, ast)
+    end
+  end
+
+  defp range_for_explicit_alias(%Document{} = document, ast) do
```
zachallaun, line 37: For both this an `range_for_implicit_alias`, it would be helpful to have comments with examples of the specific ranges we're trying to extract.

apps/common/lib/lexical/ast/analysis/import.ex
```diff
@@ -76,4 +89,14 @@ defmodule Lexical.Ast.Analysis.Import do
 
   defp expand_function_keywords(_ignored, acc),
     do: acc
+
+  defp range_for_import(%Document{} = document, quoted) do
+    case Ast.Range.fetch(quoted, document) do
```
zachallaun, line 94: Let's add `Ast.Range.get(range, document, default \\ nil)` and then we can use that directly.

apps/remote_control/test/lexical/remote_control/analyzer/aliases_test.exs
```diff
@@ -247,6 +261,56 @@ defmodule Lexical.RemoteControl.Analyzer.AliasesTest do
...
+        |> scope_aliases()
+
+      assert decorate(doc, aliases[:Baz].range) =~
+               "  «alias Foo.Bar.{\n    Baz,\n    Quux,\n    Other\n}»"
+
+      assert decorate(doc, aliases[:Quux].range) =~
+               "  «alias Foo.Bar.{\n    Baz,\n    Quux,\n    Other\n}»"
+
+      assert decorate(doc, aliases[:Other].range) =~
+               "  «alias Foo.Bar.{\n    Baz,\n    Quux,\n    Other\n}»"
```
zachallaun, line 310: Putting these in a loop makes it easier to see at a glance that they're all meant to be equal.

```suggestion
      for name <- [:Baz, :Quux, :Other] do
        assert decorate(doc, aliases[name].range) =~
                 "  «alias Foo.Bar.{\n    Baz,\n    Quux,\n    Other\n}»"
      end
```

zachallaun (changes requested):

apps/common/lib/lexical/ast/analysis/alias.ex
scohen: Jeez, i'm sorry about this. This code was refactored from using `as` to using `alias`, and the original condition was `if List.last(alias) == as` and I blindly changed it using search and replace. Duh.

apps/common/lib/lexical/ast/analysis/alias.ex
scohen: I think the right thing here is to codify what explicit and implicit aliases are in code, and then things will be more clear. Also, I think i caught something odd. We define aliases for `@protocol` and `@for`, and those are actually module attributes.

apps/common/lib/lexical/ast/analysis/alias.ex
```diff
@@ -1,19 +1,52 @@
...
+  end
+
+  defp range_for_ast(document, ast, _alias, _as) do
+    # All other kinds of aliases defined with the `alias` special form
+    Ast.Range.get(ast, document)
+  end
+
+  defp implicit_range(%Document{} = document, ast) do
+    # There are kinds of aliases that are automatically generated by elixir
+    # such as __MODULE__, these don't really have any code that defines them,
```
zachallaun, line 41:
```suggestion
    # such as __MODULE__, these don't really have any code that defines them
```

zachallaun (approved):

PR #719, opened 2024-04-27
Fix conversion of char positions to UTF-16 code units

zachallaun:
We were previously using `CodeUnit.to_utf16`, which converts a *code unit* in a UTF-8 string to a code unit in the same string encoded as UTF-16, but we were passing in a *character position* which would lead to incorrect conversion of positions if any multi-byte characters were present before the position.

There were two tests of this existing behavior that seem to have been testing the incorrect behavior.

---

I'm marking this PR as draft for now because I'm also going to delete some `CodeUnit` conversion code that we should never be using (like `CodeUnit.to_utf16`) in a separate commit.

### DIFF
--- a/apps/common/lib/lexical/code_unit.ex
+++ b/apps/common/lib/lexical/code_unit.ex
@@ -19,7 +19,11 @@ defmodule Lexical.CodeUnit do
   @spec utf8_char_to_utf16_offset(String.t(), utf8_character_position()) ::
           utf16_code_unit_offset()
   def utf8_char_to_utf16_offset(binary, character_position) do
-    do_utf16_offset(binary, character_position, 0)
+    binary
+    |> String.slice(0, character_position)
+    |> :unicode.characters_to_binary(:utf8, :utf16)
+    |> byte_size()
+    |> div(2)
   end
 
   @doc """
@@ -70,25 +74,6 @@ defmodule Lexical.CodeUnit do
   defp do_count_utf8(<<c::utf8, rest::binary>>, count) do
     increment = code_unit_size(c, :utf8)
     do_count_utf8(rest, count + increment)
-  end
-
-  defp do_utf16_offset(_, 0, offset) do
-    offset
-  end
-
-  defp do_utf16_offset(<<>>, _, offset) do
-    # this clause pegs the offset at the end of the string
-    # no matter the character index
-    offset
-  end
-
-  defp do_utf16_offset(<<c, rest::binary>>, remaining, offset) when c < 128 do
-    do_utf16_offset(rest, remaining - 1, offset + 1)
-  end
-
-  defp do_utf16_offset(<<c::utf8, rest::binary>>, remaining, offset) do
-    increment = code_unit_size(c, :utf16)
-    do_utf16_offset(rest, remaining - 1, offset + increment)
   end
 
   # UTF-8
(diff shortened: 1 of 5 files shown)


apps/protocol/test/lexical/protocol/conversions_test.exs
```diff
@@ -87,8 +87,8 @@ defmodule Lexical.Protocol.ConversionsTest do
 
     test "single line utf8" do
       doc = doc("🏳️‍🌈abcde")
-      assert {:ok, pos} = Conversions.to_lsp(ex_position(doc, 1, 15))
-      assert %LSPosition{character: 6, line: 0} == pos
+      assert {:ok, pos} = Conversions.to_lsp(ex_position(doc, 1, 2))
+      assert %LSPosition{character: 2, line: 0} == pos
```
zachallaun, line 91: I'm not honestly sure where the `15` came from in the original test.

```
       🏳️‍🌈abcde
utf8   1 23456
utf16  0123456
```

apps/server/test/lexical/convertibles/lexical.plugin.diagnostic.result_test.exs
```diff
@@ -73,7 +73,7 @@ defmodule Lexical.Convertibles.Lexical.Plugin.V1.Diagnostic.ResultTest do
     test "it can translate a diagnostic that starts after an emoji", %{uri: uri} do
       assert {:ok, %Types.Diagnostic{} = converted} = to_lsp(plugin_diagnostic(uri, {6, 10}), uri)
 
-      assert converted.range == range(:lsp, position(:lsp, 5, 7), position(:lsp, 6, 0))
+      assert converted.range == range(:lsp, position(:lsp, 5, 10), position(:lsp, 6, 0))
```
zachallaun, line 76: It makes sense that the converted UTF16 code unit should be 10 because the input UTF8 position was 10 and there's a 2-code-unit character, so you get 10 - 1 (account for 0-based) + 1 (account for extra code unit in the emoji).

zachallaun:
@scohen I'm not sure that many people have run into this bug, but I expect that more will because things like workspace/document symbols makes it a lot more likely that the server return a range for a string with a multi-byte character in it (that's how I found it). All of that is to say: is it worth cutting a v0.6.1 after this has been merged and put through its paces for a few days?

apps/common/lib/lexical/code_unit.ex
```diff
@@ -1,63 +1,36 @@
...
-  Converts a utf8 character offset into a utf16 character offset. This implementation
-  clamps the maximum size of an offset so that any initial character position can be
-  passed in and the offset returned will reflect the end of the line.
+  Converts a 0-based UTF-8 character position to a UTF-16 code unit offset.
   """
-  @spec utf16_offset(String.t(), utf8_offset()) :: utf16_offset()
-  def utf16_offset(binary, character_position) do
+  @spec utf8_char_to_utf16_offset(String.t(), utf8_character_position()) ::
+          utf16_code_unit_offset()
+  def utf8_char_to_utf16_offset(binary, character_position) do
```
scohen, line 21: not a fan of the new name, `utf8` doesn't have the notion of `char`. Suggestion is `utf8_column_to_utf16_offset` or `utf8_cursor_to_utf16_offset`

scohen (changes requested):

apps/common/lib/lexical/code_unit.ex
zachallaun: How about `utf8_position_to_utf16_offset`?

scohen (approved):

PR #721, opened 2024-04-30
Fix the issue where `Entity.resolve` cannot resolve local capture function calls

scottming:


apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -105,6 +105,18 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
     end
   end
 
+  defp resolve({:local_arity, chars}, node_range, analysis, position) do
+    current_module = current_module(analysis, position)
+
+    case Ast.zipper_at(analysis.document, position) do
+      {:ok, %Zipper{node: {:/, _, [_, {:__block__, _, [arity]}]}}} ->
```
scohen, line 112: same comment, should you be guarding arity as an integer between 0 and 256?

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -465,4 +477,25 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
...
+    case Ast.zipper_at(analysis.document, position) do
+      {:ok,
+       %Zipper{
+         node: {:/, _, [{local_func_name, meta, _}, {:__block__, _, [arity]}]}
+       }} ->
+        start_column = meta[:column]
+        function_name_length = local_func_name |> to_string() |> String.length()
+
+        begin_pos = {position.line, start_column}
+        end_pos = {position.line, start_column + function_name_length}
```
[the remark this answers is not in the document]
scottming: No, let's stay consistent with the 'remote capture function',  and the cases like `&local_func(&1, x)`.

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -465,4 +477,25 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
       _ -> nil
     end
   end
+
+  defp maybe_local_capture_func(analysis, position) do
+    case Ast.zipper_at(analysis.document, position) do
+      {:ok,
+       %Zipper{
+         node: {:/, _, [{local_func_name, meta, _}, {:__block__, _, [arity]}]}
```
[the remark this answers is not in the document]
scottming: I guess what you're concerned about is this kind of situation with the numerator or denominator, which indeed is an edge case. 
```elixir
    test "failed at the position of the molecule" do
      code = ~q[
        defmodule Parent do
          def function do
            a = 1
            a|/1
          end
        end
      ]

      assert {:error, :not_found} = resolve(code)
    end
```

However, the guard does not help here; a more sensible approach is to judge its parent. This issue has been fixed in the latest commit.

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
scottming: I don't think we need this. We simply parse the correct part of the result(`capture function`) returned by `surround_context`. Based on that, we use a zipper to fetch the arity.

scohen: ok, understood

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -465,4 +477,24 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
...
+    with {:ok, %Zipper{node: {:/, _, [_, {:__block__, _, _}]}} = zipper} <-
+           Ast.zipper_at(analysis.document, position),
+         %Zipper{node: {:&, _, _}} <- Zipper.up(zipper) do
+      {:/, _, [{local_func_name, meta, _}, {:__block__, _, [arity]}]} = zipper.node
+      start_column = meta[:column]
+      function_name_length = local_func_name |> to_string() |> String.length()
+
+      begin_pos = {position.line, start_column}
+      end_pos = {position.line, start_column + function_name_length}
+      range = to_range(analysis.document, begin_pos, end_pos)
```
scohen, line 491: I think `Lexical.Ast.Range.fetch!` would help here
```suggestion
      function_name_length = local_func_name |> to_string() |> String.length()
      range = Ast.Range.fetch!(zipper.node, document)
      range = put_in(range.end.character, range.start.character + function_name_length)
```

I'm also pretty sure if you use zipper.node, you won't need to tack the function name length on the end.
this might just be
```elixir
range = Ast.Range.fetch!(zipper.node)

File: apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex [hunk 5]
```diff
@@ -465,4 +477,24 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
       _ -> nil
     end
   end
+
+  defp maybe_local_capture_func(analysis, position) do
+    with {:ok, %Zipper{node: {:/, _, [_, {:__block__, _, _}]}} = zipper} <-
+           Ast.zipper_at(analysis.document, position),
+         %Zipper{node: {:&, _, _}} <- Zipper.up(zipper) do
+      {:/, _, [{local_func_name, meta, _}, {:__block__, _, [arity]}]} = zipper.node
```
scohen, line 485: How can this match? The line right above says the node is `{:&, _, _}`

scottming: This is the parent node of that node.

scohen: I see

apps/remote_control/lib/lexical/remote_control/code_intelligence/entity.ex
```diff
@@ -465,4 +477,21 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Entity do
       _ -> nil
     end
   end
+
+  defp maybe_local_capture_func(analysis, position) do
+    with {:ok, %Zipper{node: {:/, _, [_, {:__block__, _, _}]}} = zipper} <-
+           Ast.zipper_at(analysis.document, position),
+         %Zipper{node: {:&, _, _}} <- Zipper.up(zipper) do
```
scohen, line 484: Can we wrap this cryptic match in a function called `inside_capture?` ?
This is one of those things that'll make little sense in six months.

scottming: Sure

PR #725, opened 2024-05-02
Organize aliases

### LINKED ISSUE #94: Feature: Code action - Organize aliases
This is actually a tough one. I tried implementing this in elixir-ls but gave up because it didn't have good enough source code editing tooling. 

Initially, we should enforce the following:

Flatten nested aliases `Foo.Bar{Baz, Quux}` 
Sort aliases alphabetically

Be aware though; Aliases can depend on one another
```elixir
alias Lexical.RemoteControl
alias RemoteControl.Build 
alias Build.Server
```
This should result in:
```elixir
alias Lexical.RemoteControl
alias Lexical.RemoteControl.Build
alias Lexical.RemoteControl.Build.Server
```
Similarly, nested aliases can have internal dependencies and external dependencies

Also, aliases can appear anywhere in a source file. Aliases inside function definitions cannot be pulled out of those definitions. Aliases that are present in an outer context are visible in an inner context. So
```elixir
defmodule Outer do 
  alias Lexical.RemoteControl 

  defmodule Inner do 
    def my_fun do 
      RemoteControl.foobar()
   end
end
```
is valid.

scohen:
First pass at an opinionated organize aliases code action. All aliases in a given scope are pulled to the top of the scope, or where the first alias occurs and are alphabetized and flattened.

Fixes #94

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/code_action/handlers/organize_aliases.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_action/handlers/organize_aliases.ex
@@ -82,10 +82,11 @@ defmodule Lexical.RemoteControl.CodeAction.Handlers.OrganizeAliases do
       |> Enum.map(fn %Alias{} = alias ->
         orig_range = alias.range
 
-        %Range{
-          start: %Position{orig_range.start | character: 1},
-          end: %Position{orig_range.end | line: orig_range.end.line + 1, character: 1}
-        }
+        orig_range
+        |> put_in([:start, :character], 1)
+        |> update_in([:end], fn %Position{} = pos ->
+          %Position{pos | character: 1, line: pos.line + 1}
+        end)
       end)
 
     first_alias_index = length(ranges) - 1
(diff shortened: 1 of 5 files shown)


apps/remote_control/lib/lexical/remote_control/code_action/handlers/organize_aliases.ex
```diff
@@ -0,0 +1,173 @@
...
+      scope, nil ->
+        scope
+
+      %Scope{id: :global}, %Scope{} = current ->
+        current
+
+      %Scope{} = next_scope, %Scope{} = current_scope ->
+        Enum.min_by([next_scope, current_scope], fn %Scope{} = scope ->
+          scope_start = scope.range.start
+          position.line - scope_start.line
```
scottming, line 169: It looks like you are trying to find the scope closest to the cursor position_line. If that's the case, why not name the function  `nearest_scope`?

scottming (approved):

apps/remote_control/lib/lexical/remote_control/code_action/handlers/organize_aliases.ex
scohen: That was the initial name, my thinking is that "near" has a couple meanings, for example
```elixir
SCOPE_1 do 
end
SCOPE_2 do 
end
```
Both SCOPE_1 and SCOPE_2 are near each other, but what I want to express is that
```elixir
SCOPE_1 do
  SCOPE_2 do 
  end
  SCOPE_3 do
  end
end
```
SCOPE_1 is the enclosing scope that's closest, and maybe that's the best name `narrowest_enclosing_scope`

scohen: Sure

PR #729, opened 2024-05-06
Indexed delegated functions

### LINKED ISSUE #724: Document symbols don't include defdelegate macros
As shown in the following screenshot, the document symbols don't include the `defdelegate` macro. Both the `save_changeset` and `update` functions are included, but `create` is not.

image

I'm not sure if this is intentional or not, as I've tried to create another macro myself and it is not included either.

scohen:
Functions created with defdelegate weren't being indexed, so they weren't showing up in find references, document symbols,  or go to definition.

This change emits two entries when it encounters a `defdelegate` call, one for the local function definition in the calling module, and a function reference to the delegated module. One possibly strange thing is that the function that's being created by defdelegate shows up as `def` in document symbols. This is technically correct, you're defining a public function, but it might be confusing in the UI.

Fixes #724

### DIFF
--- a/apps/remote_control/test/lexical/remote_control/search/indexer/extractors/function_definition_test.exs
+++ b/apps/remote_control/test/lexical/remote_control/search/indexer/extractors/function_definition_test.exs
@@ -4,7 +4,7 @@ defmodule Lexical.RemoteControl.Search.Indexer.Extractors.FunctionDefinitionTest
 
   def index(source) do
     do_index(source, fn %Entry{type: type} = entry ->
-      type in [{:function, :public}, {:function, :private}] and entry.subtype == :definition
+      match?({:function, _}, type) and entry.subtype == :definition
     end)
   end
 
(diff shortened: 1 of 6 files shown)


crbelaus:
I am testing this and can confirm that delegated functions are now included in the document symbols.

Here is a screenshot that shows this. Notice how the `translate/3` function is included along with the rest, even though it is a `defdelegate` instead of a `def` or `defp`.

image

This second screenshot is showing the "peek references" functionality of VSCode in the delegated module, which shows the `defdelegate` macro that points to it. So I believe this is working good as well.

image

---

I agree with you that it may be a little bit confusing to show the `defdelegate` as `def` in the document symbols. May be enough to include a note somewhere (maybe in the README?) about this as the reasoning is sound and you are, in fact, defining a public function.

scohen:
Note: #734  will allow us to mark delegates as delegates easily.

scohen:
@crbelaus we now show `defdelegate` in document symbols

apps/remote_control/test/lexical/remote_control/search/indexer/extractors/function_definition_test.exs
```diff
@@ -165,6 +165,54 @@ defmodule Lexical.RemoteControl.Search.Indexer.Extractors.FunctionDefinitionTest
...
+      assert "map(enumerable, other)" = extract(doc, function.range)
+      assert decorate(doc, function.range) =~ "defdelegate «map(enumerable, other)», to: Enum"
+    end
+
+    test "finds functions defined with defdelegate with an aliased module" do
+      {:ok, [function | _], doc} =
+        ~q[
+          defmodule MyModule do
+            alias Parent.Child
+            defdelegate map(enumerable, other), to: Child
```
scottming, line 188: My only wish for the delegate is that when the cursor is on `Mymodule.map|`, I hope it can quickly go to the definition of `Child.map` as well. So, in this case, should we return two definitions? With the existing LS, trying to jump to the original function's definition is really difficult, or even impossible to jump to.

scohen: We can definitely return two definitions for delegate functions.

scohen: oh heh, the reason it does this is because we're still using elixirsense to find definitions for functions.

scohen: ...and that should be a separate PR.

apps/remote_control/test/lexical/remote_control/search/indexer/extractors/function_reference_test.exs
```diff
@@ -112,6 +112,49 @@ defmodule Lexical.RemoteControl.Search.Indexer.Extractors.FunctionReferenceTest
...
+          alias Grandparent.Parent.Child
+          defdelegate map(enumerable, func), to: Child
+        ]
+        |> in_a_module()
+        |> index()
+
+      assert reference.subject == "Grandparent.Parent.Child.map/2"
+      assert reference.type == {:function, :usage}
+      assert reference.subtype == :reference
+      assert "defdelegate map(enumerable, func), to: Child" = extract(doc, reference.range)
```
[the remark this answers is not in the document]
scohen: This is intentional. The implementation intentionally doesn't include the call to `defdelegate`

apps/remote_control/test/lexical/remote_control/search/indexer/extractors/function_definition_test.exs
scohen: > When I find a function's definition, I might need to determine whether the definition's location is that of a delegate definition and its delegate reference, and then look for the original function's definition?

No, when you find a function's definition entry, you check the type of it, and if it's `{:function, :delegate}` then you know it's a delegated function and can look up the original. Another option would be to emit another definition entry for the function that's the target of `defdelegate`, then it'd just be two results for the definition call, but I don't know if that's semantically correct.

apps/remote_control/test/lexical/remote_control/search/indexer/extractors/function_reference_test.exs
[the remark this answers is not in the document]
scohen: To go into more detail, for things like functions, there's a block range and a detail range. The block range encompasses the entire function, while the detail is its arguments without the block (which includes `def`/`defp` and everything in the `do`/`end`. We use the detail to build the entry in the sidebar. `defdelegate` doesn't define a block, so it felt inappropriate to add a block range for it, and I felt it made sense to keep the detail focused on the call. 
If you're adamant, I suppose I could expand the notion of the block range to the entire `defdelegate` call, but I really don't like calling things blocks when they're not.

scohen:
I'm going to merge this, @scottming, you've had ample time to review this, two weeks for a small PR is too much.

PR #730, opened 2024-05-06
Update server's project after remote control init

scohen:
When remote_control bootstraps, it modifies the project struct with information about the mix project that it discovers. However, this update is not reflected in the server's project in its config, which is used whenever we make a call. Thus the two can differ, which is a bad thing.

This change fetches the project from remote control and updates it in the server's configuration after the bootstrap process has completed on the project node. They should be equal after this.

Fixes #727


scohen:
@soundmonster Try this branch and see if it fixes your problem.

PR #731, opened 2024-05-07
Stopped passing the server project over to remote_control

scohen:
RemoteControl has a global project object that's been augmented with extra information, but sometimes we pass server's project over to remote control. The server doesn't have this extra information, and in cases of compilation, it can cause spurious failures, like not being able to find a project's `.formatter.exs`

Fixes #727

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/api.ex
+++ b/apps/remote_control/lib/lexical/remote_control/api.ex
@@ -34,7 +34,7 @@ defmodule Lexical.RemoteControl.Api do
   end
 
   def format(%Project{} = project, %Document{} = document) do
-    RemoteControl.call(project, CodeMod.Format, :edits, [project, document])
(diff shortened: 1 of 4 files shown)


scohen:
@soundmonster this is a much simpler change that should also fix your issue.

soundmonster (approved):
Having a single source of truth for the project config makes a lot of sense 👍 I can confirm that it fixes the formatter issue I've been having. Thank you!

zachallaun (approved):
This looks good to me. 

Something that comes to mind that would be a little more verbose in places but would ensure that the correct project is always used in remote control: `RemoteControl.call(project, m, :f, [...])` could always pass the project as the first argument, so the actual call would end up being `m.f(RemoteControl.get_project(), ...)`. Some functions would have to ignore this argument, but it would be a stronger guarantee that functionality added in the future uses the right thing.

PR #734, opened 2024-05-09
Increased type detail

scohen:
Switched from a single atom to a tuple for types, which will be easier to generically match on and will allow us to provide richer type information for some follow up PRs.

Note, this will require a reindex of source code


scohen:
@scottming, @zachallaun ping!

scottming (approved):

PR #736, opened 2024-05-13
Proxy for expensive API requests

scohen:
When doing complicated refactors, the language server will often spam remote_control with a ton of operations when the refactor is going on. These operations will slow down both the editor and remote_control so that it's not particularly usable. In order to mitigate this, certain operations, like project compilation, can be buffered and completed when the refactor has completed.

This PR implements a proxy module that buffers these complicated operations, and monitors the calling process. When that process exits, the buffered operations are emitted in remote_control, allowing them to complete without disrupting the refactor.


apps/remote_control/lib/lexical/remote_control/api/proxy/records.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Lexical.RemoteControl.Api.Proxy.Records do
+  import Record
+
+  defrecord :message, body: nil
+  defrecord :mfa, module: nil, function: nil, arguments: []
```
zachallaun, line 5: After spending some time looking through `Proxy` and `Proxy.State`, I'm thinking that `mfa` is the wrong kind of record. `mfa` is this super generic thing that doesn't (in my opinion) enforce what we want, which is that the mfas should only ever contain certain calls that we're specifically equipped to handle. (The implementation of `collapse_commands` in the proxy state, for instance, just drops commands it doesn't expect.)

I think it would be more clear and actually simplify some code if we had records for each command we're expecting to proxy.

```elixir
defrecord :schedule_compile, [:project, :force?]
defrecord :compile_document, [:project, :document]
defrecord :reindex
defrecord :reindex_running?
defrecord :format, [:document]
```

Maybe the name of this module becomes `Command` instead of `Records` as well and then has a `run/1` that can run each kind of thing:

```elixir
def run(message(body: body)) do
  RemoteControl.Dispatch.broadcast(body)
end

def run(schedule_compile(project: project, force?: force?)) do
  RemoteControl.Build.schedule_compile(project, force?)
end

...
```

Then the `apply` call in `Proxy` is just replaced by `Command.run/1` and becomes:

```elixir
state
|> State.flush()
|> Enum.each(&Command.run/1)
```

zachallaun (changes requested):

apps/remote_control/lib/lexical/remote_control/api/local.ex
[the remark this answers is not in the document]
scohen: Same here, I struggled mightily here. One thought was to put all of this directly into `RemoteControl`, but I also didn't like that, since it didn't have a clean separation.

Suggestions welcome.

apps/remote_control/lib/lexical/remote_control/api/proxy.ex
```diff
@@ -0,0 +1,180 @@
...
+  end
+
+  # proxied functions
+
+  def broadcast(percent_progress() = message) do
+    RemoteControl.Dispatch.broadcast(message)
+  end
+
+  def broadcast(message) do
+    message = message(body: message)
```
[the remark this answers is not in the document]
scohen: ehh, that's not equivalent. I'm creating a message here, not matching on one.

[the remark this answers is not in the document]
zachallaun: My fault, misread it.

apps/remote_control/lib/lexical/remote_control/api/proxy/state.ex
```diff
@@ -0,0 +1,153 @@
...
+  alias Lexical.Identifier
+  alias Lexical.RemoteControl.Api
+  alias Lexical.RemoteControl.Build
+  alias Lexical.RemoteControl.Commands
+
+  import Api.Messages
+  import Api.Proxy.Records
+  import Record
+
+  defrecord :buffered, id: nil, message: nil, type: nil
```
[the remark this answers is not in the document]
zachallaun: It seems like explicit sorting shouldn't be necessary. Records are coming into the proxy in a certain order and then are added to the buffer in that order, so we know that `state.buffer` will always be a list of reverse-chronological records. It seems like that ordering (or the reverse of that) could be retained when commands are being collapsed.

apps/remote_control/lib/lexical/remote_control/api/proxy/state.ex
```diff
@@ -0,0 +1,153 @@
...
+    |> Enum.concat(collapse_messages(messages, project_compile, document_compiles))
+    |> Enum.filter(&match?(buffered(), &1))
+    |> Enum.sort_by(fn
+      buffered(message: mfa(module: Commands.Reindex)) ->
+        # atoms are always greater than integers, so this will go last
+        :reindex
+
+      buffered(id: id) ->
+        id
+    end)
```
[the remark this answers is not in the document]
scohen: we split apart the messages and the commands earlier so we can dedupe the messages, the overall ordering will get lost.

apps/remote_control/lib/lexical/remote_control/api/proxy/records.ex
scohen: the mfa is a deeply internal API to proxy and Proxy.State, how will this abstraction leak out through the public APIs ? 

One of the issues you flagged earlier was that there were too many records. What is the advantage of making what are effectively mfa records with a 1:1 correspondence to functions elsewhere?

scohen: so, i think that this suggestion will have a knock-on effect in `Proxy.State` that will make it more complicated. Rather than checking for `mfa()` records inside of flush() we'll now need to check for each of the command records. Correct?

scohen:
@zachallaun can you please respond to the comments?

apps/remote_control/test/support/lexical/test/mfa_support.ex
zachallaun, line 1: Doesn't look like this is being used?

zachallaun (approved):
I made one other comment about a `Lexical.Test.MfaSupport`, which I don't think is being used. Otherwise looks good.

apps/remote_control/lib/lexical/remote_control/api/proxy/state.ex
```diff
@@ -0,0 +1,146 @@
...
+      reindex: reindex
+    } = grouped
+
+    project_compile =
+      Enum.reduce(project_compiles, nil, fn
+        mfa(arguments: [_, true]) = mfa, _ ->
+          mfa
+
+        mfa(arguments: [true]) = mfa, _ ->
+          mfa
```
zachallaun, line 87: Looking at `Build.schedule_compile/2`, the signature is `schedule_compile(project, force? \\ false)`, so I think this case shouldn't ever be hit?

scohen: so that's actually two MFAs, and I think both need to be accounted for, no?

scottming:
To summarize my exploration today, I found that removing `Build.with_lock` indeed avoids the deadlock issue. However, occasionally the formatter fails to recognize .formatter.exs. I'm not sure if it's because my [commit](https://github.com/lexical-lsp/lexical/compare/api-proxy...api-proxy-sc?expand=1) is missing something.

scottming:
Additionally, I tried integrating this PR with the Rename module branch, and I found that the renaming can be completed instantly. However, we can't avoid a series of `mix compile` commands queuing up, which puts a heavy load on the CPU. The current Rename PR does not compile even once before the renaming is completed.

@scohen I hope you can take a look at this [commit](https://github.com/lexical-lsp/lexical/commit/66591aa69aa2a2b212325e936fb2eea8f601dea4) when you have time; maybe I'm using the API.Proxy method incorrectly.

scohen:
@scottming removing the locks is the likely cause for the queueing you’re seeing. That’s not the correct path forward

PR #739, opened 2024-05-14
Complete callables without parens if present in locals_without_parens

### LINKED ISSUE #601: Functions listed in `locals_without_parens` should not have parens in their completions
This has been a minor annoyance for a while, but when we complete `assert`, the completion comes in as `assert(assertion)`, but it should be `asert assertion`.

zachallaun:
This is scratching an itch I had while I was doing some work on another library.

Fixes #601

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/completion.ex
+++ b/apps/remote_control/lib/lexical/remote_control/completion.ex
@@ -9,91 +9,6 @@ defmodule Lexical.RemoteControl.Completion do
 
   import Document.Line
 
-  @built_in_locals_without_parens [
-    # Special forms
-    alias: 1,
-    alias: 2,
-    case: 2,
-    cond: 1,
-    for: :*,
-    import: 1,
-    import: 2,
-    quote: 1,
-    quote: 2,
-    receive: 1,
-    require: 1,
-    require: 2,
-    try: 1,
-    with: :*,
-
-    # Kernel
-    def: 1,
-    def: 2,
-    defp: 1,
-    defp: 2,
-    defguard: 1,
-    defguardp: 1,
-    defmacro: 1,
-    defmacro: 2,
-    defmacrop: 1,
-    defmacrop: 2,
-    defmodule: 2,
-    defdelegate: 2,
-    defexception: 1,
-    defoverridable: 1,
-    defstruct: 1,
-    destructure: 2,
-    raise: 1,
-    raise: 2,
-    reraise: 2,
-    reraise: 3,
-    if: 2,
-    unless: 2,
-    use: 1,
-    use: 2,
-
-    # Stdlib,
-    defrecord: 2,
-    defrecord: 3,
-    defrecordp: 2,
-    defrecordp: 3,
-
-    # Testing
-    assert: 1,
-    assert: 2,
-    assert_in_delta: 3,
-    assert_in_delta: 4,
-    assert_raise: 2,
-    assert_raise: 3,
-    assert_receive: 1,
-    assert_receive: 2,
-    assert_receive: 3,
-    assert_received: 1,
-    assert_received: 2,
-    doctest: 1,
-    doctest: 2,
-    refute: 1,
-    refute: 2,
-    refute_in_delta: 3,
-    refute_in_delta: 4,
-    refute_receive: 1,
-    refute_receive: 2,
-    refute_receive: 3,
-    refute_received: 1,
-    refute_received: 2,
-    setup: 1,
-    setup: 2,
-    setup_all: 1,
-    setup_all: 2,
-    test: 1,
-    test: 2,
-
-    # Mix config
-    config: 2,
-    config: 3,
-    import_config: 1
-  ]
-
   def elixir_sense_expand(%Env{} = env) do
     {doc_string, position} = strip_struct_operator(env)
 
@@ -105,9 +20,7 @@ defmodule Lexical.RemoteControl.Completion do
       []
     else
       {_formatter, opts} = Format.formatter_for_file(env.project, env.document.path)
-
-      locals_without_parens =
-        Keyword.get(opts, :locals_without_parens, []) ++ @built_in_locals_without_parens
+      locals_without_parens = Keyword.fetch!(opts, :locals_without_parens)
 
       for suggestion <- ElixirSense.suggestions(doc_string, line, character),
           candidate = from_elixir_sense(suggestion, locals_without_parens),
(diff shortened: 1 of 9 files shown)


scohen:
This fixes #601 right? Awesome!

apps/remote_control/lib/lexical/remote_control/completion.ex
```diff
@@ -18,14 +19,46 @@ defmodule Lexical.RemoteControl.Completion do
...
+       when struct in [Candidate.Function, Candidate.Macro] do
+    atom_name = String.to_atom(candidate.name)
+
+    suppress_parens? =
+      locals_without_parens
+      |> Keyword.take([atom_name])
+      |> Keyword.values()
+      |> Enum.any?(&suppress_parens?(&1, candidate.arity))
+
+    if suppress_parens? do
```
scohen, line 49: this is equivalent to `%{candidate | parens?: not suppress_parens}`

apps/remote_control/lib/lexical/remote_control/completion/candidate.ex
```diff
@@ -4,7 +4,18 @@ defmodule Lexical.RemoteControl.Completion.Candidate do
...
+      :argument_names,
+      :arity,
+      :name,
+      :origin,
+      :type,
+      :visibility,
+      :spec,
+      :summary,
+      :metadata,
+      parens: true
```
scohen, line 17:
```suggestion
      parens?: true
```

apps/remote_control/lib/lexical/remote_control/completion/candidate.ex
```diff
@@ -21,7 +32,17 @@ defmodule Lexical.RemoteControl.Completion.Candidate do
...
+    defstruct [
+      :argument_names,
+      :arity,
+      :metadata,
+      :name,
+      :origin,
+      :spec,
+      :summary,
+      :type,
+      parens: true
```
scohen, line 44:
```suggestion
      parens?: true
```

apps/remote_control/lib/lexical/remote_control/completion/candidate.ex
```diff
@@ -38,7 +59,17 @@ defmodule Lexical.RemoteControl.Completion.Candidate do
...
+    defstruct [
+      :argument_names,
+      :arity,
+      :name,
+      :origin,
+      :type,
+      :visibility,
+      :spec,
+      :metadata,
+      parens: true
```
scohen, line 71:
```suggestion
      parens?: true
```

apps/remote_control/lib/lexical/remote_control/completion/candidate.ex
```diff
@@ -55,7 +86,18 @@ defmodule Lexical.RemoteControl.Completion.Candidate do
...
+      :argument_names,
+      :arity,
+      :doc,
+      :metadata,
+      :type,
+      :name,
+      :origin,
+      :signature,
+      :spec,
+      parens: true
```
scohen, line 99:
```suggestion
      parens?: true
```

apps/remote_control/lib/lexical/remote_control/completion.ex
```diff
@@ -18,14 +19,46 @@ defmodule Lexical.RemoteControl.Completion do
...
   end
 
+  defp from_elixir_sense(suggestion, locals_without_parens) do
+    suggestion
+    |> Candidate.from_elixir_sense()
+    |> maybe_suppress_parens(locals_without_parens)
+  end
+
+  defp maybe_suppress_parens(%struct{} = candidate, locals_without_parens)
+       when struct in [Candidate.Function, Candidate.Macro] do
```
scohen, line 40:
```suggestion
       when struct in [Candidate.Function, Candidate.Macro, Candidate.Typespec] do
```

scohen (commented):
This is awesome. I feel this pain every time I type `assert` and it adds `assert(`
Can we have some tests?

zachallaun:
@scohen Thanks for the feedback! Agreed on tests; it does need more.

I'm actually not 100% sure that `assert` is going to complete without parens because the ExUnit assertions aren't actually in `locals_without_parens`. That makes me wonder what the formatter rules are for macros in built-in packages. Do they all exclude parens?

scohen:
I tried this branch, and it still adds parens to all the exunit macros. I think this is because it only checks the `locals_withut_parens` and not `import_deps`.

zachallaun:
> I tried this branch, and it still adds parens to all the exunit macros. I think this is because it only checks the `locals_withut_parens` and not `import_deps`.

See my message above -- the ExUnit macros aren't included for whatever reason, I think we need to special-case some things. I'm going to look into how the Elixir formatter decided whether to put parens around calls.

Re: `import_deps`, they're expanded automatically into `locals_without_parens`. Here's what it looks like for a library of mine, for instance:

```elixir
iex(3)> Mix.Tasks.Format.formatter_for_file("lib/mneme.ex")
{#Function<14.79746710/1 in Mix.Tasks.Format.find_formatter_for_file/2>,
 [
   sigils: [],
   inputs: ["{mix,.formatter}.exs", "{config,lib}/**/*.{ex,exs}", "test/**/*.{ex,exs}",
    "test_integration/**/*.{ex,exs}", "examples/*.{ex,exs}"],
   locals_without_parens: [
     all: :*,
     check: 1,
     check: 2,
     property: 1,
     property: 2,
     auto_assert: :*,
     auto_assert_raise: :*,
     auto_assert_receive: :*,
     auto_assert_received: :*
   ],
   export: [
     locals_without_parens: [
       auto_assert: :*,
       auto_assert_raise: :*,
       auto_assert_receive: :*,
       auto_assert_received: :*
     ]
   ],
   import_deps: [:stream_data],
   plugins: [Styler],
   line_length: 98
 ]}
```

`all`, `check` and `property` come from `:stream_data`.

zachallaun:
Looks like the formatter [special-cases](https://github.com/elixir-lang/elixir/blob/main/lib/elixir/lib/code/formatter.ex#L60) a bunch of things. We should likely copy this list and can set `parens?: false` for those.

zachallaun:
@scohen Still need to add more tests but give things a try now.

scohen:
@zachallaun much better. This is good to go when the tests are done.

zachallaun:
@scohen Pushed some additional tests. Are there other cases you'd like to see?

scohen (approved):
consider my comment, and then you're good to go.

apps/remote_control/lib/lexical/remote_control/completion.ex
```diff
@@ -4,10 +4,96 @@ defmodule Lexical.RemoteControl.Completion do
   alias Lexical.Document
   alias Lexical.Document.Position
   alias Lexical.RemoteControl
+  alias Lexical.RemoteControl.CodeMod.Format
   alias Lexical.RemoteControl.Completion.Candidate
 
   import Document.Line
 
+  @built_in_locals_without_parens [
```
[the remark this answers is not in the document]
zachallaun: Makes sense! Just pushed that change -- if you could confirm that it does what you were thinking, I think we're good to go.

PR #744, opened 2024-05-18
Returns two definitions for delegate function's definition

scottming:
And change `go to definition` from elixir-sense to Indexer for functions.

## Peek definition
CleanShot 2024-05-18 at 11 00 45@2x

## Go to definition
CleanShot 2024-05-18 at 11 01 00@2x

PR #746, opened 2024-05-21
Fix go to definition behavior for same-name, same-arity functions by directing to the first function

scottming:
After this PR #744 was merged, I spent a whole day using the latest features. In the case of delegates, returning multiple definitions is undoubtedly useful, but in another scenario, such as when there are when clauses, returning multiple definitions is undoubtedly a distraction (in Neovim, I need to press the keys twice more and think about it). Therefore, the solution in this PR is to directly jump to the position of the first function for such cases where there are same-name, same-arity functions in the same file.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
@@ -52,7 +52,7 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Definition do
     definitions =
       mfa
       |> query_search_index(subtype: :definition)
-      |> Enum.flat_map(fn entry ->
+      |> Stream.flat_map(fn entry ->
         case entry do
           %Entry{type: {:function, :delegate}} ->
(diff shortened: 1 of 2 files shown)


apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -62,6 +63,7 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Definition do
             [entry]
         end
       end)
+      |> Enum.uniq_by(& &1.path)
```
scohen, line 66: won't `uniq_by(& &1.subject)` also take care of it?

scottming: > why's the sort before the uniq_by? Shouldn't you be sorting all the results?

I later realized that this was unnecessary, but I forgot to push the code yesterday.

> won't uniq_by(& &1.subject) also take care of it? 
Indeed, this is clearer.

scohen (approved):

PR #748, opened 2024-05-22
Code Action: Remove unused aliases

scohen:
This code action removes any unused aliases. It handles both single aliases where it deletes the entire line up to the next line, and multiple aliases, where it deletes the single alias in the list, handling edge cases when it deletes both the last entry in the list, deleting the previous line's comma and when there's only one entry in the list when it deletes the whole alias.

The approach here is much more complicated than I wanted, I described the reasons why it's that way in the module's documentation, hopefully, we can get a fix from sourceror soon and improve this.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/code_action/handlers/remove_unused_alias.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_action/handlers/remove_unused_alias.ex
@@ -31,6 +31,7 @@ defmodule Lexical.RemoteControl.CodeAction.Handlers.RemoveUnusedAlias do
   alias Lexical.RemoteControl.Analyzer
   alias Lexical.RemoteControl.CodeAction
   alias Lexical.RemoteControl.CodeAction.Diagnostic
+  alias Sourceror.Zipper
 
   import Record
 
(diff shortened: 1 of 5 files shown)


apps/remote_control/lib/lexical/remote_control/code_action/handlers/remove_unused_alias.ex
```diff
@@ -0,0 +1,191 @@
...
+    :multi_alias_range,
+    :removed_alias_range,
+    :alias_count
+  ]
+
+  defrecordp :single_alias_metadata, [:document, :range]
+  @behaviour CodeAction.Handler
+
+  @impl CodeAction.Handler
+  def actions(%Document{} = document, %Range{} = range, diagnontics) do
```
zachallaun, line 48: typo: diagnontics

apps/remote_control/lib/lexical/remote_control/code_action/handlers/remove_unused_alias.ex
```diff
@@ -0,0 +1,191 @@
...
+  end
+
+  @impl CodeAction.Handler
+  def kinds do
+    [:source]
+  end
+
+  @alias_regex ~r/unused alias (\w+)/
+  defp to_edit(%Document{} = document, %Position{} = position, %Diagnostic{} = diagnostic) do
+    with [[_, module_string]] <- Regex.scan(@alias_regex, diagnostic.message),
```
zachallaun, line 70: Recommend splitting out a function for this:

```elixir
  @alias_regex ~r/unused alias (\w+)/
  defp fetch_unused_alias(diagnostic) do
    case Regex.scan(@alias_regex, diagnostic.message) do
      [[_, module_string]] -> {:ok, String.to_atom(module_string)}
      _ -> :error
    end
  end
```

```suggestion
    with {:ok, last_segment} <- fetch_unused_alias(diagnostic),
```

PR #751, opened 2024-05-23
Completion: show type spec for struct fields

### LINKED ISSUE #749: Struct autocomplete: Show the type
Hey,

Lexical is fantastic! ❤️

Sorry if this issue has already been mentioned. Could not find anything in the issues.

When ElixirLS completes access to a struct field, it shows some type information (given that the struct has a spec):

Screenshot 2024-05-22 at 16 52 10

However, Lexical doesn't (the examples are from Neovim and VS Code respectively):

Screenshot 2024-05-22 at 16 49 55

Screenshot 2024-05-22 at 16 48 31

Is there an easy way to fix that? 🤔 I have very little experience in LSP but happy to jump in with some guidance.

kirillrogovoy:
Fixes #749 

`:type_spec` is already a part of the `elixir_sense_map`, so defining it in `defstruct` is enough.

If the struct doesn't have a type spec, it would show "Field" in my editor probably because that's how it handles `completion.detail == nil`

Let me know if it's more desirable to fall back to some string explicitly, e.g.:
```elixir
builder.plain_text(env, struct_field.name,
  detail: struct_field.type_spec || "Field",
  label: struct_field.name,
  kind: :field
)
```

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/completion/candidate.ex
+++ b/apps/remote_control/lib/lexical/remote_control/completion/candidate.ex
@@ -159,7 +159,7 @@ defmodule Lexical.RemoteControl.Completion.Candidate do
 
   defmodule StructField do
     @moduledoc false
-    defstruct [:call?, :name, :origin]
+    defstruct [:call?, :name, :origin, :type_spec]
 
     def new(%{} = elixir_sense_map) do
       struct(__MODULE__, elixir_sense_map)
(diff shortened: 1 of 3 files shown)


scohen:
I think "Field" is coming from the LSP "kind" of the completion, and detail is nil-able, which means it'll just not display anything. It might be interesting to display "dynamic" or "unknown type" there, but that might just be noisy. 

FYI, my editor (emacs) also displays "Field" on completions.

This looks great. Thanks for contributing!

PR #756, opened 2024-05-29
Track uses and requires

scohen:
For the ecto PR, it would be great to only extract schemas if the `Ecto.Schema` module is used, so to help with that, this PR adds uses and requires to the metadata in the analyzer

### DIFF
--- a/apps/common/lib/lexical/ast/analysis.ex
+++ b/apps/common/lib/lexical/ast/analysis.ex
@@ -7,8 +7,10 @@ defmodule Lexical.Ast.Analysis do
 
   alias Lexical.Ast.Analysis.Alias
   alias Lexical.Ast.Analysis.Import
+  alias Lexical.Ast.Analysis.Require
   alias Lexical.Ast.Analysis.Scope
   alias Lexical.Ast.Analysis.State
+  alias Lexical.Ast.Analysis.Use
   alias Lexical.Document
   alias Lexical.Document.Position
   alias Lexical.Document.Range
@@ -363,6 +365,40 @@ defmodule Lexical.Ast.Analysis do
   # wholesale import import MyModule
   defp analyze_node({:import, _meta, [{:__aliases__, _aliases, module}]} = quoted, state) do
     State.push_import(state, Import.new(state.document, quoted, module))
+  end
+
+  # require of a module using as
+  defp analyze_node(
+         {:require, _meta, [{:__aliases__, _, module}, options]} = quoted,
+         state
+       ) do
+    {_, as_module} =
+      Macro.prewalk(options, nil, fn
+        {:__aliases__, _, as} = ast, nil ->
+          {ast, as}
+
+        ast, acc ->
+          {ast, acc}
+      end)
+
+    State.push_require(state, Require.new(state.document, quoted, module, as_module))
+  end
+
+  # require a module
+  defp analyze_node(
+         {:require, _meta, [{:__aliases__, _, module}]} = quoted,
+         state
+       ) do
+    State.push_require(state, Require.new(state.document, quoted, module))
+  end
+
+  # use statement
+
+  defp analyze_node(
+         {:use, _meta, [{:__aliases__, _, module} | opts]} = use,
+         state
+       ) do
+    State.push_use(state, Use.new(state.document, use, module, opts))
   end
 
   # stab clauses: ->
--- a/apps/common/lib/lexical/ast/analysis/scope.ex
+++ b/apps/common/lib/lexical/ast/analysis/scope.ex
@@ -8,7 +8,9 @@ defmodule Lexical.Ast.Analysis.Scope do
     :range,
     module: [],
     aliases: [],
-    imports: []
+    imports: [],
+    requires: [],
+    uses: []
   ]
 
   @type import_mfa :: {module(), atom(), non_neg_integer()}
@@ -23,12 +25,22 @@ defmodule Lexical.Ast.Analysis.Scope do
         }
 
   def new(%__MODULE__{} = parent_scope, id, %Range{} = range, module \\ []) do
+    uses =
+      if module == parent_scope.module do
+        # if we're still in the same module, we have the same uses
+        parent_scope.uses
+      else
+        []
+      end
+
     %__MODULE__{
       id: id,
       aliases: parent_scope.aliases,
       imports: parent_scope.imports,
+      requires: parent_scope.requires,
       module: module,
-      range: range
+      range: range,
+      uses: uses
     }
   end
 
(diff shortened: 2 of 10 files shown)


apps/common/lib/lexical/ast/analysis.ex
```diff
@@ -365,6 +367,40 @@ defmodule Lexical.Ast.Analysis do
     State.push_import(state, Import.new(state.document, quoted, module))
   end
 
+  # require of a module using as
```
zachallaun, line 370: IMO examples like this are a lot more clear than writing out in English

```suggestion
  # require MyModule, as: Alias
```

apps/common/lib/lexical/ast/analysis.ex
```diff
@@ -365,6 +367,40 @@ defmodule Lexical.Ast.Analysis do
...
+          {ast, as}
+
+        ast, acc ->
+          {ast, acc}
+      end)
+
+    State.push_require(state, Require.new(state.document, quoted, module, as_module))
+  end
+
+  # require a module
```
zachallaun, line 387:
```suggestion
  # require MyModule
```

apps/common/lib/lexical/ast/analysis.ex
```diff
@@ -365,6 +367,40 @@ defmodule Lexical.Ast.Analysis do
...
+  defp analyze_node(
+         {:require, _meta, [{:__aliases__, _, module}]} = quoted,
+         state
+       ) do
+    State.push_require(state, Require.new(state.document, quoted, module))
+  end
+
+  # use statement
+
```
zachallaun, line 396:
```suggestion
  # use MyModule
```

apps/common/lib/lexical/ast/analysis/scope.ex
```diff
@@ -23,12 +25,22 @@ defmodule Lexical.Ast.Analysis.Scope do
         }
 
   def new(%__MODULE__{} = parent_scope, id, %Range{} = range, module \\ []) do
+    uses =
+      if module == parent_scope.module do
+        # if we're still in the same module, we have the same uses
```
zachallaun, line 30: Even if we're in an inner module, should we consider that we have the same uses? We do that for aliases, imports, requires. 

```elixir
defmodule Parent do
  use SomeModule

  defmodule Child do
    # anything imported/required/etc. from `use SomeModule` is still available here
  end
end
```

apps/common/lib/lexical/ast/analysis.ex
```diff
@@ -365,6 +367,40 @@ defmodule Lexical.Ast.Analysis do
...
+    {_, as_module} =
+      Macro.prewalk(options, nil, fn
+        {:__aliases__, _, as} = ast, nil ->
+          {ast, as}
+
+        ast, acc ->
+          {ast, acc}
+      end)
+
+    State.push_require(state, Require.new(state.document, quoted, module, as_module))
```
zachallaun, line 384: We have a `fetch_alias_as` helper defined in this module.

```suggestion
    case fetch_alias_as(options) do
      {:ok, as_module} ->
        State.push_require(state, Require.new(state.document, quoted, module, as_module)
        
      :error ->
        state
    end
```

apps/remote_control/lib/lexical/remote_control/analyzer.ex
```diff
@@ -1,15 +1,43 @@
...
+  def uses_at(%Analysis{} = analysis, %Position{} = position) do
+    analysis
+    |> Uses.at(position)
+    |> Enum.reduce([], fn %Use{} = use, acc ->
+      case expand_alias(use.module, analysis, position) do
+        {:ok, expanded} -> [expanded | acc]
+        _ -> [Module.concat(use.module) | acc]
+      end
+    end)
+  end
```
zachallaun, line 39: It looks like the `.module` for aliases/imports are expanded during analysis, which means we can just use a simple `defdelegate` for those, but not for requires/uses? Is there a reason we can't expand the module before the `%Require{}/%Use{}` is created in `analysis.ex`? It seems like we should either do all expansion there or none of the expansion there (i.e. refactor analyzing aliases/imports so that it's consistent).

apps/remote_control/lib/lexical/remote_control/analyzer/requires.ex
```diff
@@ -0,0 +1,26 @@
...
+      [%Scope{} = scope | _] ->
+        scope.requires
+        |> Enum.filter(fn %Require{} = require ->
+          require_end = require.range.end
+
+          if require_end.line == position.line do
+            require_end.character <= position.character
+          else
+            require_end.line < position.line
+          end
```
zachallaun, line 18: We have this logic duplicated or *nearly* duplicated in a lot of places, but sometimes with superficial differences that can make it hard to know whether the code is equivalent. For instance, this code here has `require_end OP position` whereas the similar code in `Uses` has `position OP use_end`.

Can we define a `Position.compare/2` and start using it instead?

```suggestion
          Position.compare(position, require.range.end) in [:eq, :gt]
```

This could be in a separate PR that could audit the codebase for other places where code like this could be rewritten in terms of `Position.compare/2`.

apps/remote_control/lib/lexical/remote_control/analyzer/requires.ex
```diff
@@ -0,0 +1,26 @@
...
+        end)
+        |> Enum.uniq_by(& &1.as)
```
zachallaun, line 20: I notice that `Requires` has this `uniq_by`, but `Uses` doesn't. Just wanted to make sure that's deliberate.

zachallaun (changes requested):

apps/common/lib/lexical/ast/analysis/scope.ex
scohen: this is the trick, uses can do a lot more than import and require, they can add functions to the current module and greatly affect its behavior, this is not carried through to nested modules. 
If we want to actually be accurate for imports and requires, we need to introspect the module.

apps/remote_control/lib/lexical/remote_control/analyzer/requires.ex
scohen: it was, my thinking is that you can require multiple things as long as the destination is unique, but I don't see a reason to do this with use.

apps/remote_control/lib/lexical/remote_control/analyzer/requires.ex
scohen: This is a good call,  was considering this very suggestion, but felt it was out of scope for the PR.

apps/remote_control/lib/lexical/remote_control/analyzer.ex
scohen: i was slightly worried about performance, but I think it's important to be consistent.

scohen: Actually, depending on what you mean by expanded, I don't think this is correct. The aliases are still a list of atoms, and not a module at this point. Calling expand_alias on them makes them into a module. Same thing with aliases, but they have a `to_module` function in them that I didn't think was necessary, since aliases and imports return maps when requires and uses just return a simple list.

PR #758, opened 2024-05-30
Added struct definition detection for ecto schemas

### LINKED ISSUE #695: Index `Ecto` schema struct
For example, when we define a module like:

```elixir
defmodule User do
  use Ecto.Schema

  schema "users" do
    field :name, :string
    field :age, :integer, default: 0
    field :password, :string, redact: true
    has_many :posts, Post
  end
end
```

and we can't go to the definition when the cursor at `%Use|r{}` right now, it is pretty annoying.

* https://hexdocs.pm/ecto/Ecto.Schema.html#schema/2
* https://hexdocs.pm/ecto/Ecto.Schema.html#embedded_schema/1
* https://hexdocs.pm/ecto/Ecto.Schema.html#embeds_many/3-inline-embedded-schema
* https://hexdocs.pm/ecto/Ecto.Schema.html#embeds_one/3

scohen:
We now have an extractor that emits a struct definition where ecto structs are defined.

Fixes #695

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/ecto_schema.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/ecto_schema.ex
@@ -58,7 +58,7 @@ defmodule Lexical.RemoteControl.Search.Indexer.Extractors.EctoSchema do
 
       definition =
         Entry.block_definition(
-          document,
+          document.path,
           Reducer.current_block(reducer),
           struct_module,
           :struct,
--- a/apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/ecto_schema.ex
+++ b/apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/ecto_schema.ex
@@ -6,36 +6,23 @@ defmodule Lexical.RemoteControl.Search.Indexer.Extractors.EctoSchema do
   alias Lexical.RemoteControl.Search.Indexer.Metadata
   alias Lexical.RemoteControl.Search.Indexer.Source.Reducer
 
-  @schema_types [:schema, :embedded_schema]
   def extract(
-        {schema_type, meta, [{:__block__, _, [_source]} | _]} = schema_block,
+        {:schema, meta, [{:__block__, _, [_source]} | _]} = schema_block,
         %Reducer{} = reducer
-      )
-      when schema_type in @schema_types do
-    document = reducer.analysis.document
-    position = Reducer.position(reducer)
+      ) do
+    case extract_schema_entry(schema_block, meta, reducer) do
+      {:ok, _} = success -> success
+      :error -> :ignored
+    end
+  end
 
-    with true <- defines_schema?(reducer, position),
-         {:ok, current_module} <- Analyzer.current_module(reducer.analysis, position),
-         {do_line, do_column} <- Metadata.position(meta, :do),
-         {:ok, range} <- Ast.Range.fetch(schema_block, document) do
-      detail_range = put_in(range.end, Position.new(document, do_line, do_column + 2))
-
-      definition_entry =
-        Entry.block_definition(
-          document.path,
-          Reducer.current_block(reducer),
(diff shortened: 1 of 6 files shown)


apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/ecto_schema.ex
```diff
@@ -0,0 +1,97 @@
...
+
+  defp block_form?(ast) do
+    {_, result} =
+      Macro.prewalk(ast, false, fn
+        {:__block__, _, [:do]}, false ->
+          {nil, true}
+
+        ast, acc ->
+          {ast, acc}
+      end)
```
zachallaun, line 93: Something that I think should go in a different PR:

We should add an `Ast.prewalk_while/3` that's similar to Sourceror's [`Zipper.traverse_while/3`](https://hexdocs.pm/sourceror/Sourceror.Zipper.html#traverse_while/3) in that the function can return `{:cont, ast, acc}`, `{:skip, ast, acc}`, or `{:halt, ast, acc}` so that we can short-circuit instead of walking the whole tree in cases like this.

```elixir
{_, result} =
  Ast.prewalk_while(ast, false, fn
    {:__block__, _, [:do]}, _ ->
      {:halt, nil, true}

    ast, acc ->
      {:cont, ast, acc}
  end)
```

apps/remote_control/test/lexical/remote_control/search/indexer/extractors/ecto_schema_test.exs
```diff
@@ -0,0 +1,258 @@
...
+      {:ok, [struct], _doc} =
+        ~q[
+         defmodule MySchema do
+           alias Ecto.Schema , as: SCM
+           use SCM
+           schema "my_schema" do
+             field :last_name, :string
+           end
+        end
+        ]
```
zachallaun, line 33: There's a lot of inconsistent indentation in the code examples in these tests. Can we clean it up?

```suggestion
        ~q[
          defmodule MySchema do
            alias Ecto.Schema , as: SCM
            use SCM
            schema "my_schema" do
              field :last_name, :string
            end
          end
        ]
```

zachallaun (changes requested):

apps/remote_control/test/lexical/remote_control/search/indexer/extractors/ecto_schema_test.exs
scohen: most of the inconsistent indentation in the expected cases is due to how extract works. Since it's pulling out a range that has a surrounding indentation, but the range pulls out the context so it starts with no indentation, however, the next line still contains the context's indentation.

apps/remote_control/lib/lexical/remote_control/search/indexer/extractors/ecto_schema.ex
scohen: how do you stop a prewalk though?

zachallaun: I'm not sure I understand the question.

scohen: Pre walk is implemented via Macro.prewalk, I don't know how to stop it, and it will iterate through all the ast nodes eventually

apps/remote_control/test/lexical/remote_control/search/indexer/extractors/ecto_schema_test.exs
```diff
@@ -0,0 +1,258 @@
...
+
+      assert [] == results
+    end
+
+    test "consisting of a single field" do
+      {:ok, [struct], doc} =
+        ~q[
+         defmodule MySchema do
+           use Ecto.Schema
+           embedded_schema "my_schema" do
```
scottming, line 119:
```suggestion
           embedded_schema  do
```

PR #760, opened 2024-05-30
Elixir 1.17

scohen:
Changed single quoted charlists to use the ~c sigil


zachallaun:
I see some errors from ElixirSense. If ElixirSense is using `:elixir_tokenizer`, we may not be able to complete this upgrade unless we update to the latest version of ElixirSense. I believe [this commit](https://github.com/elixir-lang/elixir/commit/d7ea2fa2e4e5de1990297be19495fc93740b2e8b) changed the return value of `:elixir_tokenizer.tokenize/4` to both return an extra value in the `:ok` tuple (`rev_terminators`) and to return tokens in reverse order.

zachallaun:
Yep, here's the commit in `elixir_sense` that adds support for the 1.17 tokenizer: https://github.com/elixir-lsp/elixir_sense/commit/a41777a8d47b64b5886f1c2e7b09e38c462ad3ff

kirillrogovoy:
Just wanted to say that, except for all the deprecation warnings, the latest `main` version (as of right now) of Lexical runs well on 1.17.0

I had to locally "allow" it though through editing boot.ex.

Maybe allow it for everyone today? Or is it actually buggy and I'm failing to see it?

scohen:
@kirillrogovoy are you using this branch?

apps/server/lib/lexical/server/boot.ex
```diff
@@ -86,12 +86,14 @@ defmodule Lexical.Server.Boot do
     "1.13.0" => ">= 1.13.0",
     "1.14.0" => ">= 1.14.0",
     "1.15.0" => ">= 1.15.3",
-    "1.16.0" => ">= 1.16.0"
+    "1.16.0" => ">= 1.16.0",
+    "1.17.0-rc" => ">= 1.17.0-rc"
```
bigardone, line 90: 👋🏼 "1.17.0" is already available :)

kirillrogovoy:
@scohen I was referring to the latest `main` where I locally added `"1.17.0" => ">= 1.17.0"` to boot.ex (not `-rc` anymore)

Haven't tried using this branch. My point was that, since 1.17.0 release is out now, a bunch of people are upgrading their setup (e.g. I now run 1.17.0 both locally and in production), and Lexical refuses to start due to that version check although technically it works well if you bypass the check. Or at least if feels like it to me.

So my suggestion was that maybe we could add that `"1.17.0" => ">= 1.17.0"` line to `main` already while fixing the compiler warnings here in this branch. Otherwise, many people might get caught off guard with no straightforward remedy. At, least, this way they can just `git pull`, start the LSP, and continue on with their work.

scohen:
The problem getting those changes merged is that we have a bunch of test failures under 1.17 that need to be addressed before we merge. They look legitimate to me, and will affect how errors are shown in the UI.

miguno:
I, too, am looking forward to Elixir 1.17 support. Thanks in advance for making it happen, even if it should take some time. Your work on lexical is much appreciated!

scottming:
@miguno @kirillrogovoy @bigardone I have fixed all the test errors for the 5 versions. Please pull the latest code from this branch and test it when you have a chance. If there are no further issues, we will merge it ASAP after the review.

cc @scohen @zachallaun

kirillrogovoy:
Thanks a lot @scohen!

I've just pulled, checked out the branch, and ran mix deps.get, mix package.

As far as I can tell, Lexical has compiled successfully, and all basic LSP functions work in my Neovim.

Will post any issues here.

bigardone:
👋🏼 @scottming, thanks for the update! Same as @kirillrogovoy here. I just pulled the branch, and everything seems to be working fine. 🙌🏼

Moosieus (approved):

PR #761, opened 2024-06-01
Fix the interference issue caused by organizing aliases.

scottming:
When I'm writing functions, I occasionally need to handle unused variables, but I find that organizing aliases still returns results. This causes a lot of interference, especially when there's a bug in nvim, as pressing keys quickly can mess up my code.

CleanShot 2024-06-01 at 10 19 06@2x

We should not return the results of organizing aliases when the cursor is not on a line with an alias or when the current token is not an alias.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/code_action/handlers/organize_aliases.ex
+++ b/apps/remote_control/lib/lexical/remote_control/code_action/handlers/organize_aliases.ex
@@ -17,7 +17,7 @@ defmodule Lexical.RemoteControl.CodeAction.Handlers.OrganizeAliases do
   @impl CodeAction.Handler
   def actions(%Document{} = doc, %Range{} = range, _diagnostics) do
     with {:ok, _doc, analysis} <- Document.Store.fetch(doc.uri, :analysis),
-         :ok <- check_aliases(doc, analysis, range) do
+         :ok <- check_aliases(analysis, range) do
       aliases = CodeMod.Aliases.in_scope(analysis, range)
       {insert_position, trailer} = CodeMod.Aliases.insert_position(analysis, range.start)
       edits = CodeMod.Aliases.to_edits(aliases, insert_position, trailer)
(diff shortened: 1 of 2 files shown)


apps/remote_control/lib/lexical/remote_control/code_action/handlers/organize_aliases.ex
```diff
@@ -37,9 +40,32 @@ defmodule Lexical.RemoteControl.CodeAction.Handlers.OrganizeAliases do
...
-      %Scope{aliases: [_ | _]} -> :ok
-      _ -> :error
+    with %Scope{aliases: [_ | _]} <- Analysis.module_scope(analysis, range),
+         true <-
+           ancestor_is_alias?(analysis, range.start) or
+             token_at_cursor_is_alias?(analysis, range.start) do
+      :ok
+    else
+      _ ->
+        :error
```
zachallaun, line 50: Perhaps we don't need to check whether the current module scope has aliases, since we're explicitly looking for an alias ancestor or current token?

```suggestion
    if ancestor_is_alias?(analysis, range.start) or token_at_cursor_is_alias?(analysis, range.start) do
      :ok
    else
      :error
    end
```

apps/remote_control/lib/lexical/remote_control/code_action/handlers/organize_aliases.ex
```diff
@@ -37,9 +40,32 @@ defmodule Lexical.RemoteControl.CodeAction.Handlers.OrganizeAliases do
...
+        true
+
+      _ ->
+        false
+    end)
+  end
+
+  defp token_at_cursor_is_alias?(%Analysis{} = analysis, %Position{} = position) do
+    project = RemoteControl.get_project()
+    {:ok, env} = Ast.Env.new(project, analysis, position)
```
zachallaun, line 68: Is this guaranteed to return `{:ok, env}`? If it doesn't, this function should probably just return `false` instead of crashing.

scohen:
I would also like organize aliases to show up if you're inside of a scope with aliases, for example
```elixir
defmodule WithAliases do
  alias Foo.Bar
  alias Baz.Quux

  |
end
```

scottming:
> I would also like organize aliases to show up if you're inside of a scope with aliases, for example
Do you have any suggestions regarding the implementation for it? It seems to conflict somewhat with the goals of this PR.

scohen:
the implementation would be  something like:
1. Get the scope of the cursor
2. determine if the immediate parent is a module
3. determine if that module has any aliases

scottming:
> determine if the immediate parent is a module

Checking this is a bit cumbersome because its ancestors include these before reaching `defmodule`. 
```elixir
{:__block__, [],
{{:__block__, [line: 1, column: 17], [:do]},
{{:__block__, [line: 1, column: 17], [:do]},
```


so I think maybe we could just check that both before and after the current cursor are blanks.

scohen:
> Checking this is a bit cumbersome because its ancestors include these before reaching 

This feels like a problem that can be solved.

scohen:
(also, blocks are annoying)

scottming:
>  This feels like a problem that can be solved.

Please take a look when you have a chance. The block is really annoying, but I think the last commit has limited room for improvement.

PR #776, opened 2024-06-18
Simplify and co-locate more completion abstractions in server

zachallaun:
This PR attempts to make completion code easier to understand and work on.

Previously, code that explicitly referred to completions lived in common, server, and remote_control; after this PR, it all lives in server and remote_control. Because server can refer to protocol types, I was also able to dramatically simplify the types/specs used in the builder.

Additionally, two behaviours have been removed in favor of referring to their implementations directly: `Lexical.Ast.Environment` (which was only implemented by `Lexical.Ast.Env`), and `Lexical.Completion.Builder` (which was only implemented by `Lexical.Server.CodeIntelligence.Completion.Builder`). Removing some of this indirection also allowed Dialyzer to catch a few bugs in the types that lived in `Lexical.Ast.Environment` (now `Lexical.Ast.Env`).


Moosieus:
Found this from a conversation a few months ago:
> I remember why builder is a behaviour, plug-ins. If it wasn't, we'd have to marry releases of lexical-plugin to releases of lexical, and move builder code to lexical-plugin.

Not sure if the same applies to `Lexical.Ast.Env` though.

zachallaun:
@Moosieus Hmm, I'm not sure I'm following that. `lexical_plugin` only has a dependency on `lexical_shared`, and `lexical_shared` is the "root" so to speak: it has no dependencies on any other part of Lexical. That should mean it's safe to move around anything in `apps/*`, as nothing in `projects/*` depends on them.

scohen:
It's the other way. The intention behind these behaviours was that we could then move things _to_ `lexical_shared`, publish the behaviour there and have lexical_shared provide an implementation, and then lexical would provide the real implementation.

zachallaun:
> It's the other way. The intention behind these behaviours was that we could then move things _to_ `lexical_shared`, publish the behaviour there and have lexical_shared provide an implementation, and then lexical would provide the real implementation.

While I think that's reasonable, I think that this simplification is worth doing for now for a few reasons:

- We *haven't* moved it to `lexical_shared`, and to my knowledge it isn't anyone's priority
- There are questions about the utility/design of plugins anyways
- This simplifies things for current contributors at the *possible* expense of a slightly harder refactoring if we want to turn this into a behaviour later

scohen:
Yes, i'm generally on board with this simplification, just giving you the context.

Moosieus (approved):

PR #777, opened 2024-06-19
Refactor request handling

zachallaun:
The goal of this PR is to make request handling on the server simpler and easier to understand.

The commits are mostly atomic, so reviewing commit-by-commit may be best. Notes on some of the changes:

- Delete `Server.Provider.Env`, which for a long time has been a struct containing only `:project`, and instead pass projects directly to handlers. This removes a level of indirection and gets rid of any confusion between it and `Ast.Env`.
- Delete `Server.Provider.Queue.Supervisor` in favor of starting a `Task.Supervisor` explicitly and moving the couple helper functions directly into the module in which they're used.
- Stop converting request IDs to strings. According to the LSP spec, request IDs can be `string | integer`, but from everything I found, implementations should *not* consider `1 == "1"`. If there's a reason we need to do this conversion (e.g. to handle a badly-behaving client), we can add it back in, but it seemed like an unnecessary complication.
- Replace `Server.Provider.Queue` with `Server.TaskQueue`, which just operates on IDs and MFAs. This means `TaskQueue` doesn't have to know or care about handlers. The server now finds the handler for the request before calling `TaskQueue.add/2`, instead of the queue having to do it. This also means that adding to the queue always succeeds.


Moosieus (approved):
Tests look good + working for me locally. This is definitely a good stride towards ease-of-contribution.

apps/server/lib/lexical/server.ex
```diff
@@ -156,4 +157,42 @@ defmodule Lexical.Server do
       error -> {error, state}
     end
   end
+
+  # credo:disable-for-next-line Credo.Check.Refactor.CyclomaticComplexity
+  defp handler_for(%_{} = request) do
```
scohen, line 162: maybe we should call this `fetch_handler` ?

apps/server/lib/lexical/server/task_queue.ex
```diff
@@ -0,0 +1,201 @@
...
+
+  @doc false
+  def run_task(fun) when is_function(fun) do
+    Task.Supervisor.async_nolink(task_supervisor_name(), fun)
+  end
+
+  @doc false
+  def cancel_task(%Task{} = task) do
+    Task.Supervisor.terminate_child(task_supervisor_name(), task.pid)
+  end
```
scohen, line 143: why are run_task and cancel_task these here and not in the state?

scottming:
> Delete Server.Provider.Env, which for a long time has been a struct containing only :project, and instead pass projects directly to handlers. This removes a level of indirection and gets rid of any confusion between it and Ast.Env

It is worth noting that we used `client_name` in the rename PR. So, will we also need to add a `client_name` field to the `Project` struct after this pr merged?

zachallaun:
@scottming What if we remove `Server.Provider.Env` but, instead of passing `project` to handlers, we just pass the whole `%Server.Configuration{}`?

The `:client_name` comes from the `Server.Configuration` struct. It seems to me that `Server.Provider.Env` is just a subset, but I don't think we need to hide the configuration struct from modules in the server app.

apps/server/lib/lexical/server/task_queue.ex
zachallaun: Good question; they could and should be.

PR #778, opened 2024-06-19
Simplifications in `remote_control`

zachallaun:
Commits are atomic and can be reviewed individually.


apps/remote_control/lib/lexical/remote_control/build/project.ex
```diff
@@ -92,14 +92,7 @@ defmodule Lexical.RemoteControl.Build.Project do
     end
 
     with_progress "mix deps.compile", fn ->
-      deps_compile =
-        if Features.compile_wont_change_directory?() do
-          "deps.compile"
-        else
-          "deps.safe_compile"
-        end
```
scohen, line 100: can you explain removing this? We've vendored `compile` and I would like to not have to use that vendored compile forever.

scohen: ah, never mind, i see it, though it might be clearer if we call `Features.compile_wont_change_directory?` in safe_compile instead. Pretty soon, we won't know why we're doing that if statement.

scohen: also, `Features.compile_keeps_current_directory?` is a better name.

scohen:
Left one comment, otherwise it looks good.

scohen (approved):

PR #781, opened 2024-06-28
Handled formatting invalid module names

scohen:
I'm working in a project that generates invalid module names, which would crash lexical due to lexical using Module.split, which causes a crash when it gets an invalid module name.

Switched to use `String.split` instead.

### DIFF
--- a/projects/lexical_shared/lib/lexical/formats.ex
+++ b/projects/lexical_shared/lib/lexical/formats.ex
@@ -68,15 +68,10 @@ defmodule Lexical.Formats do
     string_name = Atom.to_string(module_name)
 
     if String.contains?(string_name, ".") do
-      module_part =
-        case string_name do
-          "Elixir." <> rest -> rest
-          other -> other
-        end
-
-      module_part
-      |> String.split()
-      |> Enum.join(".")
+      case string_name do
+        "Elixir." <> rest -> rest
+        other -> other
+      end
     else
       # erlang module_name
       ":#{string_name}"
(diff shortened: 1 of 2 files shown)


projects/lexical_shared/lib/lexical/formats.ex
```diff
@@ -68,8 +68,14 @@ defmodule Lexical.Formats do
...
-      module_name
-      |> Module.split()
+      module_part =
+        case string_name do
+          "Elixir." <> rest -> rest
+          other -> other
+        end
+
+      module_part
+      |> String.split()
```
zachallaun, line 78: I'm not sure I'm understanding what this is handling. It's splitting on whitespace, which should never be in the module name, right? So this ends up being a list of one element, and then the join just unwraps the element.

scohen: I'm working in a codebase where module names end up like :"Foo.Bar.Baz", it's very wrong, but legal.

scohen: Oh, never mind, I got ya.

projects/lexical_shared/test/lexical/formats_test.exs
```diff
@@ -15,6 +15,14 @@ defmodule Lexical.FormatsTest do
     test "it correctly handles an erlang module name" do
       assert ":ets" == Formats.module(:ets)
     end
+
+    test "it drops any `Elixir.` prefix" do
+      assert "Kernel.SpecialForms" == Formats.module(Elixir.Kernel.SpecialForms)
+    end
```
scohen, line 21: FYI, @Moosieus this this is basically the same thing as the above tests `Elixir.Kernel.SpecialForms` -> `Kernel.SpecialForms`

Moosieus (approved):

PR #783, opened 2024-06-30
Elixir 1.17 PR Review changes

Moosieus:
* Remove unused `require Logger` from `ast/tokens.ex`
* Add test for `Lexical.Ast.Tokens.prefix_stream/2`
* mix format
* Fix typo
* Update erlex to remove compiler warnings


apps/common/lib/elixir/features.ex
```diff
@@ -23,7 +23,7 @@ defmodule Elixir.Features do
     Version.match?(System.version(), ">= 1.16.0")
   end
 
-  def constains_end_of_expression? do
+  def contains_end_of_expression? do
```
scottming, line 26: In the latest commit, I ran mix format and renamed the function to `contains_set_theoretic_types?`. please rebase it when you have a chance.

Moosieus: I'm gonna throw some of my changes here into a housekeeping PR, which I think is more appropriate. The changes for Elixir 1.17 seem good, no need to block them with my kibitzing here.

PR #789, opened 2024-07-09
Update sourceror in common and server

Moosieus:


scohen:
You need to run `mix deps.update sourceror` and commit `mix.lock`

scohen (approved):

PR #791, opened 2024-07-11
Fix: give proper argument to `TaskQueue.add/2` in Server.handle_message

bangalcat:
This will fix the issue that appears in lexical.log
```
10:02:38.610 [error] ** (FunctionClauseError) no function clause matching in LXical.Server.Provider.Handlers.DocumentSymbols.handle/2 
    (lx_server 0.5.0) lib/lexical/server/provider/handlers/document_symbols.ex:13: LXical.Server.Provider.Handlers.DocumentSymbols.handle...
    (lx_server 0.5.0) lib/lexical/server/task_queue.ex:77: anonymous fn/4 in LXical.Server.TaskQueue.State.as_task/2
    (elixir 1.17.1) lib/task/supervised.ex:101: Task.Supervised.invoke_mfa/2
    (elixir 1.17.1) lib/task/supervised.ex:36: Task.Supervised.reply/4
```

### DIFF
--- a/apps/server/lib/lexical/server.ex
+++ b/apps/server/lib/lexical/server.ex
@@ -132,7 +132,7 @@ defmodule Lexical.Server do
   def handle_message(%_{} = request, %State{} = state) do
     with {:ok, handler} <- fetch_handler(request),
          {:ok, req} <- Convert.to_native(request) do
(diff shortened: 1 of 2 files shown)


scottming (approved):

zachallaun:
Thank you for the fix!

PR #793, opened 2024-07-15
Make it possible to override FOD hash for Flake users

ffloyd:
It's especially useful when Lexical installed on dev machine via local developer shell (direnv, etc). So I can override FOD hash and fix hash mismatch that can occur due to different Erlang/Elixir versions. For example:

```nix
        # Using pkgs.beam.packages is a more complex way than just use
        # pkgs.elixir but allows to ensure that all stack is built
        # against specific version of Erlang/OTP.
        otp = pkgs.beam.packages.erlang_26;
        erlang = otp.erlang;
        elixir = otp.elixir_1_17;

        # Use lexical from flake instead of nixpkgs, but override
        # Erlang & Elixir versions to match with our setup.
        lexical = ((lexical-flake.lib.mkLexical {erlang = otp;}).override {
          inherit elixir;
          fodHash = "sha256-XT8+fcYDF4z5EBNx6DgPEljyvkD5NH08sBqYwEOGvLI=";
        });
```


scohen:
@hauleth thoughts?

ffloyd:
Hey! Any updates here?

PR #795, opened 2024-07-19
Ensure the remote control node and server use the same Elixir executable

zachallaun:
This PR closes #776.

Prior, we perform version manager detection twice: once in the shell when starting the Lexical server, and once in Elixir when starting the remote control node. This PR deduplicates this logic by storing the Elixir executable path used to start the server in the application environment, which is then added to the project as an attribute to be used when the remote control node starts.

### DIFF
--- a/integration/boot/set_up_mise.sh
+++ b/integration/boot/set_up_mise.sh
@@ -27,7 +27,6 @@ chmod +x ./mise
 eval "$(./mise activate bash)"
 
 export KERL_CONFIGURE_OPTIONS="--disable-debug --without-javac --without-termcap --without-wx"
-./mise plugin install -y erlang
 ./mise use --global "erlang@$ERLANG_VERSION"
 
 ./mise plugins install -y elixir


zachallaun:
I'm not 100% sure that storing the `:elixir_executable` in `%Project{}` is the right way to go. An alternative might be to use `:lexical` instead of `:server` as the config app, and then change `RemoteControl.Port` to do something like:

```elixir
exe = Application.get_env(:lexical, :elixir_executable, System.find_executable("elixir"))
```

instead of using `project.elixir_executable`.

scohen:
>  An alternative might be to use :lexical instead of :server as the config app,

There is no `:lexical` app though, I don't think this will be allowed.

zachallaun:
> > An alternative might be to use :lexical instead of :server as the config app,
> 
> There is no `:lexical` app though, I don't think this will be allowed.

That settles that. The existing direction probably makes the most sense then. I really dislike the idea of something in the remote control app accessing the server app config.

We could set it in the remote control app config, but that also doesn't feel awesome.

PR #802, opened 2024-07-25
Completion: Use existing specs to infer function signatures and vice-versa

zachallaun:


apps/common/lib/lexical/ast/env.ex
```diff
@@ -193,4 +193,37 @@ defmodule Lexical.Ast.Env do
...
+  """
+  @spec prev_significant_position(t) :: Position.t() | nil
+  def prev_significant_position(%__MODULE__{} = env) do
+    find_significant_position(env.document, env.position.line - 1, -1)
+  end
+
+  defp find_significant_position(%Document{} = document, line, inc_by) do
+    case Document.fetch_text_at(document, line) do
+      {:ok, text} ->
+        case Regex.run(~r/(^\s*)\S+/, text) do
```
scohen, line 216: does this need a regex? I might be micro-optimizing, but scanning forward in the string to find whitespace characters should be much faster and easy to implement.

apps/common/lib/lexical/ast/env.ex
```diff
@@ -193,4 +193,37 @@ defmodule Lexical.Ast.Env do
...
+
+  @doc """
+  Returns the position of the next non-whitespace token on a line before `env.position`.
+  """
+  @spec prev_significant_position(t) :: Position.t() | nil
+  def prev_significant_position(%__MODULE__{} = env) do
+    find_significant_position(env.document, env.position.line - 1, -1)
+  end
+
+  defp find_significant_position(%Document{} = document, line, inc_by) do
```
scohen, line 213: maybe give this fetch semantics?

apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
```diff
@@ -596,6 +568,57 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.Macro do
...
+         {:"::", _, [{name, _, args}, _return]} <- typespec do
+      if is_list(args) do
+        {:ok, name, length(args)}
+      else
+        {:ok, name, 0}
+      end
+    else
+      _ -> :error
+    end
+  end
```
scohen, line 620: should this go into an Ast.Typespec module?

apps/server/test/lexical/server/code_intelligence/completion/translations/macro_test.exs
```diff
@@ -46,6 +46,63 @@ defmodule Lexical.Server.CodeIntelligence.Completion.Translations.MacroTest do
...
+               |> fetch_completion("defp ")
+
+      assert apply_completion(completion) == ~q[
+        @spec my_function(term(), term()) :: term()
+        defp my_function(${1:arg_1}, ${2:arg_2}) do
+          $0
+        end
+      ]
+    end
+
```
scohen, line 105: how about a spec like `@spec my_function(first :: term(), second :: term()) :: term()

zachallaun: I'm not completely sure which case this suggestion applies to. Do you mean that we could use existing function parameter names when generating specs, e.g.

```elixir
@spe|
def my_function(x, y, z) do
  ...
end

# becomes

@spec my_function(x :: term(), y :: term(), z :: term()) :: term()
def my_function(x, y, z) do
  ...
end
```

I think this is worth doing, though we won't always be able to infer the name (we can handle the cases `var`, `var = PATTERN`, `PATTERN = var`).

Or did you mean something else?

apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
zachallaun: Maybe? I think Ast.Typespec might be too specific, since some of the logic here could/would be shared for similar, non-spec AST. e.g. getting the name and arity for any call (we do this for entity resolution). I can't think of a name just yet, but I'm thinking a more general Ast.Something with functions that can be easily composed using `with` (fetch semantics). So this particular function would end up something like

```elixir
with {:ok, spec_ast} <- Something.typespec_at(analysis, position),
     {:ok, call} <- Something.typespec_call(spec_ast) do
  Something.call_name_and_arity(call)
end
```

I think that having these in one module will prevent having to have *another* module for them to share helper functions and will make them easier to find.

zachallaun: While I was trying to come up with a name for this module, it dawned on me that the correct name is already there: `Ast.Detection`.

What if we expanded the detection behaviour to include `detect` as well as `detected?`. It would return `{:ok, node}` or `:error`. It might make the most sense as an optional callback, since there are some detectable things that don't necessarily have a node that encapsulates them, e.g. comments.

apps/server/test/lexical/server/code_intelligence/completion/translations/macro_test.exs
zachallaun: The opposite is also possible:

```elixir
@spec my_function(x :: term(), y :: term(), z :: term()) :: term()
def|

# becomes

@spec my_function(x :: term(), y :: term(), z :: term()) :: term()
def my_function(x, y, z) do
  |
end
```

apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
scohen: Love the name.

apps/server/test/lexical/server/code_intelligence/completion/translations/macro_test.exs
scohen: You understood my meaning. Inferring the names of the arguments seems like a win.

apps/server/lib/lexical/server/code_intelligence/completion/translations/macro.ex
zachallaun: Discussed on Discord; going to save this for another PR.

scohen (approved):

Issue #780, opened 2024-06-25
Best-effort go-to-definition if current arity doesn't match


zachallaun:
I was doing a refactor starting at a call site where I removed an argument to a function, so the call went from e.g. arity 4 to arity 3. If you then try to go-to-definition for that call, it fails because `name/3` doesn't exist.

In these cases, I think it would be an improvement to go to the definition that *does* exist, even if the arity doesn't match.

There's a question of which definition to go to. You could say "closest arity", but if you go-to-definition for `name/4` and there's a `name/3` and `name/5`, you still have to somewhat-arbitrarily pick. For simplicity, I'm inclined to go to the lowest-arity existing definition for `name`. This means that it would go to `name/0` if that existed, even though it's probably not what you want, but I think it's the most understandable.

PR #807, opened 2024-07-30
Provide next best definition when no exact match is present.

Moosieus:
Pursuant to #780


apps/common/lib/lexical/ast/analysis.ex
```diff
@@ -116,6 +116,20 @@ defmodule Lexical.Ast.Analysis do
...
+
+    Enum.reduce_while(enclosing_scopes, first_scope, fn
+      %Scope{module: same} = current, %Scope{module: same} ->
+        {:cont, current}
+
+      _, current ->
+        {:halt, current}
+    end)
+  end
+
```
zachallaun, line 132: Unless I'm misunderstanding this function, I don't think it's needed. Look for `Analyzer.current_module(analysis, position)` as a replacement.

apps/common/lib/lexical/ast/analysis/scope.ex
```diff
@@ -83,4 +83,8 @@ defmodule Lexical.Ast.Analysis.Scope do
   def end_line(%__MODULE__{} = scope, :end), do: scope.range.end.line
   def end_line(_, %Position{} = position), do: position.line
   def end_line(_, line) when is_integer(line), do: line
+
+  def module_alias(%__MODULE__{module: module}) do
+    Module.concat(module)
+  end
```
zachallaun, line 89: I also think that `Analyzer.current_module/2` does this, so this is probably not needed either.

apps/remote_control/test/lexical/remote_control/code_intelligence/definition_test.exs
```diff
@@ -433,6 +432,27 @@ defmodule Lexical.RemoteControl.CodeIntelligence.DefinitionTest do
     end
   end
 
+  describe "definition/2 when no exact is available" do
+    setup [:with_referenced_file]
+
+    test "", %{project: project, uri: referenced_uri} do
```
zachallaun, line 438: Missing test name

zachallaun (changes requested):
We should also add some tests to ensure that, when multiple arities are available, it goes to the correct one. I.e. `|some_call(1, 2)` should go to `some_call/1` if that's all that's there, but we should test that it still goes to `some_call/2` if there's both an arity-1 and arity-2 definition.

Moosieus:
I'll check on the test failure :mag_right: 

Go-to-definition supports multiple results, so we could just show all arities and let users select one of  them, rather than resolving to any specific one. If that's not preferable though, we can resolve to the lowest arity definition.

LMK what you think and I'll amend things accordingly.

apps/common/lib/lexical/ast/analysis.ex
zachallaun: I think this one still needs to be removed!

apps/remote_control/lib/lexical/remote_control/code_intelligence/definition.ex
```diff
@@ -49,20 +50,30 @@ defmodule Lexical.RemoteControl.CodeIntelligence.Definition do
...
+      case definitions do
+        [_ | _] ->
+          definitions
+
+        _ ->
+          # feat: search for next best definition when no exact match is present.
+          {:ok, module_at_position} = Analyzer.current_module(analysis, position)
+
+          call_prefix = Formats.mf(module, function)
+          definitions = query_search_index_prefix(call_prefix, subtype: :definition)
```
scottming, line 65: Directly querying the prefix is likely to return multiple definitions. Just like zach said in that issue: we should return the one closest to the target. 

Also, the matching on line 56 is problematic. When a module has multiple functions with the same name, we should only return one definition.

In short, we should return **only one definition** whenever possible (this allows users to jump easily), except for defdelegate.

scottming (changes requested):

Moosieus:
Found this behavior which applies to main branch as well. Public calls to non-existent arities can jump to privately defined arities.
[Screencast from 2024-08-03 23-24-56.webm](https://github.com/user-attachments/assets/d6f6770d-c854-414f-b236-8be2281ebd01)

This comes from both Lexical's existing ("exact") lookup, and elixir_sense. At the moment, I've filtered out private definitions for public calls on Lexical's lookup ("exact" + "nearest arity"), and left elixir_sense as-is.

Issue #803, opened 2024-07-28
root_dir detection should not sniff around OS HOME dir


filipmnowak:
hi there, first of all, thank you for Lexical!

this is about example Neovim config introduced [here](https://github.com/lexical-lsp/lexical/pull/240), more specifically this part of it:

```lua
return lspconfig.util.root_pattern("mix.exs", ".git")(fname) or vim.loop.os_homedir()
```

apologies for quite authoritative issue title, but i really feel like it's bad idea that software i install pokes around root of my home directory.
it's quite dangerous and in case of a bug can cause some damage.

some "prior art" to consider:

- folks at [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/) came to same conclusion it seems: https://github.com/neovim/nvim-lspconfig/issues/44
- kind of similar case of systemd viewing HOME dir like any other: https://www.phoronix.com/news/systemd-tmpfiles-purge-drama

all in all, i get it, it's just an example config nobody is forcing me to use, but still, i feel like it would be safer to skip `vim.loop.os_homedir()` from it maybe?

scottming:
@filipmnowak I think you make a good point; it can definitely be removed. I’ve forgotten where I copied or vim.loop.os_homedir() from, but it should be unnecessary. 

Additionally, I noticed that someone already submitted a [PR](https://github.com/neovim/nvim-lspconfig/pull/2874) to `nvim-lspconfig` last year. Perhaps our nvim configuration doesn't need to be this complex anymore. However, I don’t have time to explore this further in the next few days. If you have the time, you’re welcome to submit a PR to help us update the nvim-related configuration.

PR #814, opened 2024-08-11
simplify default config

filipmnowak:
related to: https://github.com/lexical-lsp/lexical/issues/803

i was looking around for an optimal pattern (for example [here](https://github.com/neovim/nvim-lspconfig/blob/ff97d376b1d22b2eaf9274605531babf0cd0cf21/doc/lspconfig.txt#L355) and [here](https://github.com/neovim/nvim-lspconfig/blob/ff97d376b1d22b2eaf9274605531babf0cd0cf21/doc/server_configurations.md)) any my conclusion as of now is:

- it's quite hard to avoid opinionated example.
- for what it's worth, majority of example configs are rather simple.
- good example configuration probably should target first-time user (more experienced user will likely have their own patterns to follow and experience to implement those).

i did some basic testing with `v0.7.0` compiled against `Elixir 1.17.0 (compiled with Erlang/OTP 27)`.

### DIFF
--- a/pages/installation.md
+++ b/pages/installation.md
@@ -176,30 +176,15 @@ of whether you are using mason or others, you can use this
 configuration below as a reference:
 
 ```lua
-    local lspconfig = require("lspconfig")
-    local configs = require("lspconfig.configs")
-
-    local lexical_config = {
-      filetypes = { "elixir", "eelixir", "heex" },
-      cmd = { "/my/home/projects/_build/dev/package/lexical/bin/start_lexical.sh" },
-      settings = {},
-    }
-
-    if not configs.lexical then
-      configs.lexical = {
-        default_config = {
-          filetypes = lexical_config.filetypes,
-          cmd = lexical_config.cmd,
-          root_dir = function(fname)
-            return lspconfig.util.root_pattern("mix.exs", ".git")(fname) or vim.loop.os_homedir()
-          end,
-          -- optional settings
-          settings = lexical_config.settings,
-        },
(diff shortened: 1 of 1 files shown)


scohen:
@scottming how does this look to you?

pages/installation.md
```diff
@@ -176,30 +176,15 @@ of whether you are using mason or others, you can use this
...
-        },
-      }
-    end
-
-    lspconfig.lexical.setup({})
+require('lspconfig').lexical.setup {
+  cmd = { "my/home/projects/_build/dev/package/lexical/bin/start_lexical.sh" },
+  root_dir = function(fname)
+    return util.root_pattern("mix.exs", ".git")(fname) or vim.loop.cwd()
+  end,
```
scottming, line 183:
```suggestion
```
As I understand it, these lines are not necessary because the default config already contains similar settings, https://github.com/neovim/nvim-lspconfig/blob/97e72e3a3bd3ae2034c2884da1253341409c2076/lua/lspconfig/server_configurations/lexical.lua#L6-L8

filipmnowak: i am not sure; i think config in Lexical's own docs needs to be integral, and it kind of replaces / is alternative to what you are linking from `github.com/neovim/nvim-lspconfig`. so personally, i would keep this part in the PR. if you still want to remove it, please confirm @scottming.

scottming: It is fine

scottming (approved):

PR #819, opened 2024-08-16
Respect existing capitalization in defmodule completion

### LINKED ISSUE #786: Respect existing capitalization in `defmodule` completion
When completing `defmodule`, Lexical infers the default module name from the path and file name. For instance:

```elixir
# lib/my_app/foo/bar.ex
defmodule MyApp.Foo.Bar do
  |
end
```

How you capitalize acronyms in modules is a point of preference. For instance, `lib/my_app/http.ex` might contain `MyApp.Http` or `MyApp.HTTP`; both are equally valid. Lexical will always generate the camel-case version, which is fine.

However, if you've chosen uppercase and have an existing module `MyApp.HTTP` and then complete `defmodule` at `lib/my_app/http/request.ex`, the default module will be `MyApp.Http.Request`.

It should be possible for Lexical to infer that you want `MyApp.HTTP.Request` based on the existence of `MyApp.HTTP`.

mat-hek:
closes #786

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/modules.ex
+++ b/apps/remote_control/lib/lexical/remote_control/modules.ex
@@ -197,6 +197,41 @@ defmodule Lexical.RemoteControl.Modules do
     with_prefix("Elixir." <> prefix, mfa)
   end
 
+  @doc """
+  Adjusts module name capitalization to existing modules' names in the project.
+
+  For example, when passed `Project.Http.Server`, it returns:
+  - `Project.HTTP.Server`, if there's any module named `Project.HTTP` or `Project.HTTP.*`
+  - `Project.Http.Server` otherwise
+  """
+  @spec adjust_name_capitalization(String.t()) :: String.t()
+  def adjust_name_capitalization(name) do
+    name_downcase = String.downcase(name)
+    name_split = String.split(name, ".")
+    project_modules = all_modules()
+
+    closest_name_split =
+      project_modules
+      |> Enum.map(fn {module_string, _already_loaded?} ->
+        with "Elixir." <> ms <- module_string, do: ms
+      end)
+      |> Enum.max_by(fn module_string ->
+        :binary.longest_common_prefix([String.downcase(module_string), name_downcase])
+      end)
+      |> String.split(".")
+
+    prefix =
+      closest_name_split
+      |> Enum.zip(name_split)
+      |> Enum.take_while(fn {closest_name, name} ->
+        String.downcase(closest_name) == String.downcase(name)
+      end)
+      |> Enum.map(fn {closest_name, _name} -> closest_name end)
+
+    suffix = Enum.drop(name_split, length(prefix))
+    Enum.join(prefix ++ suffix, ".")
+  end
+
   defp apply_predicate(module_arg, {invoked_module, function, args}) do
     apply(invoked_module, function, [module_arg | args])
   end
--- a/apps/remote_control/lib/lexical/remote_control.ex
+++ b/apps/remote_control/lib/lexical/remote_control.ex
@@ -50,6 +50,10 @@ defmodule Lexical.RemoteControl do
 
   defdelegate modules_with_prefix(prefix, predicate), to: RemoteControl.Modules, as: :with_prefix
 
+  defdelegate adjust_module_name_capitalization(name),
+    to: RemoteControl.Modules,
+    as: :adjust_name_capitalization
+
   defdelegate docs(module, opts \\ []), to: CodeIntelligence.Docs, as: :for_module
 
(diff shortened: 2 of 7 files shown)


apps/remote_control/lib/lexical/remote_control/modules.ex
```diff
@@ -197,6 +197,41 @@ defmodule Lexical.RemoteControl.Modules do
...
+  @spec adjust_name_capitalization(String.t()) :: String.t()
+  def adjust_name_capitalization(name) do
+    name_downcase = String.downcase(name)
+    name_split = String.split(name, ".")
+    project_modules = all_modules()
+
+    closest_name_split =
+      project_modules
+      |> Enum.map(fn {module_string, _already_loaded?} ->
+        with "Elixir." <> ms <- module_string, do: ms
```
mat-hek, line 216: not sure if this is the way 🤔

zachallaun:
Thanks for the PR! Sorry about the delay in reviewing it -- I think everyone has been in a bit of a holding pattern after the announcement of the combined language server effort. But I think we should move on this and not let it languish!

I can't fully review this right now, but I did have one thought: 

What if, instead of looking for the longest existing prefix, we looked at existing capitalization on a per-segment basis? For instance, imagine you have a `Foo.HTTP` and now you're adjusting capitalization for `Bar.HTTP`. Currently, it wouldn't be adjusted because of the prefix match. However, we could store a map of segments, so it would have `%{"foo" => "Foo", "http" => "HTTP"}` and then you just iterate over the list of segments in the module you're adjusting and do something like `Map.get(case_map, lower_segment, original_segment)`.

There may be some edge cases I've not thought of, and we'd probably want to limit the case map to only modules in the current application as opposed to those in all dependencies. Thoughts? @scohen?

mat-hek:
> What if, instead of looking for the longest existing prefix, we looked at existing capitalization on a per-segment basis? 

Good idea. Currently, having `Foo.HTTP` we'd also adjust `Foo.Httpserver` -> `Foo.HTTPServer`. To preserve this, we can do both per-segment and per-prefix matching ;)

apps/remote_control/lib/lexical/remote_control/modules.ex
```diff
@@ -197,6 +197,41 @@ defmodule Lexical.RemoteControl.Modules do
...
+      closest_name_split
+      |> Enum.zip(name_split)
+      |> Enum.take_while(fn {closest_name, name} ->
+        String.downcase(closest_name) == String.downcase(name)
+      end)
+      |> Enum.map(fn {closest_name, _name} -> closest_name end)
+
+    suffix = Enum.drop(name_split, length(prefix))
+    Enum.join(prefix ++ suffix, ".")
+  end
```
zachallaun, line 233: I made a separate comment about a potentially different way of adjusting capitalization, but if we stick with this approach, I think this can be done with one iteration over existing modules:

```suggestion
  def adjust_name_capitalization(name) do
    name_downcase = String.downcase(name)

    {prefix, byte_length} =
      Enum.reduce(all_modules(), {"", 0}, fn
        {"Elixir." <> existing, _loaded?}, {prefix, length} ->
          new_length =
            :binary.longest_common_prefix([String.downcase(existing), name_downcase])

          if new_length > length do
            {binary_slice(existing, 0, new_length), new_length}
          else
            {prefix, length}
          end

        _, acc ->
          acc
      end)

    <<_::binary-size(byte_length), suffix::binary>> = name

    prefix <> suffix
  end
```

zachallaun:
> > What if, instead of looking for the longest existing prefix, we looked at existing capitalization on a per-segment basis?
> 
> Good idea. Currently, having `Foo.HTTP` we'd also adjust `Foo.Httpserver` -> `Foo.HTTPServer`. To preserve this, we can do per-prefix matching on the segments ;)

Cool! Maybe the "existing capitalizations" map can store title case segments, so having an existing `Foo.HTTPServer` would store `%{"foo" => "Foo", "http" => "HTTP", "server" => "Server"}`. It looks like the following regex does the job of correctly splitting title case, then the operation just becomes a nested map where we map over module segments split on `.` and then over title case segments:

```elixir
r = ~r"""
# last char was lower and next char is upper:
# FooBAR
#   ^
(?<=[a-z])(?=[A-Z])

# or
|

# last char was upper and next chars are upper followed by lower:
# FOOBar
#   ^
(?<=[A-Z])(?=[A-Z][a-z])
"""x

[
  "FooBarBaz",
  "FOOBarBaz",
  "FooBARBaz",
  "FooBarBAZ",
  "foobarbaz",
  "A",
  "AB",
  "Ab",
  "AbC",
  "aBC"
]
|> Enum.each(fn name ->
  name
  |> String.split(r)
  |> IO.inspect()
end)
```

PR #825, opened 2024-09-05
Server bottlenecks

scohen:
Fixed two bottlenecks in the server. 

1. The mix lock was using a much more contentious build lock rather than its own lock.
2. Using `Task.Supervisor.terminate_child` was causing timeouts when bulk operations happen.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/mix.ex
+++ b/apps/remote_control/lib/lexical/remote_control/mix.ex
@@ -1,7 +1,6 @@
 defmodule Lexical.RemoteControl.Mix do
   alias Lexical.Project
   alias Lexical.RemoteControl
-  alias Lexical.RemoteControl.Build
 
   def in_project(fun) do
     if RemoteControl.project_node?() do
@@ -17,7 +16,7 @@ defmodule Lexical.RemoteControl.Mix do
 
     old_cwd = File.cwd!()
 
-    Build.with_lock(fn ->
+    with_lock(fn ->
       try do
         Mix.ProjectStack.post_config(prune_code_paths: false)
 
@@ -47,5 +46,9 @@ defmodule Lexical.RemoteControl.Mix do
         File.cd!(old_cwd)
       end
     end)
+  end
+
+  defp with_lock(fun) do
+    RemoteControl.with_lock(__MODULE__, fun)
   end
 end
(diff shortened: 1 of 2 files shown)


apps/server/lib/lexical/server/task_queue.ex
```diff
@@ -119,21 +119,19 @@ defmodule Lexical.Server.TaskQueue do
...
+      error = ResponseError.new(code: code, message: message)
 
       Transport.write(%{id: id, error: error})
     end
 
-    defp run_task(fun) when is_function(fun) do
-      Task.Supervisor.async_nolink(task_supervisor_name(), fun)
+    defp run_task(fun, mfa) when is_function(fun) do
+      task = Task.Supervisor.async_nolink(task_supervisor_name(), fun)
+      %Task{task | mfa: mfa}
```
zachallaun, line 129: I'm not sure it matters, but the mfa in a task should match the [mfa type](https://hexdocs.pm/elixir/1.17.2/typespecs.html#built-in-types), which is `{module, fun, arity}` instead of `{module, fun, args}`.

So we could do something like:

```suggestion
    defp run_task(fun, {m, f, args}) when is_function(fun) do
      task = Task.Supervisor.async_nolink(task_supervisor_name(), fun)
      %Task{task | mfa: {m, f, length(args)}}
```

apps/remote_control/lib/lexical/remote_control/mix.ex
```diff
@@ -11,13 +10,17 @@ defmodule Lexical.RemoteControl.Mix do
...
+  end
+
   def in_project(%Project{} = project, fun) do
     # Locking on the build make sure we don't get a conflict on the mix.exs being
     # already defined
 
     old_cwd = File.cwd!()
 
-    Build.with_lock(fn ->
+    with_lock(fn ->
```
zachallaun, line 23: If it's not used elsewhere, my preference would be to not introduce a new function, and instead inline this:

```suggestion
    RemoteControl.with_lock(__MODULE__, fn ->
```

zachallaun (approved):
LGTM, made a few optional suggestions.

zachallaun (changes requested):
Actually: Looking at the CI failures, 1.13 doesn't support the `:mfa` attribute in tasks.

zachallaun (approved):

apps/remote_control/lib/lexical/remote_control/mix.ex
scohen: I'll make this private.

PR #826, opened 2024-09-05
Improve performance under batch operations

scohen:
This commit seeks to improve the performance under batch operations by holding off compilations until after a period of quiescence. Prior, we were building every 100 or so milliseconds, but this would cause builds to happen during batch operations.

Instead, this PR dramatically simplifies the build process to utilize timeouts to detect a quiet period after which builds can commence.

I also found a missed case while transforming diagnostics into elixir that caused crashes that took down the project node. I was unaware that the uri of a diagnostic can be nil (how?!?).

Prior to this PR, mass renames would never finish in emacs, and now they can.

### DIFF
--- a/apps/remote_control/lib/lexical/remote_control/build/state.ex
+++ b/apps/remote_control/lib/lexical/remote_control/build/state.ex
@@ -24,23 +24,20 @@ defmodule Lexical.RemoteControl.Build.State do
   end
 
   def on_timeout(%__MODULE__{} = state) do
-    case state.project_compile do
-      :none -> :ok
-      :force -> compile_project(state, true)
-      :normal -> compile_project(state, false)
-    end
-
     new_state =
-      state.uri_to_document
-      |> Map.values()
-      |> Enum.reduce(
-        state,
-        fn document, state ->
-          new_state = increment_build_number(state)
-          compile_file(new_state, document)
-          new_state
-        end
-      )
+      case state.project_compile do
+        :none -> state
+        :force -> compile_project(state, true)
+        :normal -> compile_project(state, false)
+      end
+
+    # We need to compile the individual documents even after the project is
+    # compiled because they might have unsaved changes, and we want that state
+    # to be the latest state of the project.
+    new_state =
+      Enum.reduce(new_state.uri_to_document, state, fn {_uri, document}, state ->
+        compile_file(state, document)
+      end)
 
     %__MODULE__{new_state | uri_to_document: %{}, project_compile: :none}
   end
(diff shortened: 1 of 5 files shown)


apps/remote_control/lib/lexical/remote_control/build/state.ex
```diff
@@ -14,33 +14,54 @@ defmodule Lexical.RemoteControl.Build.State do
...
+  def on_timeout(%__MODULE__{} = state) do
+    case state.project_compile do
+      :none -> :ok
+      :force -> compile_project(state, true)
+      :normal -> compile_project(state, false)
+    end
+
+    new_state =
+      state.uri_to_document
+      |> Map.values()
```
zachallaun, line 35: Optional: can get rid of this and just do `Enum.reduce(state.uri_to_document, fn {_uri, document}, state -> ... end)`

apps/remote_control/lib/lexical/remote_control/build/state.ex
```diff
@@ -14,33 +14,54 @@ defmodule Lexical.RemoteControl.Build.State do
...
-            {new_state, [uri | compiled_uris]}
-          else
-            {state, compiled_uris}
-          end
-      end)
+  def on_timeout(%__MODULE__{} = state) do
+    case state.project_compile do
+      :none -> :ok
+      :force -> compile_project(state, true)
+      :normal -> compile_project(state, false)
```
zachallaun, line 30: Do we need to increment the build number when we compile the project? If so, I wonder if compile_project and compile_file should do it internally and return the state.

apps/remote_control/lib/lexical/remote_control/build/state.ex
```diff
@@ -14,33 +14,54 @@ defmodule Lexical.RemoteControl.Build.State do
...
+    if force? do
+      %__MODULE__{state | project_compile: :force}
+    else
+      %__MODULE__{state | project_compile: :normal}
+    end
+  end
+
   def compile_scheduled?(%__MODULE__{} = state, uri) do
-    Map.has_key?(state.uri_to_source_and_edit_time, uri)
+    Map.has_key?(state.uri_to_document, uri)
```
zachallaun, line 64: I'm not sure where `compile_scheduled?` is used, but should this be either the uri is scheduled or :project_compile is not :none?

apps/remote_control/lib/lexical/remote_control/build/state.ex
scohen: yes, we do, it's how we keep diagnostics in sync.

apps/remote_control/lib/lexical/remote_control/build/state.ex
scohen: seems like it's only used for testing. I can remove.

zachallaun (approved):
Looks good to me after changes 👍

PR #828, opened 2024-09-18
Enable ets compression on OTP 27.1

dvic:
https://github.com/erlang/otp/issues/8682#issuecomment-2359163692

I can confirm also in our projects that it's fixed now.


scohen (approved):

scohen:
@dvic The tests were failing due to dialyzer, which had nothing to do with you, Closing this in favor of #836, which is the same PR as this. Thank you, and I'm sorry it took me so long to merge.

PR #839, opened 2024-11-20
Show syntax macro completions when `Kernel.` is prefixed to the cursor.

### LINKED ISSUE #832: LSP is not suggesting some Kernel operators
Screenshot 2024-10-11 at 2 28 02 PM
Screenshot 2024-10-11 at 2 28 09 PM

It looks like it only sees two operators with `=`?

Moosieus:
Closes #832.

This solution is imperfect because it requires the prefix to end exactly with `Kernel.`

However it works in most cases because:
* Completions are usually requested at the point of `Kernel.|`
* Completions are cached by clients for the next few characters.

In the exceptional rare case where a user copy/pastes something like `Kernel.=|` and requests a completion, these won't show. Not a perfectly correct solution, but our handling of syntax macros is already a matter of taste, so ¯\\\_(ツ)\_/¯

### DIFF
--- a/apps/common/lib/lexical/ast/env.ex
+++ b/apps/common/lib/lexical/ast/env.ex
@@ -21,7 +21,8 @@ defmodule Lexical.Ast.Env do
     :suffix,
     :position,
     :position_module,
-    :zero_based_character
+    :zero_based_character,
+    :detected_contexts
   ]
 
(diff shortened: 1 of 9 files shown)


Moosieus:
Why are there extra changes 0_0

PR #841, opened 2024-12-01
Performance oriented fixes.

scohen:
This commit fixes a number of performance related issues, and reworks formatting so it's lock-free and able to be used with format-on-save in any editor. The fixes are:

  * We only call deps.compile if the vm changes directories during compilation. This dramatically speeds up compiling erlang deps.
  * Adds a timing helper to time blocks of code
  * Precalculates contexts around the cursor. Fetching the current context was called once for every completion item. This meant that for large modules, like Enum, hundreds of context calculations were happening.
  * Reworked formatting: Jose added a couple of options to Mix.Tasks.Format.formatter_for_file that allow us to not need to be in the project context in order for formatting to work. This enables lock-free formatting which makes the amount of time formatting takes to be proportional to the length of the file. Formatting on save no longer waits on compilation, and is safe for any project

### DIFF
--- a/apps/common/lib/future/mix/tasks/format.ex
+++ b/apps/common/lib/future/mix/tasks/format.ex
@@ -1,0 +1,1047 @@
+defmodule Mix.Tasks.Future.Format do
+  use Mix.Task
+
+  @shortdoc "Formats the given files/patterns"
+
+  @moduledoc """
+  Formats the given files and patterns.
+
+      $ mix format mix.exs "lib/**/*.{ex,exs}" "test/**/*.{ex,exs}"
+
+  If any of the files is `-`, then the input is read from stdin and the output
+  is written to stdout.
+
+  ## Formatting options
+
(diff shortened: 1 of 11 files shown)


scohen:
I've been working on this branch for a couple of weeks and haven't noticed any regressions or issues. The improvements are drastic, especially on my project.
| label | main | performance |
|--------|--------|--------|
| project compilation | 27s | 1.8s |
| completion of `Enum.` | 1.8s | < 1ms |
| formatting worst case | 4.7s | 110ms |

Notes: 
1. My project has a number of erlang deps, and each one would take 750ms to compile 
2. The improvement in formatting was eliminating waiting on the build so the project lock would become available.

Moosieus (approved):
Code checks out, and I'm continuing to use this branch without issue.

Would it make sense to change `Lexical.Debug.log_timed/3` to `debug` level instead of `info`? That's the only outstanding consideration I have.

PR #847, opened 2024-12-27
Add Elixir 1.18 to allowed Elixir versions

### LINKED ISSUE #846: Elixir 1.18 support
Lexical should support 1.18.

kevinschweikert:
Closes #846


apps/server/lib/lexical/server/boot.ex
```diff
@@ -88,7 +88,8 @@ defmodule Lexical.Server.Boot do
     "1.15.0" => ">= 1.15.3",
     "1.16.0" => ">= 1.16.0",
     "1.17.0-rc" => ">= 1.17.0-rc",
-    "1.17.0" => ">= 1.17.0"
+    "1.17.0" => ">= 1.17.0",
+    "1.80.0" => ">=1.18.0"
```
sorentwo, line 92: Typo here. We have a long while until v1.80!

```suggestion
    "1.18.0" => ">= 1.18.0"
```

sorentwo:
Thanks for the patch. Happy to report that with this patch (and the small correction I commented about), Lexical works normally with Elixir v1.18.

apps/server/lib/lexical/server/boot.ex
kevinschweikert: Thanks @sorentwo for the correction! Must have been wishful thinking 😉

scohen:
There's more that needs to be done to get lexical to properly support 1.18. Closing in favor of #849

PR #848, opened 2024-12-27
Fix typing violation on Elixir 1.18

kevinschweikert:
In Elixir 1.18 we get the following typing violation

```elixir
==> server
Compiling 3 files (.ex)
     warning: the following clause will never match:

         %{started_at: ts, mfa: {m, f, a}}

     because it attempts to match on the result of:

         task

     which has type:

         dynamic(%Task{mfa: term(), owner: term(), pid: term(), ref: term()})

     typing violation found at:
     │
 143 │         %{started_at: ts, mfa: {m, f, a}} ->
     │         ~
     │
     └─ (server 0.7.2) lib/lexical/server/task_queue.ex:143:9: Lexical.Server.TaskQueue.State.log_task_run_time/2

Generated server app
```

because the `started_at` field is added to the `Task` struct. To avoid this i store the started_at timestamp with the task pid as the key and retrieve it when needed.

### DIFF
--- a/apps/server/lib/lexical/server/task_queue.ex
+++ b/apps/server/lib/lexical/server/task_queue.ex
@@ -6,7 +6,7 @@ defmodule Lexical.Server.TaskQueue do
     import Lexical.Logging
     require Logger
 
-    defstruct ids_to_tasks: %{}, pids_to_ids: %{}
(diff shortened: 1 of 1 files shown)


scohen:
Thank you for doing this, but there's more that needs to be done from a project perspective to properly support 1.18. Closing in favor of #849

